PR #1, opened 2020-06-09
Fix package structure for androidTest folder

android10:
This PR moves classes around (```androidTest```) in order to be consistent with the rest of the application package structure.

EXTRA BALL: Added extra information to its corresponding ```gradle``` task.

### DIFF
--- a/app/src/androidTest/kotlin/com/wire/android/WireApplicationTest.kt
+++ b/app/src/androidTest/kotlin/com/wire/android/WireApplicationTest.kt
@@ -1,6 +1,7 @@
-package com.android.wire
+package com.wire.android
 
 import androidx.test.platform.app.InstrumentationRegistry
+import com.wire.android.InstrumentationTest
 import org.junit.Assert.assertEquals
 import org.junit.Test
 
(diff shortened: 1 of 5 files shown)


app/src/androidTest/kotlin/com/wire/android/WireApplicationTest.kt
```diff
@@ -1,6 +1,7 @@
-package com.android.wire
+package com.wire.android
```
BradleyWire, line 1: Do you think we need `android` in the package naming structure?

android10: As I naming convention I used in the past, this means you are in the main module. It is implicit and then if you have other ones you could use other extensions. But I'm not strong on this one, even though I find it more clear. 

Any suggestions?

mejdoo: I think it's ok to keep `com.wire.android`. I know that package names in Android generally follow the convention: **inversed-domain-name.product-name**. When the product name is very similar or identical to the domain, the word `android` could be used. It's not a rule, anyone if free to choose any applicationID if it's not already used by someone else in google play. In any case, we cannot use com.wire because we are going to create an other google play app not to update the current one. https://play.google.com/store/apps/details?id=com.wire

android10: This is really a good point. :heart: Thanks for making it clear.

PR #2, opened 2020-06-09
Remove local.properties file from remote repo - Part 1

gizemb:
local.properties file should not be pushed to remote repo because it contains environment specific settings.

### DIFF
--- a/local.properties
+++ b/local.properties
@@ -1,8 +1,0 @@
-## This file must *NOT* be checked into Version Control Systems,
-# as it contains information specific to your local configuration.
-#
-# Location of the SDK. This is only used by Gradle.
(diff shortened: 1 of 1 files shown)


BradleyWire:
Let's update the .gitignore if this is the case 

> I think updating `.gitignore ` from:
> 
> ```
> # Local configuration file (sdk path, etc)
> local.properties.kts
> ```
> 
> to:
> 
> ```
> # Local configuration file (sdk path, etc)
> local.properties
> ```
> 
> may also help.

Noticed you've done this in 2 parts.

gizemb:
@BradleyWire it has to be in 2 parts because if I update the gitignore right now, it won't get the "deletion" change that I made here :)

PR #4, opened 2020-06-10
Add CreateAccountActivity

gizemb:
- Adds CreateAccountActivity as the launcher screen (blank for now).
- Fixes FunctionalTest setup

Screenshot_1591956517


app/src/main/res/values/colors.xml
```diff
@@ -1,6 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="colorPrimary">#6200EE</color>
-    <color name="colorPrimaryDark">#3700B3</color>
-    <color name="colorAccent">#03DAC5</color>
+    <color name="white">#FFFFFFFF</color>
+
+    <color name="black">#FF000000</color>
+
+    <color name="graphite">#FF33373A</color>
```
mejdoo, line 7: We need to find a solution for color naming before adding more and more colors to the new codebase. Let's agree on something and not use the legacy colors names.

mejdoo: Maybe we can name all the colors by function to avoid headaches.
https://www.novatec-gmbh.de/en/blog/name-android-colors-palettes/

gizemb: do we have time for this now? I think that's more or less what Bradley suggested in the planning yesterday. As far as I understand, we decided to first quickly move the codebase, and iteratively change what's needed. @android10 how should we proceed?

app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivity.kt
```diff
@@ -0,0 +1,20 @@
...
+class CreateAccountActivity : AppCompatActivity(R.layout.activity_create_account) {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+//        replaceFragment(R.id.activityCreateAccountLayoutContainer, CreateAccountFragment.newInstance(), false)
+        initBackButton()
+    }
+
+    private fun initBackButton() {
+        activityCreateAccountBackButton.bringToFront()
```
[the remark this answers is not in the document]
mejdoo: @android10  we are bringing the button to the front not the activity. If we don't call `bringToFront` this button will be hidden by the FrameLayout. @gizemb Don't remove this logic, maybe rename `activityCreateAccountBackButton` to `createAccountBackButton` to avoid any confusion. Rename the FrameLayout as well to `createAccountLayoutContainer`

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -24,14 +24,17 @@ object AndroidSdk {
...
 
-    const val kotlinStdLib     = "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
-    const val appCompat        = "androidx.appcompat:appcompat:${Versions.jetpack}"
-    const val constraintLayout = "androidx.constraintlayout:constraintlayout:${Versions.constraintLayout}"
-    const val ktxCore          = "androidx.core:core-ktx:${Versions.ktx}"
+    const val kotlinStdLib = "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
+    const val appCompat = "androidx.appcompat:appcompat:${Versions.jetpack}"
+    const val material = "com.google.android.material:material:${Versions.material}"
+    const val constraintLayout =
+        "androidx.constraintlayout:constraintlayout:${Versions.constraintLayout}"
```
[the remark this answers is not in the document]
gizemb: I kept the style as it is, but it would be more convenient if we could just do cmd+alt+L on these build script files.

app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivity.kt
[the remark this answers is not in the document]
gizemb: I solved this problem with elevation. It felt more natural to deal with it in the xml file rather than doing it programmatically.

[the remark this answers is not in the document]
mejdoo: ok makes sense

app/src/main/res/layout/activity_create_account.xml
```diff
@@ -4,16 +4,24 @@
...
-    <TextView
-        android:id="@+id/hello"
+    <ImageButton
+        android:id="@+id/createAccountBackButton"
+        style="@style/ImageButton.Back"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:text="Hello Wire Android Reloaded!"
-        app:layout_constraintBottom_toBottomOf="parent"
+        android:elevation="1dp"
```
mejdoo, line 14: Move it to dimen.xml

app/src/main/res/layout/activity_create_account.xml
```diff
@@ -4,16 +4,24 @@
...
+        android:padding="@dimen/back_button_padding"
         app:layout_constraintLeft_toLeftOf="parent"
-        app:layout_constraintRight_toRightOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        tools:ignore="ContentDescription" />
```
BradleyWire, line 18: What's the reason for ignoring this?

BradleyWire, line 18: Instead of ignoring the warning, do we want to start considering adding content descriptions to aid accessibility in this new project?

gizemb: I put the contentDescription in the `ImageButton.Back` style, but A.S. cannot read inside the style and complains that there isn't one. As an alternative, I can move the contentDescription here

BradleyWire: It's clever, but, It might be better placed here to be honest, I personally don't think it's the responsibility of the styles. We can put it to the team for a discussion, but, ignoring this warning seems like an avoidable overhead here.

gizemb: fair enough. done 👍

app/src/main/kotlin/com/wire/android/core/extension/Activity.kt
```diff
@@ -0,0 +1,13 @@
+package com.wire.android.core.extension
+
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+
+fun AppCompatActivity.replaceFragment(frameId: Int, fragment: Fragment, addToBackStack: Boolean = true) =
+    supportFragmentManager.doTransaction {
+        replace(frameId, fragment).apply {
```
android10, line 8: In terms of Kotlin Scope functions I think we should use ```also``` here. ```apply``` should be used when you are accessing properties and static fields of the object passed in the lambda.

app/src/main/kotlin/com/wire/android/core/extension/Fragment.kt
```diff
@@ -0,0 +1,7 @@
+package com.wire.android.core.extension
+
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentTransaction
+
+inline fun FragmentManager.doTransaction(func: FragmentTransaction.() -> FragmentTransaction) =
```
android10, line 6: I have seen a pattern naming this ```inTransaction``` lately. I guess it is a matter of taste.

app/src/main/res/layout/activity_create_account.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
android10, line 2: Very concise with Styles and easy to understand :heart:

app/src/androidTest/kotlin/com/wire/android/FunctionalTest.kt
```diff
@@ -9,13 +9,8 @@ import org.junit.runner.RunWith
...
+abstract class FunctionalTest(clazz: Class<out Activity>) {
 
     @get:Rule
-    var activityRule = activityTestRule(clazz)
-
-    private fun activityTestRule(activityClass: Class<*>): ActivityTestRule<out Activity> {
-        require(activityClass is Activity) { "Wrong class type: Use Android Activity type." }
-        return ActivityTestRule(activityClass.asSubclass(Activity::class.java))
-    }
+    val activityRule = ActivityTestRule(clazz)
```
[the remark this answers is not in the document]
android10: Great then! We want to have those ```Failures``` right away :+1:

PR #6, opened 2020-06-10
Add CreateAccountFragment

gizemb:
- Adds CreateAccountFragment w/ Personal/Pro buttons

Screenshot_1591963611


app/src/main/res/values/integers.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <integer name="wire__animation__delay__short">100</integer>
```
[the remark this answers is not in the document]
gizemb: copy paste. removed anyway

app/src/main/res/values/colors_light_theme.xml
```diff
@@ -4,6 +4,8 @@
     <color name="light_theme_primary_dark">@color/black</color>
     <color name="light_theme_accent">@color/blue</color>
 
+    <color name="light_theme_clickable_item_pressed">@color/black_8</color>
```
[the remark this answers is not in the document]
gizemb: when I see black_8, I understand that it's black color with 8% alpha. it's actually commonly used. removed anyway..

app/src/main/res/values/colors.xml
```diff
@@ -1,14 +1,20 @@
...
+    <color name="transparent">#00000000</color>
+
     <color name="white">#FFFFFFFF</color>
 
     <color name="black">#FF000000</color>
+    <color name="black_8">#14000000</color>
 
     <color name="blue">#2492D3</color>
 
+    <color name="graphite">#FF33373A</color>
```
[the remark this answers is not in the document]
gizemb: any suggestions? it is a shade of grey, as graphite itself.

app/src/main/res/layout/fragment_create_account.xml
```diff
@@ -0,0 +1,110 @@
...
+    android:background="@color/authentication_background">
+
+    <androidx.constraintlayout.widget.Guideline
+        android:id="@+id/centerGuideline"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constraintGuide_percent="0.5" />
+
+    <View
```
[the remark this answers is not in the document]
gizemb: I did as much as I can.

app/src/main/res/values/themes_authentication.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="AppTheme.Authentication">
+        <item name="android:colorBackground">#f8f8f8</item>
+        <item name="android:textColorPrimary">@color/graphite</item>
```
gizemb, line 5: I'm not sure how to proceed here. If there's a non-repeating color:
should I create a color palette and reference those colors here (as in textColorPrimary),
or should I directly add the hex color value here (as in colorBackground)?

let me know what you think :)

BradleyWire: Not sure how reputable this [project](https://github.com/futurice/android-best-practices) is, but, they do offer some [guidance](https://github.com/futurice/android-best-practices#colorsxml) on colours in particular with styles. I personally like their approach.

For single use colours, maybe we need to ask why we're only using this colour once in the app? This might help us mitigate a polluted `colors.xml` file like we have in wire-android. But, normally as a rule of thumb I try and avoid hardcoded values where I can outside of their respective xml files.

app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
```diff
@@ -12,8 +12,15 @@ class CreateAccountActivityUITest : FunctionalTest(CreateAccountActivity::class.
 
     @Test
     fun launch_uiElementsVisible() {
-        //TODO check other elements
         onView(withId(R.id.createAccountBackButton)).check(matches(isDisplayed()))
+
+        //Personal
```
android10, line 17: Yeah. I want to have all these tests which will serve for setting up Megazord cc @Gardosen

app/src/main/kotlin/com/wire/android/core/extension/Fragment.kt
```diff
@@ -1,7 +1,14 @@
...
+import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import androidx.fragment.app.FragmentTransaction
 
 inline fun FragmentManager.inTransaction(func: FragmentTransaction.() -> FragmentTransaction) =
-    beginTransaction().func().commit()
\ No newline at end of file
+    beginTransaction().func().commit()
+
+//TODO: Fragment shouldn't be aware of activity's view ids
```
android10, line 11: Is it something that we can fix here instead of keeping the TODO?

gizemb: this requires a navigation structure, or a communication mechanism between the fragment & activity. so, no we can't fix it now :)

app/src/main/res/values/themes_authentication.xml
gizemb: we need one-time colors at some point to introduce new colors to our theme, which cannot be defined as a variation of existing ones. Think about accent color (`secondaryColor`), for example. 

The way I want to structure the theming is that in all widget styles & xml files under `res/drawable`, we should ONLY reference colors by theme attributes. Ideally we should be able to "paint" every widget with our existing theme colors (slight modifications like alpha is of course possible). 

For instance, a button style references  `?attr/secondaryColor` as its background color. It actually doesn't know that it's blue, it just wants to use the secondaryColor. This will be of great help to us when setting up dark theme. 

If we go with this logic, the color "blue" is only needed once, for the relation `secondaryColor = blue`, but it's actually seen in most places.

---
My question is:
should there still be a color palette somewhere in `colors.xml`? This has the advantage that it lets you see all the colors at once in one place. Keeps it tidy. But it has the disadvantage that, once I create a` @color/blue` resource, it can be referenced from everywhere in the app.

with color palette:
res/colors.xml:
```
<color name="white">#ffffff</color>
<color name="blue">#2492d3</color>
...
```
themes.xml
```
<item name="colorSecondary">@color/blue</item>
```

VS.

without color palette:
themes.xml
```
<item name="colorSecondary">#2492d3</item>
```

PR #7, opened 2020-06-11
Set up app light theme

gizemb:
Creates a Material design light theme, similar to wire-android color palette.

- Minor adjustments can be made later.

Screenshot_1591892763


app/src/main/res/color/button_secondary_textcolor.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BradleyWire, line 1: `textcolor` can be `text_color` in the file name, unless we're grouping the underscores by attributes moreover words?

gizemb: yes, my intention was grouping by attributes, but I can change if the other one is more convenient

gizemb: hmm, I gave it a second thought and i think it's better to group them by word, because in layout files, we also group by words. for instance: `activity_create_account`, NOT, `activity_createaccount`. I changed that for consistency. thanks for pointing this out.

PR #8, opened 2020-06-12
Fix/refactor SourceSets and Infrastructure Script -  Gradle Kotlin DSL

android10:
Not much to say rather than a minor refactor to the build system in order to follow up best practices. 
In this case the code speaks by itself. :1234:

### DIFF
--- a/buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
+++ b/buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
@@ -4,8 +4,10 @@ import scripts.Variants_gradle.*
 import java.util.*
 
 private object Default {
-    val BUILD_TYPE = BuildTypes.DEBUG.capitalize()
-    val BUILD_FLAVOR = ProductFlavors.DEV.capitalize()
+    const val BUILD_TYPE = BuildTypes.DEBUG
+    const val BUILD_FLAVOR = ProductFlavors.DEV
+
+    val VARIANT = "${BUILD_FLAVOR.capitalize()}${BUILD_TYPE.capitalize()}"
 }
 
 tasks.register("clean", Delete::class){
@@ -19,22 +21,22 @@ tasks.named<Wrapper>("wrapper") {
 
 tasks.register("runUnitTests") {
     description = "Runs all Unit Tests."
-    dependsOn(":app:test${Default.BUILD_FLAVOR}${Default.BUILD_TYPE}UnitTest")
+    dependsOn(":app:test${Default.VARIANT}UnitTest")
 }
 
 tasks.register("runAcceptanceTests") {
     description = "Runs all Acceptance Tests in the connected device."
-    dependsOn(":app:connected${Default.BUILD_FLAVOR}${Default.BUILD_TYPE}AndroidTest")
+    dependsOn(":app:connected${Default.VARIANT}AndroidTest")
 }
 
 tasks.register("compileApp") {
     description = "Compiles the Wire Android Client."
-    dependsOn(":app:assemble${Default.BUILD_FLAVOR}${Default.BUILD_TYPE}")
+    dependsOn(":app:assemble${Default.VARIANT}")
 }
 
 tasks.register("runApp", Exec::class) {
     val compileAppTask = "compileApp"
-    val installAppTask = ":app:install${Default.BUILD_FLAVOR}${Default.BUILD_TYPE}"
+    val installAppTask = ":app:install${Default.VARIANT}"
 
     description = "Compiles and runs the Wire Android Client in the connected device."
     dependsOn(compileAppTask, installAppTask)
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
```diff
@@ -13,31 +19,37 @@ tasks.named<Wrapper>("wrapper") {
...
 
     val localProperties = File(project.rootDir, "local.properties")
     if (localProperties.exists()) {
         val properties = Properties()
         localProperties.inputStream().use { properties.load(it) }
         val sdkDir = properties["sdk.dir"]
         val adb = "${sdkDir}/platform-tools/adb"
+        
+        val applicationPackage = "com.wire.android.${Default.BUILD_FLAVOR.toLowerCase()}.${Default.BUILD_TYPE.toLowerCase()}"
+        val launchActivity = "com.wire.android.MainActivity"
```
BradleyWire, line 51: This reference will change when #4 is merged in as shown in [these changes](https://github.com/wireapp/wire-android-reloaded/pull/4/files#diff-c8cc2dd42271f2bf39c6aa81eb6a5529)

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
```diff
@@ -13,31 +19,37 @@ tasks.named<Wrapper>("wrapper") {
 
 tasks.register("runUnitTests") {
     description = "Runs all Unit Tests."
-    dependsOn(":app:testDebugUnitTest")
+    dependsOn(":app:test${Default.BUILD_FLAVOR}${Default.BUILD_TYPE}UnitTest")
```
gizemb, line 22: `${Default.BUILD_FLAVOR}${Default.BUILD_TYPE}` is repeated too much. maybe we can combine them into 
`val variant = "${BUILD_FLAVOR.capitalize()}${BUILD_TYPE.capitalize()}"`
  in `Default`? 

In that way, we also wouldn't have to make `.toLowerCase()` for `applicationPackage`

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
```diff
@@ -1,7 +1,13 @@
 package scripts
 
+import scripts.Variants_gradle.*
 import java.util.*
 
+private object Default {
+    val BUILD_TYPE = BuildTypes.DEBUG.capitalize()
+    val BUILD_FLAVOR = ProductFlavors.DEV.capitalize()
```
gizemb, line 8: can they be `const`?

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
gizemb: big ➕ from me :)

isn't there a command to launch the `LAUNCHER` activity w/o giving activity name explicitly?

android10: I already check this out on the other PR and commented it. It will break. Unless you want to parse the Manifest, I do not see any other way. And this command is a helper that I find useful especially once we establish which is our starting Activity. 

I use the command line and like to compile and run everything from there. We could perfectly remove this task if necessary.

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
android10: They cannot. Gradle complains because this comes from autogenerated compiled scripts

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
android10: Parsing the ```AndroidManifest``` would add too much complexity for such an insignificant task. The Welcome Activity should not change in time, since Android forces you to define one.

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
android10: I do not fully understand your point. 

For ```applicationPackage``` we need to concatenate ```.dev.debug```
For ```gradle``` tasks we need "DevDebug". 

Maybe I'm missing something but I do not see a solution that applies to all with only one constant.

gizemb:
```
private object Default {
    val BUILD_TYPE = BuildTypes.DEBUG         <-- removed ".capitalize()"
    val BUILD_FLAVOR = ProductFlavors.DEV         <-- removed ".capitalize()"
    
    val VARIANT = "${BUILD_FLAVOR.capitalize()}${BUILD_TYPE.capitalize()}"     <-- combined the logic here
}


tasks.register("runAcceptanceTests") {
    ...
    dependsOn(":app:connected${Default.VARIANT}AndroidTest")
}


tasks.register("compileApp") {
     ...
    dependsOn(":app:assemble${Default.VARIANT}")
}


tasks.register("runApp", Exec::class) {
    ...
    val applicationPackage = "com.wire.android.${Default.BUILD_FLAVOR}.${Default.BUILD_TYPE}"
}
```

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
gizemb: parsing the manifest is too much work and not necessary. I can update this script on my PR, no problem

buildSrc/src/main/kotlin/scripts/infrastructure.gradle.kts
android10: Got it! Thanks for clearing it up. I will modify this and merge oki?

PR #9, opened 2020-06-12
Add landscape layout for CreateAccountFragment

gizemb:
Screenshot_1591976201


app/src/androidTest/kotlin/com/wire/android/FunctionalTest.kt
```diff
@@ -13,4 +15,12 @@ abstract class FunctionalTest(clazz: Class<out Activity>) {
 
     @get:Rule
     val activityRule = ActivityTestRule(clazz)
+
+    val uiDevice = UiDevice.getInstance(getInstrumentation())
+
+    fun inRotatedMode(block: () -> Unit) = with(uiDevice) {
+        setOrientationLeft()
+        block()
+        setOrientationNatural()
```
gizemb, line 24: This is executed pretty fast for now, but I don't know how much it slows down the total execution when it changes the orientation switching between many test cases. There are only 2 right now 🙃

As an alternative, I thought of creating a `ScreenRotation` rule that rotates the device before *all* test cases, and rotates it back after all has been executed. Then, I need to put all the landscape tests in a separate test class which has that rule. One drawback is that it requires having 2 test classes per screen, one inheriting from each other, which is a bit weird...

 We can also completely ignore landscape ui tests, but I think it's nice to have them. I just wanted to start a discussion here.

android10: Your points make sense. From my perspective I would not force to all tests to be run in both portrait and landscape mode. Instead that should be something optional. 

My vision is that we can group tests in the future and pick which are the ones more critical in landscape mode for example: Tablet support comes to my mind right now. 

We can start for now simple by just testing stuff we consider important in landscape mode and move towards complexity if we have the need. Also Functional Testing is usually slow but the faster the better of course. 

Opinions @mejdoo @BradleyWire @makingthematrix?

app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
```diff
@@ -22,5 +22,10 @@ class CreateAccountActivityUITest : FunctionalTest(CreateAccountActivity::class.
         onView(withId(R.id.createProAccountTitleTextView)).check(matches(isDisplayed()))
         onView(withId(R.id.createProAccountDescriptionTextView)).check(matches(isDisplayed()))
     }
+
+    @Test
+    fun launch_uiElementsVisibleInLandscape() = inRotatedMode {
```
android10, line 27: I would continue using ```camelCase``` for consistency. Any other reason for the usage of ```_```?

app/src/androidTest/kotlin/com/wire/android/FunctionalTest.kt
mejdoo: I think it's gonna be worth to implement functional tests for landscape mode when we support tablets. Let's keep in simple for now.

app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
gizemb: It actually adds a layer between what's being tested and the expected outcome. Kind of like a `givenThis_whenThat_thenCheckThose()` thing. Except that we don't have given in this case (Given that app is installed?). So, it's like (when screen is launched)_(check that ui elements are visible). This is fine btw, according to Kotlin coding style guidelines: https://kotlinlang.org/docs/reference/coding-conventions.html#function-names

android10: Fair enough.

PR #10, opened 2020-06-15
Registration #4: Add Create Account - Personal screen

gizemb:
- No Email/Phone entry fields for now
- Sets up TabLayout style
- Adds first unit test

Screenshot_1592237562
Screenshot_1592237548

### DIFF
--- a/app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
+++ b/app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
@@ -1,9 +1,9 @@
 package com.wire.android.feature.auth.registration
 
 import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.assertion.ViewAssertions.matches
-import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
-import androidx.test.espresso.matcher.ViewMatchers.withId
+import androidx.test.espresso.matcher.ViewMatchers.*
 import com.wire.android.FunctionalTest
 import com.wire.android.R
 import org.junit.Test
@@ -26,6 +26,22 @@ class CreateAccountActivityUITest : FunctionalTest(CreateAccountActivity::class.
     @Test
     fun launch_uiElementsVisibleInLandscape() = inRotatedMode {
         launch_uiElementsVisible()
+    }
+
+    @Test
+    fun personal_uiElementsVisible() {
+        onView(withId(R.id.createPersonalAccountTitleTextView)).perform(click())
+
+        onView(withText(R.string.create_personal_account_title)).check(matches(isDisplayed()))
+        onView(withText(R.string.authentication_tab_layout_title_email)).check(matches(isDisplayed()))
+        onView(withText(R.string.authentication_tab_layout_title_phone)).check(matches(isDisplayed()))
+
+        //TODO: check whether Email entry area is visible
+    }
+
+    @Test
+    fun personal_uiElementsVisibleInLandscape() = inRotatedMode {
+        personal_uiElementsVisible()
     }
 }
 
--- a/app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
+++ b/app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
@@ -24,7 +24,7 @@ class CreateAccountActivityUITest : FunctionalTest(CreateAccountActivity::class.
     }
 
     @Test
-    fun launch_uiElementsVisibleInLandscape() = inRotatedMode {
+    fun launch_uiElementsVisibleInLandscape() = rotateScreen {
         launch_uiElementsVisible()
     }
 
@@ -40,7 +40,7 @@ class CreateAccountActivityUITest : FunctionalTest(CreateAccountActivity::class.
     }
 
     @Test
-    fun personal_uiElementsVisibleInLandscape() = inRotatedMode {
+    fun personal_uiElementsVisibleInLandscape() = rotateScreen {
         personal_uiElementsVisible()
     }
 }
--- a/app/src/main/kotlin/com/wire/android/core/extension/Activity.kt
+++ b/app/src/main/kotlin/com/wire/android/core/extension/Activity.kt
@@ -1,5 +1,7 @@
 package com.wire.android.core.extension
 
+import android.app.Activity
+import android.view.inputmethod.InputMethodManager
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
 
@@ -11,3 +13,8 @@ fun AppCompatActivity.replaceFragment(frameId: Int, fragment: Fragment, addToBac
             }
         }
     }
+
+fun Activity.showKeyboard() {
+    val inputMethodManager = this.getSystemService(Activity.INPUT_METHOD_SERVICE) as InputMethodManager
+    inputMethodManager.toggleSoftInput(InputMethodManager.SHOW_IMPLICIT, InputMethodManager.HIDE_IMPLICIT_ONLY)
+}
--- a/app/src/main/kotlin/com/wire/android/core/extension/Activity.kt
+++ b/app/src/main/kotlin/com/wire/android/core/extension/Activity.kt
@@ -2,6 +2,8 @@ package com.wire.android.core.extension
 
 import android.app.Activity
 import android.view.inputmethod.InputMethodManager
+import android.view.inputmethod.InputMethodManager.HIDE_IMPLICIT_ONLY
+import android.view.inputmethod.InputMethodManager.SHOW_IMPLICIT
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
 
@@ -14,7 +16,5 @@ fun AppCompatActivity.replaceFragment(frameId: Int, fragment: Fragment, addToBac
         }
     }
 
-fun Activity.showKeyboard() {
-    val inputMethodManager = this.getSystemService(Activity.INPUT_METHOD_SERVICE) as InputMethodManager
-    inputMethodManager.toggleSoftInput(InputMethodManager.SHOW_IMPLICIT, InputMethodManager.HIDE_IMPLICIT_ONLY)
-}
+fun Activity.showKeyboard() =
+    (this.getSystemService(Activity.INPUT_METHOD_SERVICE) as InputMethodManager).toggleSoftInput(SHOW_IMPLICIT, HIDE_IMPLICIT_ONLY)
--- a/app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivity.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivity.kt
@@ -10,11 +10,22 @@ class CreateAccountActivity : AppCompatActivity(R.layout.activity_create_account
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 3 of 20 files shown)


app/src/main/kotlin/com/wire/android/feature/auth/registration/personal/CreatePersonalAccountFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    private fun initViewPager() {
+        val titles = listOf(
+            getString(R.string.authentication_tab_layout_title_email),
+            getString(R.string.authentication_tab_layout_title_phone)
+        )
+        createPersonalAccountViewPager.adapter = CreatePersonalAccountViewPagerAdapter(
+            childFragmentManager, titles
+        )
```
BradleyWire, line 25: If we're parameter list-wrapping, this should be :

```
 createPersonalAccountViewPager.adapter = CreatePersonalAccountViewPagerAdapter(
            childFragmentManager,
            titles
        )

```

app/src/main/kotlin/com/wire/android/feature/auth/registration/personal/email/CreatePersonalAccountEmailFragment.kt
```diff
@@ -0,0 +1,12 @@
+package com.wire.android.feature.auth.registration.personal.email
+
+import androidx.fragment.app.Fragment
+import com.wire.android.R
+
+class CreatePersonalAccountEmailFragment :
+    Fragment(R.layout.fragment_create_personal_account_email) {
```
BradleyWire, line 7: If the line length is too long, could be wrap the parameter? 

```
 Fragment(
    R.layout.fragment_create_personal_account_email
 )

```

I personally think it looks nicer than moving the whole` Fragment(...)` declaration onto another line.

app/src/main/res/layout/fragment_create_personal_account.xml
```diff
@@ -0,0 +1,44 @@
...
+    <androidx.constraintlayout.widget.Guideline
+        android:id="@+id/createPersonalAccountTopGuideline"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constraintGuide_percent="0.228" />
+
+    <androidx.appcompat.widget.AppCompatTextView
+        android:id="@+id/createPersonalAccountTitleTextView"
+        style="@style/TextView.Authentication.Header"
```
BradleyWire, line 18: Is this formatted? style attributes normally get auto-formatted to the top by default.

app/src/main/res/values/styles_textview.xml
```diff
@@ -27,4 +27,10 @@
         <item name="android:paddingEnd">@dimen/create_account_description_padding</item>
     </style>
 
+    <style name="TextView.Authentication.Header" parent="TextView.H2">
+        <item name="android:textColor">?android:textColorPrimary</item>
+        <item name="android:gravity">center_horizontal</item>
+        <item name="android:textSize">32sp</item>
```
BradleyWire, line 33: Is this header text size unique? None of the parent header styles H1 through to H6 cover this text size? I'm dubious about us having these concrete parent styles and overriding text sizes unless this is completely unique? I feel it'll take away from the consistency from a design perspective. Obviously, it's not easy when we are using old designs, so i'm not too concerned here.

gizemb: there's no pre-defined style that gives us this font + size + color combination. We have 34 dp in some of them, but then it has a different font & color. TextView.H2 was the one that I can extend with minimal changes. Unfortunately this doesn't resemble any other predefined styles, so I couldn't use my self-judgement here to change it :) instead I kept what's there on old design.

app/src/main/res/layout/fragment_create_personal_account.xml
gizemb: nope, this is the default format. it goes roughly: id, style, layout_width/height, others...

app/src/main/res/values/styles_textview.xml
BradleyWire: That's completely fair, that is one of the tradeoffs with replicating with an old design, i'll resolve this for now

app/src/main/kotlin/com/wire/android/feature/auth/registration/personal/email/CreatePersonalAccountEmailFragment.kt
gizemb: done. I better add detekt soon...

app/src/main/res/layout/fragment_create_personal_account.xml
BradleyWire: That's my bad, was looking at a file without android:id, ignore me.

app/src/main/kotlin/com/wire/android/core/extension/Activity.kt
```diff
@@ -10,4 +12,9 @@ fun AppCompatActivity.replaceFragment(frameId: Int, fragment: Fragment, addToBac
                 it.addToBackStack(fragment.tag)
             }
         }
-    }
\ No newline at end of file
+    }
+
+fun Activity.showKeyboard() {
```
android10, line 17: I'm a big fan of inlining in order to be more concise, except when it makes the code more difficult to read, but here is how it would look like:

```kotlin
fun Activity.showKeyboard()  = 
      (this.getSystemService(Activity.INPUT_METHOD_SERVICE) as InputMethodManager).toggleSoftInput(SHOW_IMPLICIT, HIDE_IMPLICIT_ONLY)
```

Then you need to import: ```InputMethodManager```

app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivity.kt
```diff
@@ -10,11 +10,22 @@ class CreateAccountActivity : AppCompatActivity(R.layout.activity_create_account
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        replaceFragment(R.id.createAccountLayoutContainer, CreateAccountFragment.newInstance(), false)
+        if (savedInstanceState == null) {
```
android10, line 13: Is this part of the official style? Because I feel like this is too verbose (java style). 
Something like this is what I see all the time over there, plus it is more readable and concise:

```kotlin
        if (savedInstanceState == null) {
            replaceFragment(R.id.createAccountLayoutContainer, CreateAccountFragment.newInstance(), false)
        }
        initBackButton()
```

Another alternative:
```kotlin
savedInstanceState ?: replaceFragment(R.id.createAccountLayoutContainer, CreateAccountFragment.newInstance(),false)

initBackButton()
```

Can we discuss all these style guidelines?

app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountFragment.kt
```diff
@@ -17,9 +18,9 @@ class CreateAccountFragment : Fragment(R.layout.fragment_create_account) {
 
     private fun initCreatePersonalAccount() {
         createPersonalAccountLayoutContainer.setOnClickListener {
-//            replaceFragment(
-//                R.id.activityCreateAccountLayoutContainer,
-//                CreatePersonalAccountFragment.newInstance())
+            replaceFragment(
```
android10, line 21: Same here with the style. If so, can you point me to where we got this from?

android10: As a rule of thumb, 140 characters per line is the standard.

android10: Also for brackets and expressions maybe it could be a good catch but for parameters, they should be in the same line.

I checked the Kotlin Lang official repository and I have not found anything like this. Here are 2 examples:
 - https://github.com/JetBrains/kotlin/blob/master/compiler/container/src/org/jetbrains/kotlin/container/Storage.kt
 - https://github.com/JetBrains/kotlin/blob/master/compiler/container/src/org/jetbrains/kotlin/container/Container.kt

app/src/test/kotlin/com/wire/android/feature/auth/registration/personal/CreatePersonalAccountViewPagerAdapterTest.kt
```diff
@@ -0,0 +1,46 @@
...
+        `when`(titles.size).thenReturn(TEST_SIZE)
+        assertEquals(TEST_SIZE, createPersonalAccountViewPagerAdapter.count)
+    }
+
+    @Test
+    fun `given getPageTitle(position) is called, then return the title to show for the given position`() {
+        `when`(titles[0]).thenReturn(TEST_TITLE_EMAIL)
+        `when`(titles[1]).thenReturn(TEST_TITLE_PHONE)
+
+        assertEquals(TEST_TITLE_EMAIL.toUpperCase(), createPersonalAccountViewPagerAdapter.getPageTitle(0))
```
android10, line 37: Can we use assertJ, it has a fluent api (you can concatenate assertions) plus more readable code. It already exists in the dependencies. AFAIK, this is something we discussed in the Collective. 

This would look like something like this:

```kotlin
assertThat(TEST_TITLE_EMAIL.toUpperCase()).isEqualTo(createPersonalAccountViewPagerAdapter.getPageTitle(0)))
```
Here is the doc:
https://joel-costigliola.github.io/assertj/

app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountFragment.kt
gizemb: > As a rule of thumb, 140 characters per line is the standard.

[Here](https://developer.android.com/kotlin/style-guide#line_wrapping) it says 100 characters, we decided to use 120 before, I updated it to 140, but seems like there's no true "standard"

> Same here with the style. If so, can you point me to where we got this from?

I got it from copy paste :) I think the main concern here was line character count. Since we agreed on 140 now, we can write all in one line.
Really didn't put much attention, will try to write in a more Kotlin-esque way in the future.

> Also for brackets and expressions maybe it could be a good catch but for parameters, they should be in the same line.
> 
Again, line character count comes into play here. we should think this thoroughly. so if not all the parameters fits in 140 characters, should I cut if off in the middle? Sth like this?

```
method(param1, param2, param3, |
    param4)                    |
```

app/src/test/kotlin/com/wire/android/feature/auth/registration/personal/CreatePersonalAccountViewPagerAdapterTest.kt
gizemb: hmm, i missed that part. done!

app/src/androidTest/kotlin/com/wire/android/feature/auth/registration/CreateAccountActivityUITest.kt
```diff
@@ -27,5 +27,21 @@ class CreateAccountActivityUITest : FunctionalTest(CreateAccountActivity::class.
...
+
+        onView(withText(R.string.create_personal_account_title)).check(matches(isDisplayed()))
+        onView(withText(R.string.authentication_tab_layout_title_email)).check(matches(isDisplayed()))
+        onView(withText(R.string.authentication_tab_layout_title_phone)).check(matches(isDisplayed()))
+
+        //TODO: check whether Email entry area is visible
+    }
+
+    @Test
+    fun personal_uiElementsVisibleInLandscape() = inRotatedMode {
```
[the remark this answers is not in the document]
gizemb: it just rotates the screen to the left and back, so it's not guaranteed to be in landscape. Hence I changed it to `rotateScreen`

app/src/main/kotlin/com/wire/android/feature/auth/registration/CreateAccountFragment.kt
android10: Thanks for the points Gizem. The only cosmetic change was the `params` since this is how the official langauge is doing it, which for me looked weird.

PR #11, opened 2020-06-16
Registration #5: Add e-mail entry layout

gizemb:
- Sets up dependency injection infrastructure
- Adds TextInputLayout style
- Adds circular button

**TODO:** Still needs updating cursor color on editable fields

Screenshot_1592298272

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -38,7 +38,6 @@ dependencies {
     implementation(Libraries.material)
     implementation(Libraries.livedataKtx)
     implementation(Libraries.Koin.androidCore)
-    implementation(Libraries.Koin.androidScope)
     implementation(Libraries.Koin.viewModel)
 
     // Unit/Android tests dependencies
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -23,6 +23,10 @@ android {
     }
 
     sourceSets { map { it.java.srcDir("src/${it.name}/kotlin") } }
+
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
 }
 
 dependencies {
@@ -32,6 +36,10 @@ dependencies {
     implementation(Libraries.ktxCore)
     implementation(Libraries.constraintLayout)
     implementation(Libraries.material)
+    implementation(Libraries.livedataKtx)
+    implementation(Libraries.Koin.androidCore)
+    implementation(Libraries.Koin.androidScope)
+    implementation(Libraries.Koin.viewModel)
 
     // Unit/Android tests dependencies
     testImplementation(TestLibraries.junit4)
--- a/app/src/main/kotlin/com/wire/android/core/di/Injector.kt
+++ b/app/src/main/kotlin/com/wire/android/core/di/Injector.kt
@@ -25,11 +25,7 @@ object Injector {
     fun start(context: Context) {
         startKoin {
             androidContext(context)
-            modules(listOf(
-                coreModules,
-                sharedModules,
-                featureModules
-            ).flatten())
+            modules(listOf(coreModules, sharedModules, featureModules).flatten())
         }
     }
 }
--- a/app/src/main/kotlin/com/wire/android/core/di/Injector.kt
+++ b/app/src/main/kotlin/com/wire/android/core/di/Injector.kt
@@ -1,0 +1,35 @@
+package com.wire.android.core.di
+
+import android.content.Context
+import com.wire.android.feature.auth.registration.di.registrationModule
+import org.koin.android.ext.koin.androidContext
+import org.koin.core.context.startKoin
+import org.koin.core.module.Module
+
+object Injector {
+
+    private val coreModules: List<Module> = listOf()
+
+    /**
+     * Shared modules should contain dependencies that can
+     * build up multiple features
+     */
+    private val sharedModules: List<Module> = listOf()
+
+    /**
+     * Feature modules should contain dependencies that build up specific
+     * features and don't tend to live outside of that feature
+     */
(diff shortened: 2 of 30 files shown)


app/build.gradle.kts
```diff
@@ -23,6 +23,10 @@ android {
     }
 
     sourceSets { map { it.java.srcDir("src/${it.name}/kotlin") } }
+
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
```
gizemb, line 29: @android10 is there a better place to put this? I couldn't put it under `compilation.gradle.kts` because it requires kotlin plugin.

BradleyWire: Could we not add the kotlin plugin to the `buildSrc/build.gradle.kts` dependencies to allow for this to be used within the scripts? (also could you make the `Dependencies` object private if we do decide to add this there inside that `build.gradle`)

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -26,13 +26,22 @@ object Libraries {
...
     const val kotlinStdLib     = "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
     const val appCompat        = "androidx.appcompat:appcompat:${Versions.jetpack}"
     const val constraintLayout = "androidx.constraintlayout:constraintlayout:${Versions.constraintLayout}"
     const val ktxCore          = "androidx.core:core-ktx:${Versions.ktx}"
     const val material         = "com.google.android.material:material:${Versions.material}"
+    const val livedataKtx     = "androidx.lifecycle:lifecycle-livedata-ktx:${Versions.lifecycleKtx}"
+
+    object Koin {
+        const val androidCore  = "org.koin:koin-android:${Versions.koin}"
+        const val androidScope = "org.koin:koin-android-scope:${Versions.koin}"
```
mejdoo, line 42: Do we wanna do Koin scoping in this MVP? if not, this line is not needed.

app/build.gradle.kts
gizemb: I need some guidance from Fernando on this. I couldn't even create a .gradle.kts file :D

buildSrc/src/main/kotlin/Dependencies.kt
gizemb: hmm, yes. I am trying to avoid it until I really need it. I deleted this one. good catch 👍

app/build.gradle.kts
android10: @Gizem I would not start mixing things up. AFAIK we do not need this right? By default it compiles against Java 8. 
In case you need it, maybe I can raise a PR tomorrow and provide more context but I would extract all this and keep the ```build.gradle.kts``` as small and clean possible. 

@BradleyWire I do not understand your point. "Dependencies" cannot be private. ```buildSrc``` is a separated project that builds stuff up, which becomes available to the rest of gradle scripts.

app/src/main/kotlin/com/wire/android/core/di/Injector.kt
```diff
@@ -0,0 +1,35 @@
...
+    private val sharedModules: List<Module> = listOf()
+
+    /**
+     * Feature modules should contain dependencies that build up specific
+     * features and don't tend to live outside of that feature
+     */
+    private val featureModules: List<Module> = listOf(registrationModule)
+
+    fun start(context: Context) {
+        startKoin {
```
android10, line 26: As I pointed out in the other PR with the styles: Brackets {} are fine to keep them this way in order to favor the readability of the expression, but the arguments...mmm...I do not see it and I think we are not writing concise Kotlin code. This is how it would look for me:

```kotlin
        startKoin {
            androidContext(context)
            modules(listOf(coreModules, sharedModules, featureModules).flatten())
        }
 ```

I would also be of the idea of extracting the list of modules to a constant.

File: app/build.gradle.kts (hunk 1 above)
Comment (BradleyWire) [reply app/build.gradle.kts:29]: @android10 `Dependencies` is the object inside your `build.gradle.kts` inside the `buildSrc` module, it doesn't need to have public access seeing as it's only constant belongs in the same file as it's own 

```
object Dependencies {
    const val AndroidBuildTools = "com.android.tools.build:gradle:4.0.0"
}
```

I'd be safe in assuming this only need to be private to the file. It's a minor cosmetic change and is out of the scope of this PR which is why I asked to only change it if you're adding a dependency change to the gradle file.

android10: Ok. I thought you were talking about something else. 
Here is the fix:
https://github.com/wireapp/wire-android-reloaded/pull/12/files

app/src/main/kotlin/com/wire/android/core/di/Injector.kt
gizemb: combined them into a single line. I didn't feel the need to create a new variable for the module list. If we have one more in the future, I can add it.

Gizem:
Hi i have no idea why i am in :) 
Could you please delete my email
Thank you


Gizem Ergun

app/build.gradle.kts
BradleyWire: Thanks @android10, Also, you'll need the `kotlinOptions` closure eventually otherwise you'll run into:

> Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6

I've ran into with this previously with the `observe(..)` extension function when observing a `LiveData` instance (which is heavily used in our architecture)

On the topic of where it belongs though, to me I understand the point of keeping the scripts concise and simple, but, It'd also make sense to have all the options for compilation in the same script to keep in line with single responsibility principle.

PR #217, opened 2021-05-17
Feature: receive instant text message - part 0 - implement WebSocket

ohassine:
I raised this draft PR to get your feedback/suggestions about websocket implementation


mejdoo (commented):
I wrote those comment few days ago but I forget to publish them 😅 The web socket implementation needs further investigations from my side, @ohassine  could you please specify in which thing you are stuck and what are your questions regarding the web socket ?

gizemb (changes requested):
I had a few immediate change requests.


Overall opinon:

I think there are too many intertwined components which can be reduced.

My first question is: what is EventHandler designed for? Is it designed to reflect the Events on the UI side? That's what I understand from that "subscribe" call in the ViewModel. 

We actually have more scenarios that requires websocket connection:
1- App is opened and comes into foreground
2- App is killed and we receive a notification via FCM. Then the app is awakened by O.S. and we can start websocket to query the actual event.
3- The user enabled "Maintain connection to server" in Settings. Then the websocket we opened at 1st case remains open.

In Scala code base, all 3 scenarios create some changes in the database immediately. All the data is observed from the UI via Signals, which are equivalent to Flows. So I feel like if we're going to expose Flows from repositories & use cases, we shouldn't need to subscribe to EventsHandler in ViewModels. If we're not gonna use Flows in use cases, then EventHandler could make sense, but then maybe it shouldn't start the websocket itself, only subscribe to it.

So, in current structure EventHandler only handles the 1st case for now. Do we plan to extend it to handle 2nd or 3rd cases, or will there be more components?

PR #260, opened 2021-09-21
WIP: Consider Locale when formatting date and time

vitorhugods:
# Work In Progress

## The problem

Since flying to Brazil some weeks ago, I've been having issues with one test that was asserting `UTC` Timezone and `US` Locale.

My system is currently at `BRT` timezone and the language is in German. So this test was **always** failing.

We should mock/set locale when testing this.

## The solution

I've made sure all of our date formatting functions take the system locale into consideration and added a helping function that executes a block using a given Locale and returns it back to default when testing this.

### The remaining problem

Different JVM have different implementations of localization and date-time formatting.
For the same date and Locale:
On OpenJDK 8 : `20 de Julho de 2021 7h10min59s BRT`
On OpenJDK 11:   `20 de julho de 2021 07:10:59 BRT`

## Observations

Ideally, I think we should have a home-coocked `DateTimeFormatter` that takes a `UserLocaleProvider` as dependency (so we can override system locale based on user preferences). And use this home-coocked `DateTimeFormatter` instead of static functions.


vitorhugods:
build again

mejdoo (approved):

vitorhugods:
build again

PR #261, opened 2021-09-22
Fetch clients of users

vitorhugods:
## What's new?

We can now fetch clients of users using their qualified user id.

## Implementation

I'm decided to use the new endpoints which use take the Qualified ID for it.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/feature/auth/client/datasource/remote/ClientRemoteDataSource.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/auth/client/datasource/remote/ClientRemoteDataSource.kt
@@ -46,6 +46,6 @@ class ClientRemoteDataSource(
     ): Either<Failure, Map<QualifiedId, List<String>>> = request {
         val dtos = userIds.map(clientRemoteMapper::fromQualifiedIdToQualifiedIdDTO)
         clientApi.clientsOfUsers(authorizationToken, ClientsOfUsersRequest(dtos))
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/feature/auth/client/datasource/remote/ClientRemoteDataSource.kt
```diff
@@ -35,4 +39,13 @@ class ClientRemoteDataSource(
     ): Either<Failure, Unit> = request {
         clientApi.updatePreKeys(authorizationToken, clientId, updatePreKeysRequest)
     }
+
+    suspend fun clientIdsOfUsers(
+        authorizationToken: String,
+        userIds: List<QualifiedId>
+    ): Either<Failure, Map<QualifiedId, List<String>>> = request {
+        val dtos = userIds.map(clientRemoteMapper::fromQualifiedIdToQualifiedIdDTO)
```
mejdoo, line 47: What the meaning of `dtos` here?

mejdoo (approved):

app/src/main/kotlin/com/wire/android/feature/auth/client/datasource/remote/ClientRemoteDataSource.kt
vitorhugods: Data Transfer Object.

Used to represent a data structure that is commonly known between different applications. Like "Api Model".

PR #262, opened 2021-09-24
CryptoBoxClient Update: QualifiedId and easier utilization without prekeys

vitorhugods:
# UserId vs QualifiedId

### The problem
`CryptoBoxClient` is using `userId` (String), instead of the new `QualifiedId`. Better to fix this now than re-factor later in the future.

### The fix 
I changed the internals, replacing `userId: String` with `userId: QualifiedId`.
I've also identified some touching points (like when dealing with user-specific `SharedPreferences` and `MessageEvent`, for example) that remain using `userId`, but goes too far out of the scope of this PR.

# Encrypt messages without PreKeys

### The problem
It is currently impossible to make sure a session exists without calling the `assertSession`. And that requires a `PreKey`, something the sender may not have yet, and may not even need, if the session is already established.

### The fix
I renamed `assertSession` to `createSessionIfNeeded` (@gizemb was right, the old name sucks).
I added a function `doesSessionExists` that enables us to easily check if a session already exists. We can fetch PreKeys only if the session doesn't exist.

### DIFF
--- a/app/src/androidTest/kotlin/com/wire/android/core/crypto/CryptoBoxClientIntegrationTest.kt
+++ b/app/src/androidTest/kotlin/com/wire/android/core/crypto/CryptoBoxClientIntegrationTest.kt
@@ -8,12 +8,12 @@ import com.wire.android.core.crypto.model.CryptoClientId
 import com.wire.android.core.crypto.model.CryptoSessionId
 import com.wire.android.core.crypto.model.EncryptedMessage
 import com.wire.android.core.crypto.model.PlainMessage
-import com.wire.android.core.crypto.model.UserId
 import com.wire.android.core.exception.MessageAlreadyDecrypted
 import com.wire.android.core.exception.SessionNotFound
 import com.wire.android.core.functional.Either
 import com.wire.android.framework.functional.shouldFail
 import com.wire.android.framework.functional.shouldSucceed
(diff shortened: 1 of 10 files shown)


app/src/androidTest/kotlin/com/wire/android/core/crypto/CryptoBoxClientIntegrationTest.kt
```diff
@@ -60,11 +60,30 @@ class CryptoBoxClientIntegrationTest : InstrumentationTest() {
...
+        bobClient.createSessionIfNeeded(aliceSessionId, aliceKey)
+
+        bobClient.doesSessionExists(aliceSessionId)
+            .shouldSucceed {
+                it shouldBeEqualTo true
+            }
+    }
+
     @Test
     fun givenBobWantsToTalkToAlice_whenSendingTheFirstMessageAndTheSessionIsAsserted_itShouldBeEncryptedSuccessfully() {
```
ohassine, line 83: It should be changed to` whenSendingTheFirstMessageAndTheSessionIsCreated`, right ?

vitorhugods: Yes, and welcome back!

I'll tackle that in another PR.

ohassine (approved):

PR #264, opened 2021-10-04
Refactor: Provide more control over network requests

vitorhugods:
## Current Issues

1. The method `rawRequest` is currently very far from "raw", as it calls `performRequest` internally, handling all the HTTP errors and does not allow us to intercept the HTTP result at all.
It is only "raw" when the request succeeds, returning the response.

2. The `default` argument in the method `request` is too strong.
Maybe I just want a different `default` depending on the error `response`?

## Solution

Change `performRequest` to do the bare minimal: check with `NetworkHandler` for connection and perform the request. Handle possible network exceptions.

Simplify `rawRequest` and add an option for it to take a different  `onResponseError` lambda, which allows us to intercept the request on error as well. By default, it goes with the default behaviour of analysing the HTTP Code and mapping to it to a `Failure`.

Replace `default` argument in `request` for a lambda that enables us to inspect the result before returning anything.

### DIFF
--- a/app/src/test/kotlin/com/wire/android/core/network/ApiServiceTest.kt
+++ b/app/src/test/kotlin/com/wire/android/core/network/ApiServiceTest.kt
@@ -1,7 +1,16 @@
 package com.wire.android.core.network
 
 import com.wire.android.UnitTest
-import com.wire.android.core.exception.*
+import com.wire.android.core.exception.BadRequest
+import com.wire.android.core.exception.EmptyResponseBody
+import com.wire.android.core.exception.Failure
+import com.wire.android.core.exception.Forbidden
+import com.wire.android.core.exception.InternalServerError
+import com.wire.android.core.exception.NetworkConnection
+import com.wire.android.core.exception.NotFound
+import com.wire.android.core.exception.ServerError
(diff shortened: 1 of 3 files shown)


app/src/test/kotlin/com/wire/android/core/network/ApiServiceTest.kt
```diff
@@ -1,16 +1,8 @@
...
-import com.wire.android.core.exception.EmptyResponseBody
-import com.wire.android.core.exception.Failure
-import com.wire.android.core.exception.Forbidden
-import com.wire.android.core.exception.InternalServerError
-import com.wire.android.core.exception.NetworkConnection
-import com.wire.android.core.exception.NotFound
-import com.wire.android.core.exception.ServerError
-import com.wire.android.core.exception.TooManyRequests
-import com.wire.android.core.exception.Unauthorized
+import com.wire.android.core.exception.*
```
ohassine, line 4: As I remember wildcard import is not allowed

vitorhugods: Forgot to bring IDE settings to new setup 😅 

Done :)

ohassine (approved):

PR #275, opened 2021-10-13
update preKeys

ohassine:
This PR aims to generate new prekeys if needed

We need to generate new prekeys when the remaining ids are less than the threshold


app/src/main/kotlin/com/wire/android/core/crypto/CryptoBoxClient.kt
```diff
@@ -46,7 +46,25 @@ class CryptoBoxClient(
         }
 
     fun createNewPreKeysIfNeeded(remainingPreKeysIds: List<Int>): Either<Failure, List<PreKey>> = useBox {
-        TODO("Handled in another PR")
+        val remainingIds = remainingPreKeysIds.filter { it <= CryptoBox.MAX_PREKEY_ID }
+        val maxId = remainingIds.maxOrNull() ?: -1
+        val lastPreKeyId = clientPropertyStorage.lastPreKeyId(UserId(userId.id)) ?: -1
```
vitorhugods, line 51: Can we move this line to inside the `if` statement? So we avoid fetching values if they are not going to be used.

vitorhugods (approved):
Approved with a minor change request.

PR #288, opened 2021-10-29
Send messages: Part 14 - SEND IT!

vitorhugods:
Contains changes from #295 
> Sometimes you just gotta lick the stamp and send it 👅📮

### Core logic for sending a message

`MessageSender` coordinates handles the sending. `SendMessageService` will call it directly or schedule a `AndroidSendMessageWorker` if there's no connection right now.

1. It will get the recipients by using `OutgoingMessageRecipientsRetriever`, which will get all the contacts and clients that belong to a conversation, while creating missing Crypto Sessions if needed.
2. For each client a message is encrypted and added to an envelope. This envelope is sent!
3. In case of failure because clients or contacts are missing, `MessageSendFailureHandler` comes to rescue, sorts things out. `MessageSender` will try again by going back to the first step.


app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    private suspend fun createOutgoingEnvelope(
+        detailedContacts: List<DetailedContact>,
+        senderClientId: String,
+        senderUserId: String,
+        message: Message
+    ): ChatMessageEnvelope {
+        val entries = detailedContacts.fold(mutableListOf<RecipientEntry>()) { acc, detailedContact ->
```
ohassine, line 89: acc refers to accumulator ?

app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.wire.android.feature.conversation.content.SendMessageFailure
+import com.wire.android.feature.messaging.ChatMessageEnvelope
+import com.wire.android.feature.messaging.ClientPayload
+import com.wire.android.feature.messaging.RecipientEntry
+import com.wire.android.shared.session.SessionRepository
+import kotlinx.coroutines.ObsoleteCoroutinesApi
+import kotlinx.coroutines.newSingleThreadContext
+import kotlinx.coroutines.withContext
+
+class MessageSender(
```
ohassine, line 21: This class has 2 responsibilities. It sends message and creates envelope
Why not to extract the envelope creation method to a separate class

app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
vitorhugods: Yes. Renamed.

app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
vitorhugods: It could be split...

I have been fantasising about this in my head for a while. The main drawback at the moment of doing something like that is growing even further the list of dependencies. `MessageSender` already has 5 dependencies, and at least one more is needed for the single thread coroutine context.

Maybe `networkHandler` can be removed. As `trySendingOutgoingMessage` is only supposed to be called:
1. When `SendMessageService` calls it (after checking for internet connection).
2. When WorkManager launches `AndroidSendMessageWorker`.

Also, it is expected that a network failure can happen during the sending process. So it doesn't make much sense to check for it inside `MessageSender`.

vitorhugods: With all of that in mind. My propose is:

1. Remove `networkHandler` from the class and the checking for internet connection, as it is very redundant.
2. Move the `Envelope` creation to another specialised class.

ohassine: Yeah, I think It would be better like this

ohassine (approved):

app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
vitorhugods: Done! Check #295 for the envelope creator. Let's merge that before this one.

app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val messageRepository: MessageRepository,
+    private val sessionRepository: SessionRepository,
+    private val messageSendFailureHandler: MessageSendFailureHandler,
+    private val outgoingMessageRecipientsRetriever: OutgoingMessageRecipientsRetriever,
+    private val messageEnvelopeCreator: MessageEnvelopeCreator,
+    dispatcherProvider: DispatcherProvider
+) {
+
+    private val singleThreadedContext = dispatcherProvider.newSingleThreadedDispatcher(WORK_POOL_NAME)
```
ohassine, line 25: For curiosity, why are you using newSingleThreadedDispatcher?

ohassine (approved):

app/src/main/kotlin/com/wire/android/feature/conversation/content/domain/MessageSender.kt
vitorhugods: To make sure messages are sent in order one at a time and no duplicated work is done.

If, by any reason we have multiple calls to send a message, it may be that the order is not preserved or that we fetch multiple preKeys for the same device and try to establish multiple sessinos for it without any need.

The `singleThreadDispatcher` will make sure we process one by one.

ohassine: Thank you 👌

PR #290, opened 2021-10-31
Send messages: Part 11 - Handle message send failure

vitorhugods:
## New

Handler for when sending of message fails.

If there are contacts or clients that were added, removed or are redundant to a conversation, this handler should take action and do what's needed to fix it and give sending another shot.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/feature/conversation/content/usecase/MessageSendFailureHandler.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/conversation/content/usecase/MessageSendFailureHandler.kt
@@ -1,0 +1,26 @@
+package com.wire.android.feature.conversation.content.usecase
+
+import com.wire.android.core.exception.Failure
+import com.wire.android.core.functional.Either
+import com.wire.android.core.functional.suspending
+import com.wire.android.feature.contact.ContactClient
+import com.wire.android.feature.contact.ContactRepository
+import com.wire.android.feature.conversation.content.SendMessageFailure
+
+class MessageSendFailureHandler(private val contactRepository: ContactRepository) {
+    /**
+     * Handle a failure when attempting to send a message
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/wire/android/feature/conversation/content/usecase/MessageSendFailureHandler.kt
```diff
@@ -0,0 +1,26 @@
+package com.wire.android.feature.conversation.content.usecase
```
ohassine, line 1: I think this class shouldn't be here as it's not a usecase

vitorhugods: What about renaming the package to `domain`?

And inside we can have another package called `usecase` ?

Something like this:

```shell
|com
  |wire
    |android
      |feature
        |conversation
          |content
            |domain
              MessageSendFailureHandler.kt
              MessageSender.kt
              |usecase
                SendTextMessageUseCase.kt
                SendTextMessageUseCase.kt
```

ohassine: Yes it's better, we could also move data classes there also

vitorhugods: Done

ohassine (approved):

app/src/main/kotlin/com/wire/android/feature/conversation/content/usecase/MessageSendFailureHandler.kt
ohassine: Maybe we should do the same for other packages for consistency

PR #293, opened 2021-11-02
Send messages: Part 16 (Final) - Send messages UI

vitorhugods:
## New

User interface for sending a text message.

Super simple, just enough so we can get started with internal testing.

https://user-images.githubusercontent.com/9389043/139818693-44a50ed1-778e-45a9-bdf1-0cfde1490029.mp4


ohassine (approved):
And the last PR 👌 
Well done ! 🚀

PR #294, opened 2021-11-02
fix: app build error on m1 macbook

MohamadJaara:
forcing room to use the latest sqlite jdbc driver (for compatibility issues with m1 macs)

- This should be fixed in the upcoming room version 2.4.0


vitorhugods (approved):
First PR! 🍾 🎉 🚀

PR #296, opened 2021-11-05
Refactor: qualified conversation id

MohamadJaara:
Using the new conversaiton id (value, domain) pair

### DIFF
--- a/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
@@ -1,3 +1,9 @@
 package com.wire.android.feature.conversation
 
-data class Conversation(val id: String, val name: String? = null, val type: ConversationType)
+data class Conversation(
+    val id: ConversationID,
+    val name: String? = null,
+    val type: ConversationType
+)
+
+data class ConversationID(val value: String, val domain: String)
--- a/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
@@ -1,7 +1,5 @@
 package com.wire.android.feature.conversation
 
-import com.wire.android.core.extension.EMPTY
-
 data class Conversation(
     val id: ConversationID,
     val name: String? = null,
--- a/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
@@ -1,8 +1,6 @@
 package com.wire.android.feature.conversation
 
-import android.os.Parcelable
 import com.wire.android.core.extension.EMPTY
-import kotlinx.android.parcel.Parcelize
 
 data class Conversation(
     val id: ConversationID,
@@ -10,10 +8,4 @@ data class Conversation(
     val type: ConversationType
 )
 
-@Parcelize
-data class ConversationID(val value: String, val domain: String) : Parcelable {
-    companion object {
-        fun blankID() = ConversationID(String.EMPTY, String.EMPTY)
-    }
-}
-
+data class ConversationID(val value: String, val domain: String)
--- a/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
@@ -1,9 +1,19 @@
 package com.wire.android.feature.conversation
 
+import android.os.Parcelable
+import com.wire.android.core.extension.EMPTY
+import kotlinx.android.parcel.Parcelize
+
 data class Conversation(
     val id: ConversationID,
     val name: String? = null,
     val type: ConversationType
 )
 
-data class ConversationID(val value: String, val domain: String)
+@Parcelize
+data class ConversationID(val value: String, val domain: String) : Parcelable {
+    companion object {
+        fun blankID() = ConversationID(String.EMPTY, String.EMPTY)
+    }
+}
+
--- a/app/src/main/kotlin/com/wire/android/feature/conversation/content/ui/ConversationActivity.kt
+++ b/app/src/main/kotlin/com/wire/android/feature/conversation/content/ui/ConversationActivity.kt
@@ -11,19 +11,25 @@ import org.koin.android.viewmodel.ext.android.viewModel
 
 class ConversationActivity : AppCompatActivity(R.layout.activity_conversation) {
 
-    private val conversationId get() = intent.getParcelableExtra<ConversationID>(ARG_CONVERSATION_ID)
+    private var _conversationId: ConversationID? = null
+    private val conversationId get() = _conversationId!!
     private val conversationTitle get() = intent.getStringExtra(ARG_CONVERSATION_TITLE)
     private val viewModel by viewModel<ConversationViewModel>()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        val conversationIdValue = intent.getStringExtra(ARG_CONVERSATION_ID)
+        val conversationDomain = intent.getStringExtra(ARG_CONVERSATION_DOMAIN)
+        if (conversationDomain != null && conversationIdValue != null) {
+            _conversationId = ConversationID(conversationIdValue, conversationDomain)
+        }
         setUpConversationTitle()
         setUpBackNavigation()
         cacheConversationId()
     }
 
     private fun cacheConversationId() {
-        conversationId?.let { viewModel.cacheConversationId(it) }
+        viewModel.cacheConversationId(conversationId)
     }
 
     private fun setUpBackNavigation() {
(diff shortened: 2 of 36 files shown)


vitorhugods:
I think your IDE is set to hard wrap at 100 chars, maybe?
We tolerate longer lines than that 😅.

Detekt will screm at you if lines get bigger than 140 chars, tho.

For next PRs in this project, you can configure that:
`Preferences -> Editor -> Code Style -> Kotlin -> Wrapping and Braces -> Hard wrap at: 140`

Maybe we could add a [`.editorconfig`](https://editorconfig.org/) file to the project in the near future.

app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
```diff
@@ -1,3 +1,19 @@
...
+import kotlinx.android.parcel.Parcelize
+
+data class Conversation(
+    val id: ConversationID,
+    val name: String? = null,
+    val type: ConversationType
+)
+
+@Parcelize
+data class ConversationID(val value: String, val domain: String) : Parcelable {
```
vitorhugods, line 14: Core models should not depend on `android.os` things.

If this is sused for Bundles, perhaps we should deconstruct the ID and actually put `value` and `domain` separately in the bundle.

app/src/main/kotlin/com/wire/android/feature/conversation/content/ui/ConversationActivity.kt
```diff
@@ -36,7 +37,7 @@ class ConversationActivity : AppCompatActivity(R.layout.activity_conversation) {
     }
 
     companion object {
-        fun newIntent(context: Context, conversationId: String, conversationTitle: String) =
+        fun newIntent(context: Context, conversationId: ConversationID, conversationTitle: String) =
```
vitorhugods, line 40: Here the `ConversationID` can be mapped to two different bundle extras, to avoid having `android` dependencies on domain modules.

app/src/main/kotlin/com/wire/android/feature/conversation/content/ui/ConversationFragment.kt
```diff
@@ -59,6 +59,6 @@ class ConversationFragment : Fragment(R.layout.fragment_conversation) {
 
     override fun onPause() {
         super.onPause()
-        viewModel.updateCurrentConversationId(String.EMPTY)
+        viewModel.updateCurrentConversationId(ConversationID.blankID())
```
vitorhugods, line 62: _Maybe_ when it was a String it made some sense to do this way. But at this point, I guess making the `currentConversationId` nullable is just better.

Using `String.EMPTY` was arguably worse than making it nullable anyway. Kotlin has null safety, and it's awesome that we can use it.

If we don't have a current conversation ID, we can just make it `null`. That's not ugly, that's easy to understand.

If I needed to use the `currentOpenedConversationId` function, and I wanted to know if there was any conversation ID, I would probably check for nullability, left side of an `Either`, `String.EMPTY` or `ConversationID.blankID()` are not discoverable.

vitorhugods (changes requested):
Overall it looks great, just a couple of points that we could change for future-proofing.

1. Make the `currentConversationId` more discoverable when it's blank/non-existing.
2. Avoid depending on `android` for business models. Especially having a try-out with KMM in mind.

app/src/test/kotlin/com/wire/android/feature/conversation/content/ui/ConversationViewModelTest.kt
```diff
@@ -91,7 +97,17 @@ class ConversationViewModelTest : UnitTest() {
         coVerify(exactly = 1) { sendTextMessageUseCase.run(SendTextMessageUseCaseParams(TEST_CONVERSATION_ID, textMessage)) }
     }
 
+    @Test
+    fun `given is current conversation id rested, when calling resetCurrentConversationId, then the use case is called`() {
+        conversationViewModel.cacheConversationId(TEST_CONVERSATION_ID)
+
+        conversationViewModel.resetCurrentConversationId()
+
+    }
```
vitorhugods, line 106: Incomplete test, missing assertions.

app/src/main/kotlin/com/wire/android/feature/conversation/usecase/ResetCurrentConversationIdUseCase.kt
```diff
@@ -0,0 +1,12 @@
+package com.wire.android.feature.conversation.usecase
+
+import com.wire.android.core.exception.Failure
+import com.wire.android.core.functional.Either
+import com.wire.android.core.usecase.UseCase
+import com.wire.android.feature.conversation.data.ConversationRepository
+
+class ResetCurrentConversationIdUseCase(private val conversationRepository: ConversationRepository) :
```
vitorhugods, line 8: Missing tests for this class.

Should at least test that:

* `given the repository succeeds, when reseting the current conversation id, then the success is propagated`
* `given the repository fais, when reseting the current conversation id, then the failure is propagated`
* `given any parameters, when reseting the current conversation id, then repository is used`

vitorhugods (changes requested):
Just missing a few tests

app/src/main/kotlin/com/wire/android/feature/conversation/Conversation.kt
```diff
@@ -1,3 +1,11 @@
 package com.wire.android.feature.conversation
 
-data class Conversation(val id: String, val name: String? = null, val type: ConversationType)
+import com.wire.android.core.extension.EMPTY
```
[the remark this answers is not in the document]
MohamadJaara: strang since it passed detekt even that NoUnusedImports is active.
I will look into detekt configration

app/src/test/kotlin/com/wire/android/feature/conversation/content/ui/ConversationViewModelTest.kt
MohamadJaara: Opps

app/src/test/kotlin/com/wire/android/feature/conversation/usecase/ResetCurrentConversationIdUseCaseTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        coVerify(exactly = 1) { conversationRepository.restCurrentConversationId() }
+    }
+
+    @Test
+    fun `given the repository fails, when resetting the current conversation id, then the failure is propagate`() = runBlocking {
+        coEvery { conversationRepository.restCurrentConversationId() } returns Either.Left(EmptyCacheFailure)
+
+        val result = resetCurrentConversationIdUseCase.run(Unit)
+        assert(true) {result == Either.Left(EmptyCacheFailure)}
+
```
vitorhugods, line 43: On top of the extension for assertion, I suggest mocking the failure. If `EmptyCacheFailure` implementation changes and it becomes something other than an object, it would break these tests. Also, you guarantee that whatever the repository is throwing is being propagated exactly as it is.

```suggestion
        val failure = mockk<Failure>()
        coEvery { conversationRepository.restCurrentConversationId() } returns Either.Left(failure)

        val result = resetCurrentConversationIdUseCase.run(Unit)

        result.shouldFail {
            it shouldBeEqualTo failure
        }
```

app/src/test/kotlin/com/wire/android/feature/conversation/usecase/ResetCurrentConversationIdUseCaseTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        resetCurrentConversationIdUseCase = ResetCurrentConversationIdUseCase(conversationRepository)
+    }
+
+    @Test
+    fun `given the repository succeeds when resetting the current conversation id, then the success is propagated`() = runBlocking {
+        coEvery { conversationRepository.restCurrentConversationId() } returns Either.Right(Unit)
+
+        val result = resetCurrentConversationIdUseCase.run(Unit)
+        assert(true) {result == Either.Right(Unit)}
+
```
vitorhugods, line 33:
```suggestion
        result.shouldSucceed {}
```

app/src/test/kotlin/com/wire/android/feature/conversation/usecase/ResetCurrentConversationIdUseCaseTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.wire.android.core.exception.EmptyCacheFailure
+import com.wire.android.core.functional.Either
+import com.wire.android.feature.conversation.data.ConversationRepository
+import io.mockk.coEvery
+import io.mockk.coVerify
+import io.mockk.impl.annotations.MockK
+import kotlinx.coroutines.runBlocking
+import org.junit.Before
+import org.junit.Test
+import org.mockito.ArgumentMatchers.any
```
vitorhugods, line 13: If you take the suggestions below, these imports are needed:

```suggestion
import com.wire.android.UnitTest
import com.wire.android.core.exception.Failure
import com.wire.android.core.functional.Either
import com.wire.android.feature.conversation.data.ConversationRepository
import com.wire.android.framework.functional.shouldFail
import com.wire.android.framework.functional.shouldSucceed
import io.mockk.coEvery
import io.mockk.coVerify
import io.mockk.impl.annotations.MockK
import io.mockk.mockk
import kotlinx.coroutines.runBlocking
import org.amshove.kluent.shouldBeEqualTo
import org.junit.Before
import org.junit.Test
import org.mockito.ArgumentMatchers.any
```

vitorhugods (changes requested):
A small improvement on testing failures, and suggestion to use some sugary extensions we've created for testing `Either`:

`shouldFail` and `shouldSucceed`

vitorhugods (approved):

PR #302, opened 2022-01-13
feat: add Jetpack Compose libs to the project

borichellow:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Added Jetpack Compose lib to the project.

### Causes (Optional)

We'll need it in the project.

----

----

----
##### References

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -18,13 +18,26 @@
         <activity
             android:name=".LauncherActivity"
             android:screenOrientation="portrait"
-            android:theme="@style/AppTheme.Dark">
+            android:theme="@style/AppTheme.Dark"
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+
+        <activity
+            android:name = "androidx.test.core.app.InstrumentationActivityInvoker$EmptyFloatingActivity"
+            android:exported="true" />
+
+        <activity
+            android:name = "androidx.test.core.app.InstrumentationActivityInvoker$EmptyActivity"
+            android:exported="true" />
+
+        <activity
+            android:name = "androidx.test.core.app.InstrumentationActivityInvoker$BootstrapActivity"
+            android:exported="true" />
 
     </application>
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -27,18 +27,6 @@
             </intent-filter>
         </activity>
 
-        <activity
-            android:name = "androidx.test.core.app.InstrumentationActivityInvoker$EmptyFloatingActivity"
-            android:exported="true" />
-
-        <activity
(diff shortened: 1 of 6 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -18,14 +18,27 @@
...
+            android:theme="@style/AppTheme.Dark"
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
 
+        <activity
```
saleniuk, line 30: I think instead of copying all these activities into our manifest, it's better to just update these test libraries (newest versions already have `android:exported="true"`; in my case I needed to upgrade `androidx.test:core`, `androidx.test:runner`, `androidx.test:rules`, `androidx.fragment:fragment-testing` - all to `1.4.0` and espresso to `3.4.0`)

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -89,6 +90,12 @@ object Libraries {
     const val scarletLifecycle =  "com.tinder.scarlet:lifecycle-android:${Versions.scarlet}"
     const val scarletGson      =  "com.tinder.scarlet:message-adapter-gson:${Versions.scarlet}"
     const val fragment         =  "androidx.fragment:fragment:${Versions.fragment}"
+    const val composeUi        =  "androidx.compose.ui:ui:${Versions.compose}"
+    const val composeMaterial  =  "androidx.compose.material:material:${Versions.compose}"
+    const val composeTooling   =  "androidx.compose.ui:ui-tooling:${Versions.compose}"
+    const val composeActivity  =  "androidx.activity:activity-compose:1.4.0"
+    const val composeNavigation=  "androidx.navigation:navigation-compose:2.4.0-beta02"
```
saleniuk, line 97: I think the convention is to keep all version numbers separate in the `object Versions` 🤔

gongracr (dismissed):
LGTM 🍪

buildSrc/src/main/kotlin/Dependencies.kt
borichellow: good point

app/src/main/AndroidManifest.xml
borichellow: thanks for the idea, fixed

buildSrc/src/main/kotlin/Dependencies.kt
vitorhugods: Yes, that's because some versions are applied to multiple dependencies, like `kotlinx-coroutines-core` and `kotlinx-coroutines-android`. So having a single version declared changes the version to all the related dependencies.

The idea is having everything under `Versions` so we have everything in the same place.

vitorhugods (approved):
Just a tiny change, to remove duplication of `jvmTarget` setting.

Other than that, looks good!

gongracr (approved):

vitorhugods (approved):

PR #312, opened 2022-01-19
feat: display_mention_items (AR-807)

trOnk12:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-807" title="AR-807" target="_blank">StoryAR-807</a>  Conversation List - mentions
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- display mention item elements

### Attachments (Optional)

image

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/conversation/common/EventBadge.kt
```diff
@@ -0,0 +1,93 @@
...
+                textAlign = TextAlign.Center,
+                fontWeight = FontWeight.W700,
+                letterSpacing = 0.25.sp,
+                fontStyle = FontStyle.Normal,
+                lineHeight = 14.sp
+            ),
+            modifier = modifier
+                .wrapContentWidth()
+                .background(
+                    color = WireColor.Dark90Gray,
```
saleniuk, line 77: We shouldn't refer directly to WireColor, we should use MaterialTheme.colors.X because this way our theme works correctly and we get the proper light/dark version colors - if we want more colors than it has by default then we have to create our own custom Colors palette that extends the default one.

It might stay that way for now, I was thinking about sorting it out soon, while there aren't too many of these references yet. I wanted to know if someone has any opinion on this subject?

If we want to do it the Material way, then we should pass the "gray" color as a custom one, and assume that all shades like "gray90" are just "gray" with some alpha thus we should use `ContentAlpha`, but I'm not sure if our designs are prepared this way  so another approach is to prepare some custom colors like "badgeColor" and set proper values in our light/dark palettes.

app/src/main/kotlin/com/wire/android/ui/conversation/common/MembershipQualifier.kt
```diff
@@ -0,0 +1,38 @@
...
+            .wrapContentWidth()
+            .background(
+                color = WireColor.Alpha10LightBlue,
+                shape = RoundedCornerShape(8.dp)
+            )
+            .padding(
+                start = 4.dp,
+                top = 2.dp,
+                bottom = 2.dp,
+                end = 4.dp
```
saleniuk, line 35: You can use `horizontal` and `vertical` when both sides are the same 😄

app/src/main/kotlin/com/wire/android/ui/theme/Typography.kt
```diff
@@ -0,0 +1,81 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.sp
+
+
+val Typography.subLine1: TextStyle
+    get() = TextStyle(
+        color = WireColor.Dark70Gray,
```
saleniuk, line 13: Typography shouldn't contain colors - `typography`, `colors` and `shapes` are three separate elements of `MaterialTheme` and I think they shouldn't be mixed

Typography in my opinion should only specify the parameters of the font (weight, size,  letter spacing) and then colors should be set separately, depending on the style of the app, thanks to that approach, we can define a typography for a specific text and let our Theme decide if it should have light or dark coloured text

app/src/main/kotlin/com/wire/android/ui/conversation/common/EventBadge.kt
trOnk12: I think this could an topic to discuss indeed

app/src/main/kotlin/com/wire/android/ui/theme/Typography.kt
trOnk12: will leave it as it is for now, as we are going to refactor it soon

vitorhugods:
I think this PR is cursed. The actual diff between the branches is very different from what GitHub is showing. 

GitHub is bugged. I guess we can close this PR and do another one.

PR #317, opened 2022-01-20
feat: add login screen UI [AR-831]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-831" title="AR-831" target="_blank">SubtaskAR-831</a>  Implement the UI
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Login screen UI

### Attachments (Optional)

login

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+                val text = if(isLoading) stringResource(R.string.label_logging_in) else stringResource(R.string.label_login)
+                Text(
+                    text = text,
+                    textAlign = TextAlign.Center,
+                    style = MaterialTheme.typography.button,
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(horizontal = 32.dp),
+                )
+                androidx.compose.animation.AnimatedVisibility(
```
trOnk12, line 187: why this import here ?

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+                    modifier = Modifier.align(Alignment.CenterEnd),
+                    visible = isLoading,
+                    enter =  fadeIn() + scaleIn(),
+                    exit =  fadeOut() + scaleOut()
+                ) {
+                        CircularProgressIndicator(
+                            modifier = Modifier.size(16.dp).align(Alignment.CenterEnd),
+                            color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current),
```
trOnk12, line 195: abit weird line here not sure what is happening here, could be just me :P

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+                            strokeWidth = 2.dp)
+                }
+            }
+        }
+    }
+}
+
+private fun validInput(email: String, password: String): Boolean =
```
trOnk12, line 203: shouldn;'t it be done on the viewmodel side ?

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+                    onClick = { Toast.makeText(context, "Forgot password click 💥", Toast.LENGTH_SHORT).show() } //TODO
+                )
+        )
+    }
+}
+
+@OptIn(ExperimentalAnimationApi::class)
+@Composable
+private fun LoginButton(modifier: Modifier, email: String, password: String) {
+    var isLoading by remember { mutableStateOf(false) }
```
trOnk12, line 162: shouldn't it be done on the viewmodel side ? I think you can just create a "mock" ViewModel and a "mock" State object that would hold stuff like that and glue the buisness logic part from Kalium later on into that ViewModel, but maybe you  just have a different approach and you do not use ViewModel at the moment for creating the UI elements

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+    val interactionSource = remember { MutableInteractionSource() }
+    Column(modifier = modifier) {
+        val enabled = validInput(email, password) && !isLoading
+        Button(
+            interactionSource = interactionSource,
+            shape = RoundedCornerShape(16.dp),
+            elevation = ButtonDefaults.elevation(0.dp, 0.dp, 0.dp),
+            colors = AnimatedButtonColors(enabled = enabled),
+            onClick = { isLoading = true }, //TODO
```
MohamadJaara, line 171: loading state should be handled by the viewModel

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+                            color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current),
+                            strokeWidth = 2.dp)
+                }
+            }
+        }
+    }
+}
+
+private fun validInput(email: String, password: String): Boolean =
+    email.isNotEmpty() && password.isNotEmpty()
```
MohamadJaara, line 204: same here, it can be handeled by the viewModel

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
saleniuk: There is a bug, probably in the language, take a look here: https://stackoverflow.com/questions/67975569/why-cant-i-use-animatedvisibility-in-a-boxscope

The workaround is to use the fully qualified name 🤷

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
saleniuk: It just sets the color and alpha of the progress to be the same as the current button text color. I changed it to use `buttonColors` variable to be more clear hopefully 😉

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
saleniuk: Yes it can be in the ViewModel, current solution is temporary just to get the UI working, I assumed that this PR would only add UI, the logic and ViewModel would be added in the next

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
saleniuk: Yes it can be in the ViewModel, current solution is temporary just to get the UI working, I assumed that this PR would only add UI, the logic and ViewModel would be added in the next

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+        ),
+        modifier = modifier,
+    )
+}
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+private fun PasswordInput(modifier: Modifier, password: TextFieldValue, onPasswordChange: (TextFieldValue) -> Unit) {
+    val keyboardController = LocalSoftwareKeyboardController.current
+    var passwordVisibility by remember { mutableStateOf(false) }
```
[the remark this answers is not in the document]
saleniuk: all these three are equivalent:
```
val mutableState = remember { mutableStateOf(default) }
var value by remember { mutableStateOf(default) }
val (value, setValue) = remember { mutableStateOf(default) }
```

I think that the one I used is the clearest one but I'm open to change if someone thinks otherwise 😉

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+}
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+private fun LoginContent() {
+    val systemUiController = rememberSystemUiController()
+    val useDarkIcons = MaterialTheme.colors.isLight
+    val backgroundColor = MaterialTheme.colors.background
+
+    var email by remember { mutableStateOf(TextFieldValue("")) }
```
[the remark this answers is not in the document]
saleniuk: No problem, current solution is temporary just to get the UI working, it can be stored in the ViewModel or like right now as a state in the compose and then passed to ViewModel when user clicks "login" (I assume you prefer the first one 😛 ), either way it's something that we can decide when introducing the ViewModel for this UI 😉

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
```diff
@@ -0,0 +1,204 @@
...
+        visualTransformation = if (passwordVisibility) VisualTransformation.None else PasswordVisualTransformation(),
+        trailingIcon = {
+            val image = if (passwordVisibility) Icons.Filled.Visibility else Icons.Filled.VisibilityOff
+            IconButton(onClick = { passwordVisibility = !passwordVisibility }) {
+                Icon(imageVector = image, "", modifier = Modifier.size(20.dp))
+            }
+        },
+        modifier = modifier,
+    )
+}
```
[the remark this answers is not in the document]
saleniuk: Ok, I created `WirePasswordTextField` and moved everything that could be moved there

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
saleniuk: Yes it can be in the ViewModel, current solution is temporary just to get the UI working, I assumed that this PR would only add UI, the logic and ViewModel would be added in the next

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
saleniuk: Yes it can be in the ViewModel, current solution is temporary just to get the UI working, I assumed that this PR would only add UI, the logic and ViewModel would be added in the next

app/src/main/kotlin/com/wire/android/ui/login/Login.kt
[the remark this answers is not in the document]
trOnk12: I am just not familiar with remember function I think, but when I read I think that remember { mutableStateOf(default) } return a object that is mutable itself, and if you add var infront of it , to me it kind of makes me think you want to mutate object that is mutable in itself, I dont know maybe just me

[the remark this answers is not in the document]
trOnk12: I do not think you should declare somewhere in the code something like this :

value = someNewMutableStateOf(), that is why I think making it var instead of val misses the point ?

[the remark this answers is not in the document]
saleniuk: That's why you can make `val item = remember{...}` and you can also use `var item by remember {...}` - `by` uses getter (`return this.value`) and setter (`this.value = value`) under the hood and thanks to that you can simplify your code by just using `item = "abc"` instead of `item.value = "abc"`

version with `by` is actually used in the documentation so this is Google's suggested way 😉 
You can read more here: https://developer.android.com/jetpack/compose/state#state-in-composables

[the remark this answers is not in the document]
trOnk12: ahhhhh I see :P I did not see the by keyword, idk to me item.value makes more sense and is more readable

trOnk12 (approved):
looks good to me, good job

PR #322, opened 2022-01-24
feat: add login viewModel [AR-823]

MohamadJaara:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-823" title="AR-823" target="_blank">StoryAR-823</a>  Login by email
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

1.  add dagger/hilt for di
2. injecting coreLogic in the app
3. implementing a basic login ViewModel (error handling is not yet implemnted)

### Issues

AR-823

Needs releases with:

### Testing
N/A

----
##### References


app/src/main/kotlin/com/wire/android/feature/auth/login/LoginViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+        savedStateHandle.set(USER_IDENTIFIER_SAVED_STATE_KEY, userIdentifier.value)
+    }
+
+    private val password = mutableStateOf(String.EMPTY)
+    fun onPasswordChange(newText: String) {
+        password.value = newText
+    }
+
+    private val _loginResultLiveData = MutableLiveData<AuthenticationResult>()
+    val loginResultLiveData: LiveData<AuthenticationResult> = _loginResultLiveData
```
saleniuk, line 36: correct me if I'm wrong, but I think the code style guideline describes the order to have val first, followed by methods 😄

app/src/main/kotlin/com/wire/android/feature/auth/login/LoginUseCase.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.wire.kalium.logic.CoreLogic
+import javax.inject.Inject
+import com.wire.kalium.logic.feature.auth.AuthenticationResult
+import dagger.hilt.android.scopes.ViewModelScoped
+
+@ViewModelScoped
+class LoginUseCase @Inject constructor(private val coreLogic: CoreLogic) {
+    suspend operator fun invoke(userIdentifier: String, password: String): AuthenticationResult =
+        coreLogic.authenticationScope {
+            login(userIdentifier, password, true)
```
saleniuk, line 12: In this case we will have usecases in both android reloaded and kalium - in the documentation on the graph it looks like they are only supposed to be in kalium - are we changing this concept?

Would it be worse to inject and use `corelogic` directly in `LoginViewModel`?

MohamadJaara: the issue with using `CoreLogic` is `ViewModel` will become harder to understand since almost all of the view models will look like this:
```kotlin
@HiltViewModel
class MyViewModel @Inject constructor(
    private val coreLogic: CoreLogic
) : ViewModel() {
```
instead of a more descriptive:
```kotlin
@HiltViewModel
class MyViewModel @Inject constructor(
    private val getMessagesUseCase: GetMessagesUseCase,
    private val updateMessageUsecase: UpdateMessageUsecase
) : ViewModel() {
```
and it makes it a bit harder to understand what a viewModel dose from the first look

vitorhugods: Can't we inject Kalium's usecases instead of having this tiny eggshell around it?

MohamadJaara: @vitorhugods @saleniuk i did some changes to the hilt `Module`
i was able to inject the `Logic` use cases 

```kotlin
@Module
@InstallIn(SingletonComponent::class)
class CoreLogicModule {

    @Singleton
    @Provides
    fun coreLogicProvider(@ApplicationContext context: Context): CoreLogic {
        ...
        return CoreLogic(.....)
    }

    @Singleton
    @Provides
    fun loginUserCaseProvider(coreLogic: CoreLogic): LoginUseCase = coreLogic.getAuthenticationScope().loginUsingEmail
}

```

MohamadJaara: this approach will not work with other use cases since `getSessionScope(session: AuthSession)` needs session.

vitorhugods: Yeah.

The ones that require a session are troublesome. And can be split based on their nature I guess. Things like `GetSelfUserProfileUseCase` or similar are only used for the `current` session, _i.e._ the session that is currently selected and that the UI is displaying. So we could have something like:
```kotlin
@Factory
@Provides
fun getSelfUserProfileUseCase(coreLogic: CoreLogic, sessionManager: SessionManager): GetSelfUserProfileUseCase 
    = coreLogic.getSessionScope(sessionManager.currentSession).user.selfProfileData
```

Things like `SendMessageUseCase` are a bit harder. Because technically we could send messages when the user replies in a system notification of a new message from another session, different from the one being displayed in the UI.

Well, I don't know. Should we support this initially? Maybe it's something we need to discuss with the designers.
In any case, if we don't support initially and use only the `currentSession` for everything, it will suck big time if we need to refactor later.

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    @Singleton
+    @Provides
+    fun coreLogicProvider(@ApplicationContext context: Context): CoreLogic {
+        val proteusPath = context.getDir("proteus", Context.MODE_PRIVATE).path
+        val deviceLabel = DeviceLabel.label
+        return CoreLogic(applicationContext = context, rootProteusDirectoryPath = proteusPath, clientLabel = deviceLabel)
+    }
+
+    @Singleton
```
vitorhugods, line 27: There's no need to have it as a `@Singleton`, right? 
Maybe for this UseCase the `@Singleton` doesn't hurt. But for others, it may.

Especially if we end up introducing something like I mentioned in another comment, to get the current/active session in order to create the UseCase. If it is Singleton, when we change sessions, it will keep the old UseCase, of the past active session.

In any case, I guess UseCases can come out resource hungry in the end, as they consume Repositories that will be merging databases, networking, caches, etc.
So if we can free these resources when they are no longer being used, it would be for the best.
Also, better to have all UseCases behaving the same in terms of lifecycle, than have some Singletons, and some not.

So, my recommendation is to change to `@Factory`. Not for now, not specifically for this UseCase, but to set an example for the future.

build.gradle.kts
```diff
@@ -3,6 +3,10 @@ buildscript {
         google()
         mavenCentral()
     }
+    dependencies {
+        classpath(Libraries.Hilt.gradlePlugin)
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.10")
```
vitorhugods, line 8: I don't think this line is necessary, as it's already being added in `buildSrc/build.gradle.kts`

We can upgrade the version tho. From `1.5.31`, to `1.6.10`.

Perhaps moving `Hilt.gradlePlugin` there too.

vitorhugods (changes requested):

app/src/main/kotlin/com/wire/android/feature/auth/login/LoginUseCase.kt
saleniuk: In such case, the message that is shown as a notification should contain id of the user this message is sent to, so maybe we could have access to other sessions, like `sessionManager.getSessionForUserId(userId)`?

vitorhugods: @saleniuk exactly. But how to tie that with injection I have no idea yet :P

saleniuk: `SessionManager` can have two methods and be available from `CoreLogic` or even add these two methods directly to the `CoreLogic`
```
    fun getCurrentAuthSession(): AuthSession?
    fun getAuthSessionForUserId(userId: String): AuthSession?
```
but I don't know what you think about sharing `AuthSession` outside kalium, maybe `coreLogic.getSessionScope` should take `userId` as a parameter and by default it should be `currentAuthSession.userId`

vitorhugods: We are adding persistence to Kalium, and I guess we can manage the storage of multiple-sessions there, yes. 

But still, the app at some point will have to choose which session is doing the operation.
Some (probably most) will be for the current session. For that reason injecting the `UseCase` works perfectly by defining a `UseCase` as `CoreLogic` + `currentSession`.
But for some... It just can't be done and must be resolved depending on which session is receiving a call, or replying to a message through notifications, etc.

So, maybe we can start by adding `@ViewModelScoped` to all things that rely on `currentUser`. We scope down this definition to only be used with ViewModels. This means that the view must be visible, and we're dealing with the current user. Kinda makes sense.

When we need to handle system notifications or some kind of event that starts in the background, we can figure this out later. Worst case scenario we use `CoreLogic` directly in these elements.

saleniuk: If I understand it correctly, now we have to make `coreLogic. getSessionScope` without parameter, let it return the `SessionScope` for the `currentUser` and keep the `currentUser` only inside the kalium?

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -0,0 +1,32 @@
...
+import javax.inject.Singleton
+
+
+@Module
+@InstallIn(ViewModelComponent::class)
+class CoreLogicModule {
+
+    @ViewModelScoped
+    @Provides
+    fun coreLogicProvider(@ApplicationContext context: Context): CoreLogic {
```
vitorhugods, line 23: I'd say CoreLogic can - and maybe should - be a singleton. Because inside of it, we might want to have singletons in its injection mecanism.

vitorhugods (changes requested):

vitorhugods (approved):

saleniuk (approved):

PR #332, opened 2022-01-26
chore: change app navigation to more flexible

borichellow:
# What's new in this PR?

- Added Hilt
- Refactored Navigation. 
- Extracted all the navigation staff into a separate `navigation` package.
- Added `NavigationManager` singleton, all the navigations should be done by using it (it can be injected into any ViewModel by Hilt)
- Extracted `WireDrawer` and `WireTopBar` into separate packages and added ViewModel to each of them.

### Issues

Prev. navigation was not so obvious and flexible and didn't fit our needs. 

### Attachments (Optional)

----
##### References
1. [article that that was used as an inspiration for the NavigationManager pattern]( https://joebirch.co/android/modular-navigation-with-jetpack-compose/)


app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -1,25 +1,115 @@
...
+                        drawerState = drawerState,
+                        drawerContent = drawerContent
+                    ) {
+                        Scaffold(
+                            topBar = topBar,
+                        ) {
+                            NavigationGraph(navController = navController)
+                        }
+                    }
+                }
```
saleniuk, line 78: I'm afraid that with this approach, it's not possible to show other screens that aren't part of the "main" navigation tree like WelcomeScreen which doesn't show TopBar at all and LoginScreen which has a different TopBar than the MainScreen. 
I thought that every "fullscreen" screen should have its own TopBar since they are completely different and when from MainScreen we open a particular ConversationScreen it should replace the whole screen instead of just the content below the TopBar 🤔 
With this approach that I'm thinking about, we don't need to have single global navigation that handles all, we can have multiple nested ones, like: global app navigation can handle switching "fullscreen" screens (Splash, Login, Welcome, Main, Conversation, Call, Profile, Settings) and the MainScreen has its own inner navigation that handles only the content that should be shown inside the MainScreen - different items of the Drawer (Conversations, Vault, Archive), and then also ConversationsScreen can have another inner navigation that handles its tabs (All, Calls, Mentions). What do you think?

app/src/main/kotlin/com/wire/android/navigation/NavigationManager.kt
```diff
@@ -1,23 +1,25 @@
...
-
-    suspend fun drawerState(isOpened: Boolean) {
-        drawerState.emit(isOpened)
-    }
-
 }
+
+data class NavigationCommand(
+    val destination: String,
+    val removeBackStack: Boolean = false
```
saleniuk, line 23: Maybe we should rename it as I wasn't sure what it was for at first, for Activities such flag is called "NO_HISTORY", but maybe `skipBackStack` would be better? 😃

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
```diff
@@ -0,0 +1,91 @@
...
+) {
+    val scope = rememberCoroutineScope()
+    val titleText = stringResource(id = title ?: R.string.conversations_screen_title)
+
+    Column(
+        Modifier
+            .fillMaxWidth()
+            .background(MaterialTheme.colorScheme.background)
+    ) {
+        Row(
```
saleniuk, line 51: So this is now fully custom? Why don't use a ready-made `CenterAlignedTopAppBar`?

saleniuk (commented):
Two small comments, other than that 👏

app/src/main/kotlin/com/wire/android/navigation/NavigationManager.kt
borichellow: good point

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
borichellow: problem that there should be Search field at the bottom. So i tried to have `Column(CenterAlignedTopAppBar, SearchField)` but it looks a bit different from the designs. Basically because `CenterAlignedTopAppBar` has its own height and bottom padding, so the distance between top part of the TopBar and Search field is bigger than should.

gongracr: I can't build the app on my side to check how it would look like. Could you maybe attach a screenshot of it?

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -0,0 +1,66 @@
...
+    val arguments: List<NamedNavArgument> = emptyList(),
+    open val content: @Composable (NavBackStackEntry) -> Unit
+    //TODO add animations here
+) {
+
+//    object Splash  //TODO
+//    object Login  //TODO
+
+    object Home : NavigationItem(
+        route = "home/{$HOME_START_TAB_INDEX_ARGUMENT}",
```
gongracr, line 27: Can we extract all these routes to vals?

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -1,25 +1,74 @@
...
         }
     }
+
+    @Composable
+    private fun setUpNavigation(
+        navController: NavHostController,
+        scope: CoroutineScope
+    ) {
+        // with the static key here we're sure that this effect wouldn't be canceled or restarted
+        LaunchedEffect("key") {
```
gongracr, line 57: What's this effect?

borichellow: Basically, to run suspended functions from Composable we need to use `LaunchedEffect` you can read more [here](https://developer.android.com/jetpack/compose/side-effects)

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.navigation.NavBackStackEntry
+import androidx.navigation.NavType
+import androidx.navigation.navArgument
+import com.wire.android.ui.home.HomeScreen
+import com.wire.android.ui.settings.SettingsScreen
+import com.wire.android.ui.support.SupportScreen
+import com.wire.android.ui.home.userprofile.UserProfileScreen
+
+@ExperimentalMaterial3Api
+open class NavigationItem(
```
gongracr, line 16: Shouldn't this be a sealed class?

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -0,0 +1,66 @@
...
+//    object Login  //TODO
+
+    object Home : NavigationItem(
+        route = "home/{$HOME_START_TAB_ARGUMENT}",
+        content = { HomeScreen(it.arguments?.getString(HOME_START_TAB_ARGUMENT), hiltViewModel()) },
+        arguments = listOf(
+            navArgument(HOME_START_TAB_ARGUMENT) { type = NavType.StringType }
+        )
+    ) {
+        fun navigationRoute(startTabRoute: String): String = "home/$startTabRoute"
```
gongracr, line 33: Unused function. What do we need it for ?

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -0,0 +1,66 @@
...
+    val arguments: List<NamedNavArgument> = emptyList(),
+    open val content: @Composable (NavBackStackEntry) -> Unit
+    //TODO add animations here
+) {
+
+//    object Splash  //TODO
+//    object Login  //TODO
+
+    object Home : NavigationItem(
+        route = "home/{$HOME_START_TAB_ARGUMENT}",
```
gongracr, line 27: This is referencing the constant value "start_tab_index". Is this intended or shouldn't we change it dynamically with the route component of the home screen we are at?

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
borichellow: this supposes to be used to get the navigation route with the params. So I kept it here as an example

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
borichellow: this `route` is used for creating the NavGraph, there we should pass the param name that we expect. 
And the actual param is passed while navigating (`navController.navigate()`), that's why I added `fun navigationRoute()`, it should be used for `navController.navigate()`

gongracr (approved):

PR #341, opened 2022-01-31
chore: add Authentication flow

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Attach the Authentication flow to the current navigation

### Solutions

create `AuthScreen` with it's own navGraph

----

----
##### References


app/src/main/kotlin/com/wire/android/feature/auth/login/LoginViewModel.kt
```diff
@@ -39,14 +48,17 @@ class LoginViewModel @Inject constructor(
...
-        password.value = newText
+        password = newText
     }
 
+    suspend fun navigateToConvScreen() =
+        navigationManager.navigate(NavigationCommand(NavigationItem.Home.navigationRoute("home_conversations"), true))
+
     private companion object {
         const val USER_IDENTIFIER_SAVED_STATE_KEY = "user_identifier"
     }
```
trOnk12, line 64: it looks like this ViewModel uses a MVVM approach, it is recommended by Google to use a MVI approach so I would suggest to refactor it, maybe it is something we need to discuss internally first ?

app/src/main/kotlin/com/wire/android/ui/authentication/login/Login.kt
```diff
@@ -67,7 +75,11 @@ private fun LoginContent() {
                 Spacer(modifier = Modifier.height(16.dp))
                 ForgotPasswordLabel(modifier = Modifier.fillMaxWidth())
             }
-            LoginButton(modifier = Modifier.fillMaxWidth(), email = email.text, password = password.text)
+            LoginButton(modifier = Modifier.fillMaxWidth(), email = email.text, password = password.text) {
+                scope.launch {
+                    loginViewModel.navigateToConvScreen()
+                }
```
trOnk12, line 81: I wonder if action like this should be scoped to a CoroutineScope of the Composable, I think making navigateToConvScreen not a suspend function and using ViewModel CoroutineScope would be a better approach

app/src/main/kotlin/com/wire/android/ui/authentication/login/Login.kt
```diff
@@ -32,25 +33,32 @@ import androidx.compose.ui.text.input.KeyboardType
...
 @OptIn(ExperimentalComposeUiApi::class, ExperimentalMaterial3Api::class)
 @Composable
-private fun LoginContent() {
+private fun LoginContent(
+    loginViewModel: LoginViewModel = hiltViewModel(),
+    scope: CoroutineScope
+) {
 
     var email by remember { mutableStateOf(TextFieldValue("")) }
     var password by remember { mutableStateOf(TextFieldValue("")) }
```
trOnk12, line 64: I am not an expert but in MVI it should be working like this : 

UI > Send Event to ViewModel > ViewModel process the event > ViewModel emits new state of the State object 

so I would move this "email" and "password" fields into the ViewModel and react to it, do some validation etc.

app/src/main/kotlin/com/wire/android/ui/authentication/welcome/Welcome.kt
```diff
@@ -142,19 +153,19 @@ private fun WelcomeCarouselItem(pageIconResId: Int, pageText: String) {
 }
 
 @Composable
-private fun WelcomeButtons(modifier: Modifier) {
+private fun WelcomeButtons(modifier: Modifier, onLoginClick: () -> Unit, onEnterpriseAccountClick: () -> Unit) {
     Column(modifier = modifier) {
-        val context = LocalContext.current
 
+        // login button
```
trOnk12, line 159: instead of comments use a private method with naming for exmaple

app/src/main/kotlin/com/wire/android/ui/authentication/login/Login.kt
```diff
@@ -67,7 +75,11 @@ private fun LoginContent() {
                 Spacer(modifier = Modifier.height(16.dp))
                 ForgotPasswordLabel(modifier = Modifier.fillMaxWidth())
             }
-            LoginButton(modifier = Modifier.fillMaxWidth(), email = email.text, password = password.text)
+            LoginButton(modifier = Modifier.fillMaxWidth(), email = email.text, password = password.text) {
```
trOnk12, line 78: I think those field should rather consume a String field coming from the ViewModel State object

app/src/main/kotlin/com/wire/android/feature/auth/login/LoginViewModel.kt
```diff
@@ -16,20 +21,24 @@ import com.wire.kalium.logic.feature.auth.LoginUseCase
...
 
-    private val userIdentifier = mutableStateOf(
+    var userIdentifier by mutableStateOf(
         savedStateHandle.get(USER_IDENTIFIER_SAVED_STATE_KEY) ?: String.EMPTY
     )
-    private val password = mutableStateOf(String.EMPTY)
+        private set
+
+    var password by mutableStateOf(String.EMPTY)
+        private set
```
trOnk12, line 34: you could create a State object as a plain Kotlin Class and wrap it into MutableStateFlow()

data LoginState( val password : String, val userIdentifier : String) and consume it inside the LoginContent for example with collectAsState method()

saleniuk: Do we have to mix coroutine `Flow` and compose `State` just to keep the values in the ViewModel?

app/src/main/kotlin/com/wire/android/feature/auth/login/LoginViewModel.kt
MohamadJaara: this need more discussien with the team, for new i will leave the viewModel as it is

app/src/main/kotlin/com/wire/android/ui/authentication/login/Login.kt
MohamadJaara: I took `Home Drawer.kt` as inspiration, to me the navigation to a different screen is a UI task so scoping this to the `ViewModel` (which lives longer than the UI) makes no sense

app/src/main/kotlin/com/wire/android/feature/auth/login/LoginViewModel.kt
trOnk12: I guess depends on a use case, but now you say it is not really required here I tihnk but MutableStateFlow gives you more option to use operators as combine and stuff.

MohamadJaara: i don't see it as necessary since `user Identifier` and `password` are storing the state of a `TextField`, which start acting weird (the text was duplicated after each input) when assigning a string from viewModel as value e.g. 
```kotlin
@Composable
private fun UserIdentifierInput(modifier: Modifier, userIdentifier: String, onValueChange: (TextFieldValue) -> Unit) {
    WireTextField(
        value = TextFieldValue(userIdentifier), 
        ...
    )
}

use:
UserIdentifierInput(
     modifier = Modifier.fillMaxWidth(),
      userIdentifier = loginViewModel.userIdentifier, // loginViewModel.userIdentifier in this case is a string  mutableState 

```

trOnk12:
btw. I just see that you use LoginViewModel into com.wire.android.feature.auth.login is there any reason why you do not move it to ui.login ?

MohamadJaara:
@trOnk12 `LoginViewModel.it` was not created with UI (the UI was not yet implemented) that is why it's located in a diffrent package, i can move it np

trOnk12 (approved):

saleniuk (approved):

PR #349, opened 2022-02-01
feat: display group items of conversation list

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- display the group items of the conversation list

_Briefly describe the issue you have solved or implemented with this pull request. If the PR contains multiple issues, use a bullet list._

Needs releases with:

### Attachments (Optional)
image

image

image

##### References


app/src/main/kotlin/com/wire/android/ui/home/conversations/common/ConversationItemTemplate.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.wire.android.ui.home.conversationslist.common.RowItem
+import com.wire.android.ui.home.conversationslist.model.EventType
+
+@Composable
+fun ConversationItemTemplate(
+    leadingIcon: @Composable () -> Unit,
+    title: @Composable () -> Unit,
+    subTitle: @Composable () -> Unit = {},
+    eventType: EventType? = null,
+    onConversationItemClick: () -> Unit = { }
```
saleniuk, line 21: Should `RowItem` be always clickable? If so then it shouldn't have a default empty value, because it will make a ripple effect even if there is no real action after that

app/src/main/kotlin/com/wire/android/ui/home/conversations/common/GroupConversation.kt
```diff
@@ -0,0 +1,40 @@
...
+import androidx.compose.ui.unit.dp
+import com.wire.android.ui.theme.wireTypography
+
+@Composable
+fun GroupConversationAvatar(colorValue: Long) {
+    Box(
+        modifier = Modifier
+            .padding(8.dp)
+            .size(width = 32.dp, height = 32.dp)
+            .background(color = Color(colorValue), shape = RoundedCornerShape(10.dp))
```
saleniuk, line 23: We should move such values (size and corner size) to Dimensions to be able to set bigger avatar for larger screens for instance

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/Call.kt
```diff
@@ -3,16 +3,29 @@ package com.wire.android.ui.home.conversationslist.model
...
-data class Call(val callInfo: CallInfo, val conversation: Conversation)
+data class Call(
+    val callInfo: CallInfo,
+    override val conversationType: ConversationType
+) : Conversation()
 
-data class CallTime(val date: String, val time: String) {
+data class CallTime(
+    val date: String,
+    val time: String
```
saleniuk, line 13: Wouldn't be better to have `time` as a `Long` timestamp? 🤔

trOnk12: Good suggestion but I'll leave as it is and add a comment instead of changing itand wait for the Kalium so that I can format it to correct String by then

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/Conversation.kt
```diff
@@ -5,21 +5,21 @@ import com.wire.android.R
...
-    UserInfoLabel(
-        labelName = conversationInfo.name,
-        isLegalHold = conversationInfo.isLegalHold,
-        membership = conversationInfo.membership,
-    )
+    data class PrivateConversation(
+        val userInfo: UserInfo,
+        val conversationInfo: ConversationInfo
+    ) : ConversationType()
+}
```
vitorhugods, line 22: What about just having a sealed class without open/abstract `Conversation`?

```kotlin
sealed class Conversation {
    data class GroupConversation(
        val groupColorValue: Long,
        val groupName: String
    ) : Conversation()

    data class PrivateConversation(
        val userInfo: UserInfo,
        val conversationInfo: ConversationInfo
    ) : Conversation()
}
```

This way we can get exhaustive `when` branches:
```kotlin
when(conversation){
    is GroupConversation -> {}
    is PrivateConversation -> {}
} //No need for an else branch
```

trOnk12: that is actually what I did with :
```

sealed class ConversationType {
    data class GroupConversation(
        val groupColorValue: Long,
        val groupName: String
    ) : ConversationType()

fun Conversation.toUserInfoLabel() =
    UserInfoLabel(
        labelName = conversationInfo.name,
        isLegalHold = conversationInfo.isLegalHold,
        membership = conversationInfo.membership,
    )
    data class PrivateConversation(
        val userInfo: UserInfo,
        val conversationInfo: ConversationInfo
    ) : ConversationType()
}
```


but the problem is that I wanted to create a generic function that just "emits" the compose depending on what Row Item it is, so that we can just call one function, it could be either a "Call", "Mention" or "General" Row item 

you can see it here: 

```

@Composable
inline fun <reified Conversation> ConversationItemFactory(
    item: Conversation,
    eventType: EventType? = null,
    noinline onConversationItemClick: () -> Unit
) {
    when (item) {
        is GeneralConversation -> AllConversationItem(
            item,
            eventType,
            onConversationItemClick
        )
        is Call -> CallConversationItem(
            item,
            eventType,
            onConversationItemClick
        )
        is Mention -> MentionConversationItem(
            item,
            eventType,
            onConversationItemClick
        )
        else -> throw IllegalStateException("Illegal item type for ConversationItemFactory")
    }
}
```

vitorhugods: Kinda feels like the model is upside down.

Because `Mention` isn't a `Conversation`. `Call` isn't a `Conversation` either.

They can have a reference to a `Conversation` maybe?

For example:
```kotlin
sealed class Conversation {
    data class GroupConversation(
        val groupColorValue: Long,
        val groupName: String
    ) : Conversation()

    data class PrivateConversation(
        val userInfo: UserInfo,
        val conversationInfo: ConversationInfo
    ) : Conversation()
}

data class Call(val callInfo: CallInfo, val conversation: Conversation)
data class Mention(val mentionInfo: MentionInfo, val conversation: Conversation)
```

There's no need for `Mention` and `Call` to extend `Conversation`.

The way it currently is, it allows us to do weird stuff:
```kotlin
val groupConversation = GroupConversation(color, name)
val call = Call(callInfo, groupConversation)
//So far, so good.

val callConversation: Conversation = Mention(mentionInfo, call) 
// This doesn't make sense.

println(callConversation is GroupConversation) // returns false. Why? I passed a GroupConversation, I would expect this to be true.
```

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/AllConversationScreen.kt
```diff
@@ -0,0 +1,60 @@
...
+    LazyColumn {
+        folderWithElements(
+            header = { stringResource(id = R.string.conversation_label_new_activity) },
+            items = newActivities
+        ) { newActivity ->
+            ConversationItemFactory(
+                item = newActivity.conversation,
+                eventType = newActivity.eventType,
+                onConversationItemClick = { onConversationItemClick("someId") }
+            )
```
vitorhugods, line 45: Isn't the list of `conversations` always a `GeneralConversation`?

We could just do:
```suggestion
           AllConversationItem(
               item,
               eventType,
               onConversationItemClick
           )
```

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/Conversation.kt
trOnk12: hmmm you got a point here I will give a tough abnout it tommorow, if you have an idea, please comment it here

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/CallScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+    LazyColumn {
+        folderWithElements(
+            header = { stringResource(id = R.string.calls_label_missed_calls) },
+            items = missedCalls
+        ) { missedCall ->
+            ConversationItemFactory(
+                item = missedCall,
+                eventType = EventType.MissedCall,
+                onConversationItemClick = { onCallItemClick("someId") }
+            )
```
vitorhugods, line 42: Here, for example, we know it's a `Call`, and that this screen will always show `Call` items, not a `Mention` for example. 

So, we could just:

```suggestion
            CallConversationItem(
                missedCall,
                EventType.MissedCall,
                { onCallItemClick("someId") }
            )
```

vitorhugods: And it goes the same for `MentionScreen`.

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/Conversation.kt
vitorhugods: Even though the UI looks similar, the models aren't. There's no need to make them all extend the same thing. 

They are displayed in different lists, and represent different events / data structures. The only common thing they have is that they hold a reference to a Conversation.

But we don't have a need to display them all in the same list, and if we did, it doesn't mean that the model itself should change, but rather the view state - the model representing the view. 

Doesn't seem nice to mix core model entities so much - so early - in the project.

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/AllConversationScreen.kt
trOnk12: I mean we could but we have a factory for that so it takes a decision for us I think

vitorhugods (dismissed):

PR #354, opened 2022-02-03
feat: User profile screen

borichellow:
[Task-820](https://wearezeta.atlassian.net/browse/AR-820)
----
# What's new in this PR?

Added SelfProfile Screen

Screenshot_1643878991

Screenshot_1643879084


app/src/main/kotlin/com/wire/android/ui/common/SearchBar.kt
```diff
@@ -34,7 +34,7 @@ fun SearchBarUI(placeholderText: String, modifier: Modifier = Modifier, onTextTy
     var text by remember { mutableStateOf(TextFieldValue()) }
 
     WireTextField(
-        modifier = modifier.padding(bottom = 16.dp, end = 16.dp, start = 16.dp),
+        modifier = modifier.padding(bottom = 16.dp, end = 8.dp, start = 8.dp),
```
gongracr, line 37: Why this change if UserProfileScreen doesn't implement any search?

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
```diff
@@ -44,26 +45,14 @@ fun UserProfileAvatar(
             contentDescription = stringResource(R.string.content_description_user_avatar),
             contentScale = ContentScale.Crop,
             modifier = Modifier
+                .padding(dimensions().userAvatarStatusBorderSize)
                 .background(Color.Black, CircleShape)
                 .size(size)
         )
-        Box(
-            contentAlignment = Alignment.Center,
+        UserStatusDot(
```
gongracr, line 52: Dot? Can we try to rename to something that is not design specific so that we avoid possible refactorings if designs change in the future? 
Maybe `UserStatusIndicator`, `UserStatusBadge` or `UserStatusMarker` ?

app/src/main/kotlin/com/wire/android/ui/common/UserStatusDot.kt
```diff
@@ -0,0 +1,108 @@
...
+}
+
+@Composable
+private fun NoneDot() {
+
+}
+
+enum class UserStatus {
+    AVAILABLE, BUSY, AWAY, NONE
+}
```
gongracr, line 84: Let's extract it to different file so that we separate logic and models from UI implementation

app/src/main/kotlin/com/wire/android/ui/common/UserStatusDot.kt
```diff
@@ -0,0 +1,108 @@
...
+@Composable
+fun AwayPreview() {
+    UserStatusDot(UserStatus.AWAY)
+}
+
+@Preview(name = "NonePreview")
+@Composable
+fun NonePreview() {
+    UserStatusDot(UserStatus.NONE)
+}
```
gongracr, line 108: Can we make the previews functions `private` so that they are not exposed to the rest of the code?

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+            titleContentColor = MaterialTheme.colorScheme.onBackground,
+            actionIconContentColor = MaterialTheme.colorScheme.onBackground,
+            navigationIconContentColor = MaterialTheme.colorScheme.onBackground
+        ),
+        navigationIcon = {
+            IconButton(onClick = { viewModel.close() }) {
+                Icon(
+                    imageVector = Icons.Filled.Close,
+                    contentDescription = stringResource(R.string.user_profile_close_description),
+                    tint = MaterialTheme.colorScheme.onBackground
```
saleniuk, line 91: Tint isn't necessary because `CenterAlignedTopAppBar` has `colors` parameter and handles content colors itself

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+                )
+            }
+        },
+        title = {
+            Text(
+                text = stringResource(id = R.string.user_profile_tile),
+                style = MaterialTheme.wireTypography.title01,
+                color = MaterialTheme.colorScheme.onBackground,
```
saleniuk, line 99: Same with text colors

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+    val uiState by viewModel.state.collectAsState()
+
+    with(uiState) {
+        renderUserProfileScreen(state = this, viewModel = viewModel)
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun renderUserProfileScreen(state: UserProfileState, viewModel: UserProfileViewModel) {
```
saleniuk, line 61: I'm not sure about that, because here the state is extracted from the viewmodel but still the viewmodel is also passed - what I prefer is to separate it and don't pass the viewmodel to the inner composables, instead pass the state and all "onclick" functions as parameters - this way we don't need a viewmodel to create a preview and test it in the future, but it's my personal opinion

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/ConversationItem.kt
```diff
@@ -28,7 +29,7 @@ data class ConversationInfo(
 
 data class UserInfo(
     val avatarUrl: String = "",
-    val availabilityStatus: AvailabilityStatus = AvailabilityStatus.None
+    val availabilityStatus: UserStatus = UserStatus.NONE
```
gongracr, line 32: This `UserInfo` class seems like a copy of the `User` class. Any chance we can extract and unify both to a common one?

app/src/main/kotlin/com/wire/android/ui/theme/WireDimensions.kt
```diff
@@ -48,7 +77,27 @@ private val DefaultPhonePortraitWireDimensions: WireDimensions = WireDimensions(
...
+    spacing56 = 56.dp,
+    spacing64 = 64.dp,
+    corner2 = 2.dp,
+    corner4 = 4.dp,
+    corner6 = 6.dp,
+    corner8 = 8.dp,
+    corner10 = 10.dp,
+    corner12 = 12.dp,
+    corner14 = 14.dp,
+    corner16 = 16.dp
```
saleniuk, line 100: I'm still not sure which way should we use, both have its pros and cons - if we create separate dimension for every ui item then we can modify them independently but we will end up with having a lot of dependencies, with this approach we have less dimensions but for instance if we would want to support tablet screen sizes, and we use this dimension everywhere in the app then it will change also in places that we would not want it to be changed
Maybe for now we can use it the way you implemented it but I would like to know the opinion of the rest of the team 😄

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+        modifier = Modifier
+            .fillMaxWidth()
+            .fillMaxHeight()
+            .background(MaterialTheme.colorScheme.background)
+    ) {
+
+        renderTopBar(viewModel)
+        renderHeader(state, viewModel)
+        renderStatusesRow(state)
+        renderOtherAccountsList(state, viewModel)
```
gongracr, line 72: I think we should avoid this `renderXXX` naming and just create smaller composable items if it makes sense to separate them for reusage. Just to be consistent with the other screens and follow in general Compose extended practices.

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+        }
+    }
 }
 
+@OptIn(ExperimentalMaterial3Api::class)
 @Preview(showBackground = false)
 @Composable
 fun UserProfileScreenPreview() {
-    UserProfileScreen()
+    renderUserProfileScreen(
```
gongracr, line 324: Make this preview method private pls.

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+
+@Composable
+private fun ColumnScope.renderStatusesRow(state: UserProfileState) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(dimensions().spacing16),
+    ) {
+        WireSecondaryButton(
+            onClick = {},
```
gongracr, line 195: Should we at least link this click event to a common method on the VM where we will define the logic later?

gongracr (changes requested):
Small UI fixes but great job in general! I also assume that the UI logic for changing the User Status and adding a new account will be implemented in subsequent PRs?

app/src/main/kotlin/com/wire/android/ui/theme/WireDimensions.kt
gongracr: I think we should extract hardcoded dimensions from elements that are either reusable by other composable objects, or elements whose size is meant to be changed when implementing for other devices, sizes (internal paddings, width and height. 
Therefore I'm not really sure if having a semantic approach (corner12 -> 12 DP's) when extracting these dimensions makes real sense as it will not be very logic to assign `corner12 = 24.dp` on a tablet. If so, I would maybe go for a `cornerRadius1x`, `cornerRadius2x`... approach.

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
gongracr: Or even keep all the elements that depend on the VM within the same Composable function scope

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+                Icon(
+                    imageVector = Icons.Filled.Close,
+                    contentDescription = stringResource(R.string.user_profile_close_description),
+                    tint = MaterialTheme.colorScheme.onBackground
+                )
+            }
+        },
+        title = {
+            Text(
+                text = stringResource(id = R.string.user_profile_tile),
```
gongracr, line 97:
```suggestion
                text = stringResource(id = R.string.user_profile_title),
```

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
```diff
@@ -1,16 +1,343 @@
...
+                style = MaterialTheme.wireTypography.title01,
+                color = MaterialTheme.colorScheme.onBackground,
+            )
+        },
+        actions = {
+            //TODO make it red, when such WireBtn ready
```
gongracr, line 103: Can't we already specify it via the `colors` parameter?

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
borichellow: oops :)

app/src/main/kotlin/com/wire/android/ui/common/Extentions.kt
```diff
@@ -12,6 +12,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.semantics.Role
+import com.wire.android.ui.theme.wireDimensions
```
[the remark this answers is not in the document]
borichellow: I'm not sure about it, as `fun folderWithElements()` from `conversationslist.common.Extention.kt` is something specific for the `ConversationsList` only. And `ui.common.Extention.kt` suppose to contain only funs that could be useful anywhere is the project :)

app/src/main/kotlin/com/wire/android/ui/common/SearchBar.kt
borichellow: honestly: i don't know why it's here, it should be in `develop` long time ago. Maybe some merging issue, but this is how it actually should be

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
borichellow: `Indicator` sounds good to me

app/src/main/kotlin/com/wire/android/ui/common/UserStatusDot.kt
borichellow: moved it into `models. UserStatus.kt`

app/src/main/kotlin/com/wire/android/ui/common/Extentions.kt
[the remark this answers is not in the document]
gongracr: If that is so, then maybe we should move it to `package com.wire.android.ui.home.conversationslist` and mark this `fun folderWithElements()` method as `@PackagePrivate` to avoid confusion and indicate it can only used at that level.
Regarding this file, I would still fix the naming to `Extensions` 👓

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/ConversationItem.kt
borichellow: I wasn't adding any of it, so not sure why is there separate 2 data classes.
My guess: both of them are just "holders" for the data needed to display on some screen (ConversationsList and Conversation). And in the future, when we'll have some `UserDataModel` that is fetched from the API, it will be mapped into these data classes in ViewModels. 
For now they are the same, but later it could change.

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
borichellow: I could, but `colors` is `interface WireButtonColors` so i'd need to provide all the colors, but just "textColor" for example. Also Michael mentioned he is going to add an additional btn state "Error", which will be red and looks just like I need here :)

app/src/main/kotlin/com/wire/android/ui/theme/WireDimensions.kt
borichellow: @gongracr so, for example, we went to `cornerRadius1x` solution...
So it's gonna be like:            
```
for phone                     and for Tablets:
cornerRadius1x = 2.dp                cornerRadius1x = 4.dp
cornerRadius2x = 4.dp                cornerRadius2x = 8.dp
cornerRadius3x = 6.dp                cornerRadius3x = 12.dp
...
```

So what is the big difference from this 
```
corner2 = 2.dp               corner2 = 4.dp
corner4 = 4.dp               corner4 = 8.dp
corner6 = 6.dp               corner6 = 12.dp
...
```

gongracr: That for example the corner6 might give the false assumption that it is always 6.dp. Again this responds to more a personal preference than to a well valid argument. I will not oppose if you prefer leaving it like that.

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
saleniuk: It's already in the `develop` - you just use `WireButtonState.Error` 😄

saleniuk (commented):
I have no more comments, just try to make Gonzo happy and resolve his comments 😄

app/src/main/kotlin/com/wire/android/ui/home/userprofile/UserProfileScreen.kt
gongracr: More than reusing an error state button, I'd say it's better to add a `textColor` parameter in the constructor, as the error button might end up having some slightly different behavior in the future.

saleniuk: I was thinking about that, but the name "error" is because of the semantic colors approach. We have a limiter number of button types and states based on the designs so we should use these predefined ones and don't mix or modify them freely.

gongracr (approved):
Looking great now. Awesome job! 🥮 🍰

PR #364, opened 2022-02-07
feat: User profile status dialog

borichellow:
[Task-820](https://wearezeta.atlassian.net/browse/AR-820)

# What's new in this PR?
- added the dialog on user change availability status
- navigate to Settings, when user click on edit btn in the UserProfile screen

Screenshot_1644241907


app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileScreen.kt
```diff
@@ -328,6 +336,46 @@ private fun OtherAccountItem(account: OtherAccount, onClick: (String) -> Unit =
     }
 }
 
+@Composable
+private fun DialogContent(
```
gongracr, line 340: I feel this file is becoming too big. Should we extract this component to a separate file annotated with `@PrivatePackage`?
Also to improve redability:
```suggestion
private fun UserStatusChangeDialogContent(
```

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileState.kt
```diff
@@ -8,7 +10,55 @@ data class UserProfileState(
...
     val userName: String,
     val teamName: String, //maybe teamId is better here
-    val otherAccounts: List<OtherAccount>
+    val otherAccounts: List<OtherAccount>,
+    val dialogState: DialogState = DialogState.None
 )
 
 data class OtherAccount(val id: String, val avatarUrl: String, val fullName: String, val teamName: String? = null)
+
+sealed class DialogState {
```
gongracr, line 19: Maybe better if renamed to
```suggestion
sealed class StatusChangeDialogState {
```
Otherwise it indicates this class could be applied to any user dialog

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileState.kt
```diff
@@ -8,7 +10,55 @@ data class UserProfileState(
...
+
+    object None : DialogState()
+
+    sealed class StatusChange(
```
gongracr, line 23: Maybe rename to:
```suggestion
    sealed class DialogStatusInfo(
```

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileState.kt
```diff
@@ -8,7 +10,55 @@ data class UserProfileState(
...
+        @StringRes val title: Int,
+        @StringRes val text: Int,
+        val status: UserStatus,
+    ) : DialogState() {
```
gongracr, line 27: Shouldn't we make `None` an equivalent dialog state as the rest (Available, Busy, Away)? That way there is no need to have nested sealed classes and 2 different `None` and `StateNone` types of dialog.

Perhaps we could make `DialogStatusInfo` an interface that can be implemented by the different states.

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -34,9 +37,7 @@ class UserProfileViewModel @Inject constructor(private val navigationManager: Na
             )
         )
     )
-
-    val state: StateFlow<UserProfileState>
-        get() = _state
+        private set
```
gongracr, line 40: Nicee ✨

app/src/main/res/values/strings.xml
```diff
@@ -75,6 +75,16 @@ username</string>
     <string name="user_profile_status_none">None</string>
     <string name="user_profile_new_acc_text">New Team or Account</string>
 
+    <!-- User Profile stratus dialog-->
```
gongracr, line 78:
```suggestion
    <!-- User Profile status dialog-->
```

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -57,10 +58,41 @@ class UserProfileViewModel @Inject constructor(private val navigationManager: Na
...
+            }
+        }
+    }
+
+    fun changeStatusClick(status: UserStatus) {
+        if (userProfileState.status == status) return
+
+//        if (!shouldShowDialog) {
+//            changeStatus(status)
+//        } else {
```
saleniuk, line 88: maybe add a TODO to make it clear that it still needs to be implemented 😉

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -57,10 +58,41 @@ class UserProfileViewModel @Inject constructor(private val navigationManager: Na
...
 
     fun changeStatus(status: UserStatus) {
         //TODO
+        onDismissDialog()
+    }
+
+    fun doNotShowStatusDialogCheckChanged(doNotShow: Boolean) {
+        userProfileState.run {
+            if (dialogState is DialogState.StatusChange) {
+                userProfileState = copy(dialogState = dialogState.changeCheckBoxState(doNotShow))
```
gongracr, line 78: This is not really persisting the "Don't show again status dialog" checkbox state

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileScreen.kt
```diff
@@ -328,6 +336,46 @@ private fun OtherAccountItem(account: OtherAccount, onClick: (String) -> Unit =
...
+            ),
+            dismissButtonProperties = WireDialogButtonProperties(
+                onClick = onDismiss,
+                text = stringResource(id = android.R.string.cancel),
+                type = WireDialogButtonType.Secondary,
+            )
+        ) {
+            Row {
+                Checkbox(
+                    modifier = Modifier.align(Alignment.CenterVertically),
```
saleniuk, line 364: I think it's better to set `verticalAlignment = Alignment.CenterVertically` to the `Row` instead of setting it to its every child 😄

gongracr (changes requested):
Good job. Just requested some changes about nesting too many sealed classes and namings. We need to be mindful about code readability, and as redundant as possible so that we avoid future confusions with devs that have not touched this code before 🙏

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileState.kt
borichellow: hmmm, the word Dialog is useless here, as it's gonna be like `DialogState. DialogStatusInfo` (double dialog)

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileState.kt
gongracr: Imho this class should be dealing only with the specific User Status Availability dialogs. If you want to reuse this dialog class, it would be better for it to live in a different UI folder focused on dialogs or UI utils, and that can be generic enough for other components to reuse it. Then we will make all these `DialogState` objects/classes inherit or implement it.  Otherwise everything would be very tangled with the UserProfile case.

My suggestion was to implement something like this:
```
sealed class DialogState {

    abstract val title: String
    abstract val text: String
    abstract val status: UserStatus
    abstract var notShowAgain: Boolean

    data class StateAvailable(override var notShowAgain: Boolean = false) : DialogState() {
        override val title = stringResource(R.string.user_profile_change_status_dialog_available_title)
        override val text = stringResource(R.string.user_profile_change_status_dialog_available_text)
        override val status = UserStatus.AVAILABLE
    }

    data class StateBusy(override var notShowAgain: Boolean = false) : DialogState() {
        override val title = stringResource(R.string.user_profile_change_status_dialog_busy_title)
        override val text = stringResource(R.string.user_profile_change_status_dialog_busy_text)
        override val status = UserStatus.BUSY
    }

    data class StateAway(override var notShowAgain: Boolean = false) : DialogState() {
        override val title = stringResource(R.string.user_profile_change_status_dialog_away_title)
        override val text = stringResource(R.string.user_profile_change_status_dialog_away_text)
        override val status = UserStatus.AWAY
    }

    data class StateNone(override var notShowAgain: Boolean = false) : DialogState() {
        override val title = stringResource(R.string.user_profile_change_status_dialog_none_title)
        override val text = stringResource(R.string.user_profile_change_status_dialog_none_text)
        override val status = UserStatus.NONE
    }
}
```

app/src/main/kotlin/com/wire/android/datastore/UserDataStore.kt
```diff
@@ -0,0 +1,52 @@
...
+        //keys
+        private val SHOW_STATUS_RATIONALE_AVAILABLE = booleanPreferencesKey("show_status_rationale_available")
+        private val SHOW_STATUS_RATIONALE_BUSY = booleanPreferencesKey("show_status_rationale_busy")
+        private val SHOW_STATUS_RATIONALE_AWAY = booleanPreferencesKey("show_status_rationale_away")
+        private val SHOW_STATUS_RATIONALE_NONE = booleanPreferencesKey("show_status_rationale_none")
+    }
+
+    private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = PREFERENCES_NAME)
+
+    suspend fun donNotShowStatusRationaleAgain(status: UserStatus) {
```
gongracr, line 31:
```suggestion
    suspend fun donNotShowStatusRationaleAgain(status: UserStatus) {
```
```suggestion
    suspend fun dontShowStatusRationaleAgain(status: UserStatus) {
```

gongracr (approved):
Thanks for the changes. Now the code is definitely looking more self explanatory 🥞 ☕

PR #369, opened 2022-02-08
feat: Fix fab styles

borichellow:
[Task-997 ](https://wearezeta.atlassian.net/browse/AR-997)
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- updated icon in FAB 
- updated text style in FAB

Screenshot_1644341160

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/common/FloatingActionButton.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/FloatingActionButton.kt
@@ -1,25 +1,32 @@
 package com.wire.android.ui.common
 
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Add
 import androidx.compose.material3.ExtendedFloatingActionButton
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
-import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.filled.Add
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import com.wire.android.ui.theme.wireTypography
 
 @Composable
 fun FloatingActionButton(
     text: String,
-    onClick: () -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    icon: @Composable (() -> Unit)? = null,
+    onClick: () -> Unit
 ) {
     ExtendedFloatingActionButton(
         containerColor = MaterialTheme.colorScheme.primary,
         contentColor = MaterialTheme.colorScheme.onPrimary,
-        icon = { Icon(Icons.Filled.Add, "") },
-        text = { Text(text = text) },
+        icon = icon ?: { Icon(Icons.Filled.Add, "") },
+        text = {
+            Text(
+                text = text,
+                style = MaterialTheme.wireTypography.button01
+            )
+        },
         onClick = onClick,
         modifier = modifier
     )
--- a/app/src/main/kotlin/com/wire/android/ui/common/FloatingActionButton.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/FloatingActionButton.kt
@@ -14,13 +14,13 @@ import com.wire.android.ui.theme.wireTypography
 fun FloatingActionButton(
     text: String,
     modifier: Modifier = Modifier,
-    icon: @Composable (() -> Unit)? = null,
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/wire/android/ui/common/FloatingActionButton.kt
```diff
@@ -1,25 +1,32 @@
...
 import androidx.compose.ui.Modifier
+import com.wire.android.ui.theme.wireTypography
 
 @Composable
 fun FloatingActionButton(
     text: String,
-    onClick: () -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    icon: @Composable (() -> Unit)? = null,
```
trOnk12, line 17: what is the advantage of using Composable lambda here? Also I think it would be more readable to just assign a default value instead of making it nullable

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
```diff
@@ -28,7 +38,24 @@ fun ConversationRouter(conversationListViewModel: ConversationListViewModel = hi
...
+                        painter = painterResource(id = R.drawable.ic_conversation),
+                        contentDescription = stringResource(R.string.content_description_new_conversation),
+                        contentScale = ContentScale.FillBounds,
+                        colorFilter = ColorFilter.tint(MaterialTheme.colorScheme.onPrimary),
+                        modifier = Modifier
+                            .padding(start = dimensions().spacing4x, top = dimensions().spacing2x)
+                            .width(16.dp)
+                            .height(16.dp)
+                    )
+                },
```
trOnk12, line 55: ahhh I see it here a lamda allows you to do this as well

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
```diff
@@ -28,7 +38,24 @@ fun ConversationRouter(conversationListViewModel: ConversationListViewModel = hi
...
+            FloatingActionButton(
+                text = stringResource(R.string.label_new),
+                icon = {
+                    Image(
+                        painter = painterResource(id = R.drawable.ic_conversation),
+                        contentDescription = stringResource(R.string.content_description_new_conversation),
+                        contentScale = ContentScale.FillBounds,
+                        colorFilter = ColorFilter.tint(MaterialTheme.colorScheme.onPrimary),
+                        modifier = Modifier
+                            .padding(start = dimensions().spacing4x, top = dimensions().spacing2x)
```
trOnk12, line 51: maybe I was not following the discussion, but what does this spacing2x, spacing4x mean ?

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
```diff
@@ -28,7 +38,24 @@ fun ConversationRouter(conversationListViewModel: ConversationListViewModel = hi
...
+                            .width(16.dp)
```
trOnk12, line 52: move to dimensions

app/src/main/kotlin/com/wire/android/ui/common/FloatingActionButton.kt
saleniuk: If the icon should be optional then we don't want a default value and current solution with lambda and null is fine. The question is: should the icon really be optional or is it always on the FAB?

trOnk12: True true it should be optional but if you read further in the code it has a default value anyway

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
borichellow: basically, the idea here is to have such spaces in dimensions. At first I named it like `spacing2` which would be `2.dp` in regulare theme and whatever we need in others. But Gonso mentioned that such naming could be confusing, so I've added `x`.
So the idea here is to have names like: `spacing2x` where `spacing` - is just a name for the dimension, `2` - is 2.dp in regular theme `x` - is kinda multiplier sign. 
So if we decide that for tablets all the spacing should be doubled, we'll have `spacer2x = 4.dp`

trOnk12 (approved):
good job <3

PR #373, opened 2022-02-10
feat: Make home top bar scrollable

borichellow:
[Task-998](https://wearezeta.atlassian.net/browse/AR-998)
----

# What's new in this PR?
Made Home TopBar scrollable, so the search field hides on scroll down and shows again on scroll up

### Attachments (Optional)

https://user-images.githubusercontent.com/6539347/153409900-6d0f35f2-0bd8-476a-80b2-9463f6b443f2.mp4


app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
```diff
@@ -37,31 +45,50 @@ fun HomeTopBar(
...
+                shadowElevation = dimensions().topBarElevationHeight
+            ) {
+                SearchBarUI(
+                    modifier = Modifier.background(MaterialTheme.colorScheme.background),
+                    placeholderText = stringResource(R.string.search_bar_hint, titleText.lowercase())
+                )
+            }
         }
+
+        Surface(shadowElevation = firstLineElevation) {
```
saleniuk, line 74: I was thinking of creating something similar to `SurfaceBackgroundWrapper` - `TopBarElevationWrapper` with parameters `withElevation` and `content` - we could wrap every TopBar with this simple composable and keep the TopBar elevation across the whole app unified

what do you think about that?

app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
```diff
@@ -15,6 +18,23 @@ class HomeViewModel
...
+    fun updateScrollPosition(newScrollIndex: Int) {
+        if (newScrollIndex == lastScrollIndex) return
+        if (newScrollIndex > lastScrollIndex + 1) {
+            _scrollDown.value = true
+            lastScrollIndex = newScrollIndex
+        } else if (newScrollIndex < lastScrollIndex) {
+            _scrollDown.value = false
+            lastScrollIndex = newScrollIndex
+        }
+    }
```
saleniuk, line 36: I'm not sure if LiveData has a `scan` operator, but `Flow` has it so by switching to Flow and using it we wouldn't need to have the `var lastScrollIndex`

We already use Flow so I feel like we should stick to the single solution and keep using only this one and we don't need LiveData libraries in the app 🤔

app/src/main/kotlin/com/wire/android/ui/home/HomeNavigation.kt
```diff
@@ -1,30 +1,45 @@
...
-fun HomeNavigationGraph(navController: NavHostController, startDestination: String?) {
-    NavHost(navController, startDestination = startDestination ?: HomeNavigationItem.Conversations.route) {
-        HomeNavigationItem.all
-            .forEach { item ->
-                composable(route = item.route, content = item.content)
-            }
+fun HomeNavigationGraph(
+    navController: NavHostController,
+    startDestination: String?,
+    updateScrollPosition: (Int) -> Unit = {}
```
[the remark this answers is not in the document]
borichellow: about what if navigation: this is not a problem, as `Archive` has `isSearchable = false`, so the search field is always hidden there. 
And about all others: good point, thanks!

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
saleniuk: TobBars are different in terms of its content inside, but visually I believe they all should be consistent and I think that Wolfgang wants all TobBars to work the same way throughout the whole app, but that may be just my own interpretation of his words:

> If the user starts scrolling the list down, a little shadow on the bottom of the appbar header is shown. 
> If the list is scrolled all to the top, the shadow is hidden.

If we create custom wrapper then we can set the default value, so the usage will look more like just:
`ElevationWrapper { /*Content here*/ }`
😄

app/src/main/kotlin/com/wire/android/ui/home/HomeNavigation.kt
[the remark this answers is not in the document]
saleniuk: For now I think it's fine as we have only one scrollable screen that uses it, in the future, if we would like to add another one then we would have to keep in mind that this value should be cleared when changing from one scrollable to another scrollable screen, but I'm fine with the current solution as long as we all understand what would need to be done in such case 😄

trOnk12:
I am not sure if it is a good idea to pass the updateScrollPosition()  lambda all the way to each view, that should update the scroll position, but I am not sure how we could solve it, one way I could think of is to use compositionLocalOf() which would provide some object that is accesible from the whole app and would update the scroll position of the HomeScreen, but it is just a quick guess, maybe you could try to find a way to omit passing the scrollposition callback alll the way through each view ?

app/src/main/kotlin/com/wire/android/ui/home/HomeNavigation.kt
```diff
@@ -52,27 +63,27 @@ sealed class HomeNavigationItem(
...
         title = R.string.vault_screen_title,
-        content = { VaultScreen() },
+        content = { { VaultScreen() } }
     )
 
     object Archive : HomeNavigationItem(
         route = HomeDestinations.archive,
         title = R.string.archive_screen_title,
-        content = { ArchiveScreen() },
+        content = { { ArchiveScreen() } }
```
trOnk12, line 86: I am not sure if it is a good thing to pass typealias UpdateScrollPosition = (Int) -> Unit as a parameter to the lambda but do nothing with it, is this updateScrollPOsition functionality going to be used for all the views in the future, if not maybe we have to distinct it somehow with the typealias that would not take this typealias UpdateScrollPosition = (Int) -> Unit  as parameter ? also not really know a quick fix for this one

borichellow: as far as we don't have a designs for Archive and Vault we can try to guess :) 
I believe we'll have search field in Archive too, so I assume - it will be hidable there too, so it will be used there.

saleniuk: >I am not sure if it is a good idea to pass the updateScrollPosition() lambda all the way to each view, that should update the scroll position, but I am not sure how we could solve it, one way I could think of is to use compositionLocalOf() which would provide some object that is accesible from the whole app and would update the scroll position of the HomeScreen, but it is just a quick guess, maybe you could try to find a way to omit passing the scrollposition callback alll the way through each view ?

☝️ I attached also this comment as it feels connected and here we can reply 😄 

The approach with `staticCompositionLocalOf` I think should be used for things that are used throughout the whole app, so I'm not sure - passing it all the way to the view ensures that we always use the correct one and that we always have it handled, but if it's really going to be more widely used across a larger part of the app than just this particular case then it might be worth trying at least. We can make it versatile - every child that can scroll should use such local composition and update its scroll value and every parent that wants to make use of such value should provide its own implementation of it which would then be available to all his children. The problem with this is that it can be provided again by the child and then the scroll updates will no longer reach the original parent. What do you think, is it worth trying?

trOnk12: True you are probably right :p I dont not know if local conpisition is the best approach here just was giving a guess to maybe a solution that could work but i agree with you. I just have a feeling that there could be a better solution to assign the scroll position with less overhead but if passing the lamda like this is the only option we got I am fine with it :p just my Gut feeling saying but i trust Boris on this one :p

trOnk12: Btw. I like your approach maybe we can just remember not to provide composition local second time :p

trOnk12: forget it , :P boris already handled this discussion in the newest PR

saleniuk: you mean in the latest commit? 😄  
I'm fine also with the approach Boris proposed, where dependency injection is used instead of local composition

saleniuk (approved):

trOnk12 (commented):
In my opinion it looks nicer now, if Michal has no comments left, lgtm 🚀 ❤️

PR #384, opened 2022-02-16
feat: Add get Conversations UseCase from kalium

borichellow:
# What's new in this PR?

Added getConversationsUseCase from kalium to fetch and display conversations 

### Attachments (Optional)

photo5217566210537930108

### DIFF
--- a/app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
+++ b/app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
@@ -2,21 +2,19 @@ package com.wire.android.di
 
 import android.content.Context
 import com.wire.android.util.DeviceLabel
+import com.wire.kalium.logic.CoreLogic
+import com.wire.kalium.logic.feature.auth.AuthSession
+import com.wire.kalium.logic.feature.session.CurrentSessionResult
 import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
-import dagger.hilt.android.qualifiers.ApplicationContext
-import dagger.hilt.components.SingletonComponent
-import com.wire.kalium.logic.CoreLogic
 import dagger.hilt.android.components.ViewModelComponent
+import dagger.hilt.android.qualifiers.ApplicationContext
 import dagger.hilt.android.scopes.ViewModelScoped
+import dagger.hilt.components.SingletonComponent
+import kotlinx.coroutines.runBlocking
 import javax.inject.Qualifier
 import javax.inject.Singleton
-import com.wire.kalium.logic.feature.auth.AuthSession
-import com.wire.kalium.logic.feature.session.CurrentSessionResult
-import dagger.hilt.android.scopes.ActivityRetainedScoped
-import kotlinx.coroutines.runBlocking
-import java.lang.IllegalStateException
 
 @Qualifier
 @Retention(AnnotationRetention.BINARY)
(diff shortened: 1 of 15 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -24,6 +32,7 @@ import javax.inject.Inject
 @HiltViewModel
 class ConversationListViewModel @Inject constructor(
     private val navigationManager: NavigationManager,
+    private val getConversationUseCase: GetConversationsUseCase,
```
yamilmedina, line 35: Maybe we can call this var just `getConversations` so we can then invoke the function without the dot invoke -> `getConversations()` That way it looks just like any other function

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -35,26 +44,29 @@ class ConversationListViewModel @Inject constructor(
...
-            missedCalls = mockMissedCalls,
-            callHistory = mockCallHistory,
-            unreadMentions = mockUnreadMentionList,
-            allMentions = mockAllMentionList,
-            unreadMentionsCount = 12,
-            missedCallsCount = 100,
-            newActivityCount = 1
-        )
+        viewModelScope.launch {
+            getConversationUseCase.invoke()
```
yamilmedina, line 52: Same comment as above, maybe we can call this just `getConversations()` for simplicity

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -106,4 +118,31 @@ class ConversationListViewModel @Inject constructor(
...
+                        membership = Membership.None,
+                        isLegalHold = true
+                    ),
+                    conversationsId = conversation.id
+                )
+            )
+        } else {
+            GeneralConversation(
+                ConversationType.GroupConversation(
+                    groupColorValue = 0xFFFF0000,
```
yamilmedina, line 137: This color could be extracted or reuse one from `WireColorPalette` class

gongracr: Yes, we should use `MaterialTheme.wireColorScheme` instead

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -50,9 +48,11 @@ class UseCaseModule {
     // TODO: can be improved by caching the current session in kalium or changing the scope to ActivityRetainedScoped
     fun currentSessionProvider(@KaliumCoreLogic coreLogic: CoreLogic): AuthSession {
         return runBlocking {
-            return@runBlocking when(val result = coreLogic.getAuthenticationScope().session.currentSession.invoke()) {
+            return@runBlocking when (val result = coreLogic.getAuthenticationScope().session.currentSession.invoke()) {
                 is CurrentSessionResult.Success -> result.authSession
-                else -> {throw IllegalStateException("no current session where found")}
+                else -> {
+                    throw IllegalStateException("no current session where found")
```
gongracr, line 54:
```suggestion
                    throw IllegalStateException("no current session was found")
```

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/model/ConversationItem.kt
```diff
@@ -2,22 +2,29 @@ package com.wire.android.ui.home.conversationslist.model
...
-        val groupName: String
+        val groupName: String,
+        override val conversationsId: ConversationId
     ) : ConversationType()
 
     data class PrivateConversation(
         val userInfo: UserInfo,
-        val conversationInfo: ConversationInfo
+        val conversationInfo: ConversationInfo,
+        override val conversationsId: ConversationId
```
gongracr, line 27:
```suggestion
        override val conversationId: ConversationId
```

gongracr (dismissed):
Looks good enough until we find a way on how to generate the 1on1 conversation title as this seems to be not a straightforward task (maybe finding out why the private conversation `name` field keeps coming empty).

gongracr (approved):
👏 🎬 🎥

PR #388, opened 2022-02-18
fix: fixes for issues raised in playtest session (AR-1085) (AR-1092)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1085" title="AR-1085" target="_blank">SubtaskAR-1085</a>  Change status - None pop up text should be fixed
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1085
https://wearezeta.atlassian.net/browse/AR-1092
https://wearezeta.atlassian.net/browse/AR-1121

Solves some issues regarding the user profile screen raised in previous playtest sessions.
Also for the checkbox, created a component called, `WireLabelledCheckbox` so we can have this behavior across other screens when needed.

### Attachments (Optional)

fix-checkbox

https://user-images.githubusercontent.com/5806454/154973458-0b5c4896-b6b2-4886-a9a5-d3573a5f1f77.mp4

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
```diff
@@ -0,0 +1,50 @@
...
+            checked = checked,
+            onCheckedChange = null // null since we are handling the click on parent
+        )
+
+        Spacer(modifier = Modifier.size(MaterialTheme.wireDimensions.spacing8x))
+
+        Text(
+            text = label,
+            style = MaterialTheme.wireTypography.body01
+        )
```
trOnk12, line 48: I wonder if it is going to look correctly when the text is longer than one line, or will jump into next line on smaller devices, what you think, worth to take a look at in this iteration ?

app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
```diff
@@ -0,0 +1,50 @@
...
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalArrangement = horizontalArrangement,
+        modifier = modifier
+            .clickable(onClick = { onCheckClicked(!checked) })
+            .requiredHeight(ButtonDefaults.MinHeight)
+            .wrapContentWidth()
+    ) {
+        Checkbox(
+            checked = checked,
+            onCheckedChange = null // null since we are handling the click on parent
```
borichellow, line 40: what if you use `onCheckClicked` here too? Will I make a double-click?
I'm asking cause AFAIK we suppose to support Talkback (part of Android Accessibility Suit) in the future, and non-clickable checkbox could cause problems there.

app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.compose.ui.Modifier
+import com.wire.android.ui.theme.wireDimensions
+import com.wire.android.ui.theme.wireTypography
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun WireLabelledCheckbox(
+    label: String,
+    checked: Boolean,
+    onCheckClicked: ((Boolean) -> Unit),
```
saleniuk, line 26: I'm wondering if it would be better to have `onCheckedChanged` - update the `checked` value inside this item and just emit the current updated value instead of "click event" where we have to update it from outside

app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
yamilmedina: Yeah! You are right! Let me check how talkback is behaving with this component and make a proposal accordingly

app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
yamilmedina: Good catch, I made some adjustment to be able to break line and also to go for ellipsis if the text is really big

app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
yamilmedina: I made some adjustment and tested with talkback after doing some improvements, good catch thanks. 
Now it's working, recognizing it as a checkbox and using ie: "double tap to toggle" when the screen reader is passing =)

app/src/main/kotlin/com/wire/android/ui/common/WireLabelledCheckbox.kt
yamilmedina: I think it could be a good approach also, but I think I'll leave it as it is right now, because the current usages are depending on this callback, I'll mark a `// TODO` to improve, and we can come back later if we need another usage type for example

trOnk12 (approved):
Looks good to me :)

PR #396, opened 2022-02-23
feat: fill profile user profile data on screen from backend (AR-1118)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1118" title="AR-1118" target="_blank">SubtaskAR-1118</a>  Show logged-in user data from use case 
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1118

- This calls and load user's data from the real usecase (backend/local persisted info) into the user profile screen
- Still some pending work to do to load a default avatar when the user doesn't have one, should be addressed later when downloading assets is'ts working

### Attachments (Optional)
Screenshot_1645622696

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileScreen.kt
```diff
@@ -208,10 +202,12 @@ private fun UserProfileScreen(
                     userStatus = status,
                     onStatusClicked = onStatusClicked
                 )
-                OtherAccountsList(
-                    otherAccounts = otherAccounts,
-                    onAddAccountClick = onAddAccountClick
-                )
+                if (!state.otherAccounts.isNullOrEmpty()) {
```
trOnk12, line 205: will it ever be empty ?

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -14,48 +14,49 @@ import com.wire.android.navigation.NavigationCommand
...
+    init {
+        viewModelScope.launch {
+            getSelf().collect {
+                userProfileState = SelfUserProfileState(
+                    avatarBitmap = null,
+                    status = UserStatus.AVAILABLE,
+                    fullName = it.name!!,
+                    userName = it.handle!!,
+                    teamName = it.team ?: "",
+                    otherAccounts = emptyList()
```
trOnk12, line 54: it is emptyList be default I think

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -14,48 +14,49 @@ import com.wire.android.navigation.NavigationCommand
...
+                )
+            }
+        }
+    }
```
trOnk12, line 58: we are not handeling here anywhere loading of the data as well as handeling error, do not know if that is something we should do for now, if not maybe add TODO: to stuff that we need to adjust later on

trOnk12 (dismissed):
If this is for the demo I am fine with how it looks, add TODO: for the stuff we need to keep in mind and we will adjust it once the use case is finished on the Kalium side

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -14,48 +14,49 @@ import com.wire.android.navigation.NavigationCommand
...
 
+    init {
+        viewModelScope.launch {
+            getSelf().collect {
+                userProfileState = SelfUserProfileState(
+                    avatarBitmap = null,
+                    status = UserStatus.AVAILABLE,
+                    fullName = it.name!!,
+                    userName = it.handle!!,
+                    teamName = it.team ?: "",
```
trOnk12, line 53: I wonder if we have to somehow handle it on the UI in case the teamname is null ? maybe we should align the status more to be closer to the user name ? because as for now there is a bigger space in between user name and the team name

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
yamilmedina: Thanks! yes, definitely we should have some initial state for loading and play with that. Will add a TODO for in the meanwhile

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileScreen.kt
yamilmedina: In fact, it will be empty by default, changed to `isNotEmpty` 👍

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
yamilmedina: Totally agree! I move the component show logic accordingly to this parameter instead, so replaced by just an assignment

trOnk12 (dismissed):
lgtm 🚀

trOnk12 (approved):

PR #401, opened 2022-02-24
feat: custom server config deepLink [AR-943]

mchenani:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-943" title="AR-943" target="_blank">SubtaskAR-943</a>  point deep links to open the splash screen
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

### Issues
[AR-943]

### Solutions
Point deepLink to the app, 
### Testing

Manual test on computer:
`./adb shell am start -W -a android.intent.action.VIEW -d "wire://access/?config=https://nginz-https.diya.wire.link/deeplink.json"`

**On Device Testing:**
**Only with Firefox**: open below link, and from the context menu click open app:
`wire://access/?config=https://nginz-https.diya.wire.link/deeplink.json`

(chrome is not allowing users to paste non-url links to open apps, only hyperlinks are allowed)

----

----

----
##### References

[AR-943]: https://wearezeta.atlassian.net/browse/AR-943


app/src/main/kotlin/com/wire/android/navigation/NavigationGraph.kt
```diff
@@ -6,16 +6,19 @@ import androidx.compose.runtime.Composable
...
-fun NavigationGraph(navController: NavHostController, startDestination: String) {
-    NavigationItem.globalNavigationItems.also { navItems ->
-        NavHost(navController, startDestination) {
-            navItems.forEach { item ->
-                composable(route = item.route, content = item.content, arguments = item.arguments)
-            }
+fun NavigationGraph(navController: NavHostController, startDestination: String, serverConfig: ServerConfig) {
+    NavHost(navController, startDestination) {
+        NavigationItem.values().onEach { item ->
+            if (item == NavigationItem.Login)
```
saleniuk, line 18: I think it would be better not to have such `if` - soon we'll have more screens that need `ServerConfig` so for me it would be better to update `NavigationItem.content` to take it as a parameter. We don't use `NavBackStackEntry` anyway so we could change it to be `val content: @Composable (ServerConfig) -> Unit`

saleniuk:
I think you should resolve conflicts, because it seems to me that you have some changes that have already been done 🤔

app/src/main/kotlin/com/wire/android/navigation/NavigationGraph.kt
mchenani: Thanks @saleniuk, honestly me too :D but this solution is temporary and with @gongracr 's suggestion I added this, because now we have a big issue with passing arguments for start screens. At the moment this is the only working solution [ other than singletons :d (run Forrest, run!)]

saleniuk: I'm not sure what kind of issue with passing arguments it really is, I'm just suggesting that we should get rid of this not really pretty `if` and instead just pass this as a parameter to `NavigationItem`. 
In my approach I created a data class `ContentParams` that contains all that's needed and passed it as a parameter of the content creation function like that: `val content: @Composable (ContentParams) -> Unit`, and then they can be added to `NavHost` this vay:
```
NavigationItem.values().forEach { destination ->
            composable(route = .getCanonicalRoute()) {
                    item.content(ContentParams(...))
            }
        }
```

gongracr: I am afraid we can't change the type from NavBackStackEntry to ServerConfig on the composable content. The reason is that we extract the composable arguments from the backStackEntry item from some elements on the `NavigationItem` class to then inject them into the screens, for example in the `Conversation` item:
```
Conversation(
        primaryRoute = CONVERSATION,
        canonicalRoute = "$CONVERSATION/{$EXTRA_CONVERSATION_ID}",
        content = { ConversationScreen(it.arguments?.getString(EXTRA_CONVERSATION_ID), hiltViewModel()) },
        arguments = listOf(
            navArgument(EXTRA_CONVERSATION_ID) { type = NavType.StringType }
        )
    ) {
        override fun getRouteWithArgs(arguments: List<Any>): String {
            val conversationId: ConversationId? = arguments.filterIsInstance<ConversationId>().firstOrNull()
            return if (conversationId != null) "$primaryRoute/${conversationId.toJsonString()}" else primaryRoute
        }
    };

To me, there would be 2 ideal approaches with this problem:

1. Go with this approach (maybe substitute the `if` `else` for a `when` clause?) as for now we are only injecting this `ServerConfig` in the `LoginScreen`. It seems we will be using it only there for now.
2. Extract the ServerConfig params in the WelcomeScreen / WireActivity / WireActivityViewModel and persist them in memory. Then add a singleton `GetCurrentServerConfigUseCase` that can be called anywhere when needed.

Not really a fan of passing vagabonding args in between many classes, so if we are sure the `LoginScreen` will be the only one dealing with it, I would be ok living with that for the time being.

Comment (trOnk12) [reply app/src/main/kotlin/com/wire/android/navigation/NavigationGraph.kt:18]: I will add my 5 cents to this :P as I also addressed this issue a bit with this PR : https://github.com/wireapp/wire-android-reloaded/pull/395

I am not sure if it is good idea that the UI knows about some buisness stuff like serverConfig , to me it looks a bit like a work-around rather than a solution :P what we could do for objects that we want to access globally is create a HomeState that would be passed to the content and from there we can pass it further down the tree.Another solution that would be suitable here is I think since we are using Dagger it would be a good idea to pass this ServerConfig to for example AppModule as a constructor inside the Application or WireActivity and create a provider for that so that we are able to inject it globally inside the viewmodels

Comment (saleniuk) [reply app/src/main/kotlin/com/wire/android/navigation/NavigationGraph.kt:18]: We can add `NavBackStackEntry` to the `ContentParams`:
```
class ContentParams(
    val navBackStackEntry: NavBackStackEntry,
    …
)
```
in `NavigationGraph` it would be like:
```
NavigationItem.values().onEach { item ->
    composable(
        route = item.getCanonicalRoute(),
        content = { navBackStackEntry ->
            item.content(
                ContentParams(
                    navBackStackEntry = navBackStackEntry,
		…
                )
            )
        }
    )
}
```
and to create the `HomeScreen`:
```
 content = { HomeScreen(it.navBackStackEntry.arguments?.getString(EXTRA_HOME_TAB_ITEM), hiltViewModel()) }, 
```

Comment (mchenani) [reply app/src/main/kotlin/com/wire/android/navigation/NavigationGraph.kt]: @saleniuk thanks for the suggestion, I just pushed new changes based on your recommendations.🥳✌️

File: app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt [hunk 2]
```diff
@@ -176,3 +179,8 @@ fun String.parseIntoQualifiedID(): QualifiedID {
     val components = split("@")
     return QualifiedID(components.last(), components.first())
 }
+
+data class ContentParams(
+    val navBackStackEntry: NavBackStackEntry,
+    val arguments: List<Any> = emptyList()
```
saleniuk, line 186: I was thinking about declaring other variables also directly like:
```
data class ContentParams(
    val navBackStackEntry: NavBackStackEntry,
    val serverConfig: ServerConfig
)
```
This way (using `List<Any>`) is also fine, actually more versatile, it just makes this extra step of checking if the list contains the necessary argument type - I think it's easier for devs if they see which arguments are necessary when creating the instance of `ContentParams` but it may be just my personal preference 😄

saleniuk (approved):
one comment that may be more my preference 😅  other than that I think it's ok 💪

PR #403, opened 2022-02-25
fix: message composer

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Message composer was taking in a TextFieldValue as a argument into the onMessageChanged lambda as well as messageText, moving the responsibility to care about the state of the input to the hoisting parent. Now the responsibility is moved to MessageComposer. Making the parent only care about the String it is hoisting.

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/di/ClientScopeProvider.kt
+++ b/app/src/main/kotlin/com/wire/android/di/ClientScopeProvider.kt
@@ -1,7 +1,6 @@
 package com.wire.android.di
 
 import com.wire.kalium.logic.CoreLogic
-import com.wire.kalium.logic.feature.auth.AuthSession
 import com.wire.kalium.logic.feature.client.ClientScope
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
@@ -9,13 +8,13 @@ import dagger.assisted.AssistedInject
 
 class ClientScopeProvider @AssistedInject constructor(
     @KaliumCoreLogic private val coreLogic: CoreLogic,
-    @Assisted private val authSession: AuthSession
+    @Assisted private val userId: String
 ) {
     val clientScope: ClientScope
-        get() = coreLogic.getSessionScope(authSession).client
+        get() = coreLogic.getSessionScope(userId).client
 
     @AssistedFactory
     interface Factory {
-        fun create(authSession: AuthSession): ClientScopeProvider
+        fun create(userId: String): ClientScopeProvider
     }
 }
--- a/app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
+++ b/app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
@@ -27,7 +27,7 @@ annotation class KaliumCoreLogic
 
 @Qualifier
 @Retention(AnnotationRetention.BINARY)
-annotation class CurrentSession
+annotation class CurrentAccount
 
 @Module
 @InstallIn(SingletonComponent::class)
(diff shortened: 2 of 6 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewModel.kt
```diff
@@ -64,17 +63,16 @@ class ConversationViewModel @Inject constructor(
...
 
     fun sendMessage() {
         viewModelScope.launch {
-            val messageText = conversationViewState.messageText
             //TODO what if conversationId is null???
-            sendTextMessage(conversationId!!, messageText.text)
-            conversationViewState = conversationViewState.copy(messageText = messageText.copy(""))
+            sendTextMessage(conversationId!!, conversationViewState.messageText)
         }
+        conversationViewState = conversationViewState.copy(messageText = "")
```
gongracr, line 75: If the message sending action fails we still want to remove the text in the input field?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
```diff
@@ -53,21 +53,42 @@ import com.wire.android.ui.theme.wireColorScheme
...
-        messageText = messageText,
-        onMessageChanged = onMessageChanged,
+        // we are setting the TextFieldValue by copying the current value and updating the text
+        // in order to keep the correct state of cursor position etc ...
+        messageText = messageComposerState.textFieldValue.copy(text = messageText),
+        onMessageChanged = {
+            // we are setting it immediately in the UI first
+            messageComposerState.textFieldValue = it
+            // we are hoisting the TextFieldValue text value up to the parent
+            if (messageText != it.text) {
```
gongracr, line 88: I don't this logic is necessary, since the `BasicTextField` is doing it for us already

gongracr (approved):
Looking good, but I didn't have the chance to test it on a real device yet. Up to you whether you feel confident enough to merge it and include it for next demo build @trOnk12 , if so please remember to update the kalium reference to the most recent commit in `develop` there too.
GJ in any case 🍨

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewModel.kt
trOnk12: actually good question, I do not think it was in the designs or we did not think about yet, I think in general we should give more thought to how to handle errors and progress on this screen

trOnk12: after sending the message we would like to clear the message input text field, do not know if that is what you mean ?

gongracr: I meant that if the message send action fails, then why should we also clear the input field? Shouldn't we check if the result was succesful before doing so?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
```diff
@@ -53,21 +53,42 @@ import com.wire.android.ui.theme.wireColorScheme
 import com.wire.android.ui.theme.wireDimensions
 import com.wire.android.ui.theme.wireTypography
 
+
+/*
+* Message composer is a UI widget that handles the UI logic of sending of the messages,
+* it is a wrapper around the "hosting" widget. It takes in a messageText and
+* exposes a onMessageChanged lambda, giving us the option to control it's Message Text outside of the Widget.
+* it also exposes other lambda's giving us the option to handle the different message actions
+* */
```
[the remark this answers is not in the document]
gongracr: No change?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
gongracr: should we then remove this line?

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewModel.kt
trOnk12: the way it works on facebook for example is that the textfield is cleared but when the message sending fails the failure status is shown on the message itself, think this is a question to Wolfgang :P

gongracr: ok, not a blocker for now then, but maybe something to revisit once the whole chat funcionality is up and running

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
trOnk12: we actually need this line otherwise it will nto be propagated to the parent

PR #412, opened 2022-03-03
fix: coalesce to empty string name & handle in profile

tmspzz:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When name or handle are not set, the app crashes.

### Solutions

Coalesce to empty string `name` and `handle`

### Testing

Register an account on web and close the application before setting a username.
Login into the app. The app doesn't crash when accessing the profile screen.

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
@@ -44,12 +44,12 @@ class UserProfileViewModel @Inject constructor(
             getSelf().collect {
                 userProfileState = SelfUserProfileState(
                     status = UserStatus.AVAILABLE,
-                    fullName = it.name!!,
-                    userName = it.handle!!,
+                    fullName = it.name.orEmpty(),
+                    userName = it.handle.orEmpty(),
                     teamName = it.team,
                     // Add some mocked team
                     otherAccounts = listOf(
-                        OtherAccount("someId", "", it.name!!, "Wire Swiss GmbH"),
+                        OtherAccount("someId", "", it.name.orEmpty(), "Wire Swiss GmbH"),
                         OtherAccount("someId", "", "B. A. Baracus", "The A-Team"),
                     )
                 )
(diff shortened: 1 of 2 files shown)


gongracr (dismissed):
LGTM 🍑

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -44,12 +44,12 @@ class UserProfileViewModel @Inject constructor(
             getSelf().collect {
                 userProfileState = SelfUserProfileState(
                     status = UserStatus.AVAILABLE,
-                    fullName = it.name!!,
-                    userName = it.handle!!,
+                    fullName = it.name ?: "",
+                    userName = it.handle ?: "",
```
vitorhugods, line 48: Not a real suggestion for improvement, as I don't have any preference and we don't have a standard for this in the team. But just FYI:

Kotlin allows us to call functions on optionals. So things like `myString.orEmpty()` or `myString.isNullOrEmpty()` exist in the stdlib.

They are defined as extension functions of nullable types:

```kotlin
fun String?.isNullOrEmpty(): Boolean = this == null || this.isEmpty()

fun String.orEmpty(): String = this ?: ""
```

I guess we could do the same on swift:

```swift
extension Optional where Wrapped: String {
    var isNilOrEmpty: Bool {
        return self?.isEmpty ?? true
    }
}
```

Having two ways of doing the same thing is not always nice, but 🤷🏼‍♂️
It exists anyway.

vitorhugods (approved):

yamilmedina (approved):
Looks good to me! 🎸 🚀

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -44,12 +44,12 @@ class UserProfileViewModel @Inject constructor(
             getSelf().collect {
                 userProfileState = SelfUserProfileState(
                     status = UserStatus.AVAILABLE,
-                    fullName = it.name!!,
-                    userName = it.handle!!,
+                    fullName = it.name.orEmpty(),
```
MohamadJaara, line 47: the user `fullname` should not be nullable at the first place, since a user can't register without a name and if it's null that means there is an error somewhere

vitorhugods (approved):

PR #414, opened 2022-03-03
feat: "create personal account" - add "code" UI and adapt to reuse steps [AR-1069]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1069" title="AR-1069" target="_blank">SubtaskAR-1069</a>  implement the registration personal flow UI
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

User needs to be able to fill out a personal account creation flow.
Steps needs to be reusable for both cases: create personal account and create team.

### Solutions

- "enter code" screen UI with error handling
  - `CodeScreen` as a next step in the "create personal account" flow
  - `CodeViewModel` with minimal implementation needed for the UI and navigation within creation flow to work
  - `CodeTextField` - reusable and adjustable input for code - single TextField with `decorationBox` styled according to the designs, every digit is in its own rectangle
- refactored flow and ViewModels to be able to reuse them
  - ViewModel interfaces for every step of the flow that can be reused between account creation flows
  - implementation of these steps ViewModels inside the `CreatePersonalAccountViewModel`
  - separate states for every step combined in the `CreatePersonalAccountViewModel`
  - each step screen is more adjustable, allows to customize the title, texts, icons according to the specific creation flow

### Attachments (Optional)

https://user-images.githubusercontent.com/30429749/156624470-35fbb684-efe0-4170-8543-aaf719e2344e.mp4

----

----

----
##### References


trOnk12:
Would it be possible to give abit more explanation on the implementation details with this PR?

app/src/main/kotlin/com/wire/android/ui/authentication/create/code/CodeScreen.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.wire.android.ui.theme.wireDimensions
+import com.wire.android.ui.theme.wireTypography
+
+@Composable
+fun CodeScreen(viewModel: CodeViewModel, title: String) {
+    CodeContent(
+        state = viewModel.codeState,
+        title = title,
+        onCodeChange = viewModel::onCodeChange,
+        onResendCodePressed = viewModel::onResendCodePressed,
```
trOnk12, line 40: small detail I think it is a personal test, but a viewmodel function should not indicate the view action, I do not think it matters for the Viewmodel if the button is pressed or long hold or a switch or a toggle is used, resendCode will be enough I think

app/src/main/kotlin/com/wire/android/ui/authentication/create/code/CodeScreen.kt
```diff
@@ -0,0 +1,113 @@
...
+                style = MaterialTheme.wireTypography.body01,
+                modifier = Modifier.fillMaxWidth().padding(
+                        horizontal = MaterialTheme.wireDimensions.spacing16x,
+                        vertical = MaterialTheme.wireDimensions.spacing24x
+                    )
+            )
+            Spacer(modifier = Modifier.weight(1f))
+            Column(horizontalAlignment = Alignment.CenterHorizontally,) {
+                CodeTextField(
+                    codeLength = CodeViewModel.CODE_LENGTH,
```
trOnk12, line 71: I am not sure if such information should belong to the viewmodel L ?

app/src/main/kotlin/com/wire/android/ui/authentication/create/details/DetailsScreen.kt
```diff
@@ -44,13 +47,14 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
 @Composable
-fun DetailsScreen(viewModel: CreatePersonalAccountViewModel) {
+fun DetailsScreen(viewModel: CreatePersonalAccountViewModel, title: String) {
```
trOnk12, line 50: why is this not inside the state object ?

app/src/main/kotlin/com/wire/android/ui/authentication/create/personalaccount/CreatePersonalAccountViewModel.kt
```diff
@@ -10,81 +10,110 @@ import androidx.lifecycle.viewModelScope
...
 import javax.inject.Inject
 
-@Suppress("MagicNumber")
+@Suppress("MagicNumber", "TooManyFunctions")
 @OptIn(ExperimentalMaterialApi::class)
 @HiltViewModel
 class CreatePersonalAccountViewModel @Inject constructor(
     private val navigationManager: NavigationManager,
-) : ViewModel() {
+) : ViewModel(), OverviewViewModel, EmailViewModel, DetailsViewModel, CodeViewModel {
```
trOnk12, line 30: I am a bit confused why are we extending those interfaces ?

app/src/main/kotlin/com/wire/android/ui/authentication/create/code/CodeScreen.kt
saleniuk: Sure, right 🚀

app/src/main/kotlin/com/wire/android/ui/authentication/create/code/CodeScreen.kt
saleniuk: ViewModel also needs to know if the code is fully filled, but I managed to change it a bit - now `onCodeChange` returns value and flag indicating if it's filled

app/src/main/kotlin/com/wire/android/ui/authentication/create/details/DetailsScreen.kt
saleniuk: These screens can have different "title" (text shown on the `TopAppBar`) depending on the flow it's used in (personal account or team creation) so it's the parameter coming from outside the screen and not its inner state

app/src/main/kotlin/com/wire/android/ui/authentication/create/personalaccount/CreatePersonalAccountViewModel.kt
saleniuk: Thanks to that, every screen/step can work more independently and be a part of multiple flows - each interface is a set of methods that are needed for this step's UI to work and provide the required data to the flow's ViewModel which needs to gather data from multiple screens in order to execute requests

trOnk12: all righty, I am not the biggest fan of Inheritance within the ViewModel, is there no way implementing it using Composition (not the jetpack one :P) I think the problem with this approach is adding the abstraction using the interfaces and also passing CreatePersonalAccountViewModel  to the stuff where in the end we just use emailState for example and couple of methods within CreatePersonalAccountViewModel. If we used Composition we eliminate the interfaces and also we use what we really need and not pass more than we need. Do not know how hard it would be hard to achieve it with the code we currently have and how much it will improve the code or if it even will improve anything :P just a gut feeling saying it may be worth to think about Composition instead of Inheritance

saleniuk: I thought about not using inheritance, but these ViewPagers will never be fully independent and in different flows they will e.g. make different requests or require different data from other ViewPagers, also in “create team” flow there will be an extra step between “details” and “code”. So it's impossible to completely separate them, we would have to have some abstraction built over this anyway. In practice, such ViewModels would only store states, which would be edited by some other element that would perform requests and manage navigation. We could accomplish all this with many `Flow`s, but that could overcomplicate the code, in the current form I feel it's cleaner. Is this okay for you or do you have idea for a specific solution?

trOnk12: all right I trust you on this one :P

trOnk12 (commented):
looks good to me, but maybe someone else would like to take a look ? :)

app/src/main/kotlin/com/wire/android/ui/authentication/create/code/CodeScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.wire.android.ui.authentication.create.code.CodeViewState
+import com.wire.android.ui.common.textfield.CodeFieldValue
+import com.wire.android.ui.common.textfield.CodeTextField
+import com.wire.android.ui.common.textfield.WireTextFieldState
+import com.wire.android.ui.common.topappbar.WireCenterAlignedTopAppBar
+import com.wire.android.ui.theme.wireDimensions
+import com.wire.android.ui.theme.wireTypography
+
+@Composable
+fun CodeScreen(viewModel: CodeViewModel, title: String) {
```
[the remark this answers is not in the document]
saleniuk: I thought the name of the package it was in would be enough, but you're right, I changed the names to be a bit more specific 🚀

app/src/main/kotlin/com/wire/android/ui/authentication/create/email/EmailScreen.kt
```diff
@@ -66,7 +70,9 @@ fun EmailScreen(viewModel: CreatePersonalAccountViewModel, serverConfig: ServerC
 @OptIn(ExperimentalMaterial3Api::class, ExperimentalComposeUiApi::class)
 @Composable
 private fun EmailContent(
-    state: CreatePersonalAccountViewState.Email,
+    state: EmailViewState,
+    title: String,
+    subtitle: String,
```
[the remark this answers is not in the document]
saleniuk: These screens can have different "title" (text shown on the TopAppBar) or "subtitle" depending on the flow it's used in (personal account or team creation) so it's not the inner state but rather the parameter coming from outside the screen that doesn't change during the life of the specific screen

MohamadJaara (approved):
lgtm  🚀

PR #424, opened 2022-03-08
feat: integrate search query (part 1)

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Fixed the review comments from https://github.com/wireapp/wire-android-reloaded/pull/410

- Refactored TopBar with SearchBar functionality:

- to include the topbar with searchbar into any @Composable we had to remember about all the paddings and add them manually, now thanks to ConstraintLayout and hoisiting the SearchBarState the UI gets recomposed correctly so that we do not have to worry about remembering the paddings, instead we can use AppTopBarWithSearchBarLayout where we pass     content: @Composable () -> Unit, which will be wrapped inside the AppTopBarWithSearchBarLayout and the ConstraintLayout will be used to layout it correctly.

- I also removed the need of calculating the height of the SearchBar to apply the offset for the SearchBar so that the collapsing effect is achieved. Instead I used Column inside private fun AppTopBarWithSearchBarContent(...) where simply the SearchBar gets aligned under the TopBar, that way we can use this component in any densitity screens and do not have to worry on perhaps wrong calculation of the height. 

- I also moved all tthe responsibility of UI "state" transistion for example. when the searchbar collapses or the topbar dissapears into the AppTopBarWithSearchBarLayout, so that we do not have to worry about applying it in our Composables where we will use the AppTopBarWithSearchBarLayout.

NOTE: currently there is no correct implemention of the search within Conversation screen which is part of the HomeScreen  

### Dependencies

Needs releases with:

- [ ] #410

### Attachments (Optional)
search

----

----

----
##### References


vitorhugods:
I assume some of the changes here are already present in #410, right? So the changelog will be smaller once it gets merged.

In this case, it would be good to list the PR #410 as a "dependency" in the description of this PR. So we can track it down when this is ready for review/merge.

I'll go ahead and change the PR description for now. But feel free to correct me if I got this wrong.

app/src/main/kotlin/com/wire/android/ui/common/SearchBar.kt
```diff
@@ -75,16 +95,26 @@ fun SearchBarUI(placeholderText: String, modifier: Modifier = Modifier, onTextTy
                 }
             }
         },
-        textStyle = LocalTextStyle.current.copy(fontSize = 14.sp, textAlign = TextAlign.Start),
-        placeholderTextStyle = LocalTextStyle.current.copy(fontSize = 14.sp, textAlign = TextAlign.Center),
+        interactionSource = interactionSource,
+        textStyle = textStyle.copy(fontSize = 14.sp),
```
yamilmedina, line 99: I don't remember if we have already some `WireDimensions` for these values, if we have one would be a good idea to use it

app/src/main/kotlin/com/wire/android/ui/common/SearchBar.kt
```diff
@@ -75,16 +95,26 @@ fun SearchBarUI(placeholderText: String, modifier: Modifier = Modifier, onTextTy
...
         placeholderText = placeholderText,
         maxLines = 1,
         singleLine = true,
     )
 }
 
+@Composable
+fun rememberSearchInputState(): SearchInputState = remember { SearchInputState() }
+
+class SearchInputState {
```
yamilmedina, line 110: Do you think it's possible to have this class marked as private? I think, I didn't see any other usages

app/src/main/kotlin/com/wire/android/ui/common/topappbar/AppTopBarWithSearchBarLayout.kt
```diff
@@ -0,0 +1,276 @@
...
+    onSearchQueryChanged: (String) -> Unit,
+    onInputClicked: () -> Unit,
+    onCloseSearchClicked: () -> Unit,
+    appTopBar: @Composable () -> Unit,
+) {
+    Surface(
+        modifier = Modifier
+            .fillMaxWidth()
+            .wrapContentHeight(),
+        shadowElevation = if (isTopBarVisible) 8.dp else 0.dp
```
yamilmedina, line 158: I think We could use values from `WireDimensions` if possible =)

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/Extension.kt
```diff
@@ -6,7 +6,6 @@ import androidx.compose.runtime.Composable
 import com.wire.android.ui.main.conversationlist.common.FolderHeader
 import io.github.esentsov.PackagePrivate
```
yamilmedina, line 7: We could remove this unused import 👀

vitorhugods:
This is very hard to review the way it is because I don't know what's part of #410 and what isn't.


That is:
#410 is not mergeable without the fixes in this PR.
This PR is hard to review without merging #410.

Circular dependency on PRs.


For the future, I suggest:

### 1. Create a branch from develop that will contain all parts, something like `feat/search_users`

### 2. Create part branches from it and open PRs against the branch created before:

Somethign like `feat/search_users/part1`, subtmit a PR:
* `feat/search_users` ← `feat/search_users/part1`

Repeat as needed:
* `feat/search_users` ← `feat/search_users/part1`
* `feat/search_users` ← `feat/search_users/part2`
* `feat/search_users` ← `feat/search_users/part3`
* `feat/search_users` ← `feat/search_users/part4`


### 3. After all parts are merged, open a final PR:
* `develop` ← `feat/search_users`
  * This will contain all changes at once, but the code has been all reviewd.

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
```diff
@@ -103,20 +155,53 @@ class HomeState(
...
     bottomSheetContent: @Composable (ColumnScope.() -> Unit)? = null
 ): HomeState {
-    return remember {
+    val navBackStackEntry by navController.currentBackStackEntryAsState()
+
+    val navigationItem = when (navBackStackEntry?.destination?.route) {
+        HomeNavigationItem.Archive.route -> HomeNavigationItem.Archive
+        HomeNavigationItem.Vault.route -> HomeNavigationItem.Vault
+        else -> HomeNavigationItem.Conversations
+    }
```
vitorhugods, line 191: > Can we just use `HomeNavigationItem.values().first { it.route == navBackStackEntry?.destination?.route }`? 🤔 
> or at least implement all branches directly instead of using `else` -  this way we won't forget about this when for instance we add another item to this enum

https://github.com/wireapp/wire-android-reloaded/pull/410/files#r819372276

app/src/main/kotlin/com/wire/android/ui/home/newconversation/contacts/ContactsViewModel.kt
```diff
@@ -0,0 +1,63 @@
...
+                    "test3"
+                ),
+                Contact(
+                    "test1"
+                ), Contact(
+                    "test2"
+                ), Contact(
+                    "This is last contact"
+                )
+            )
```
vitorhugods, line 53: > we could do it like this
> 
> ```
> buildList<Contact> {
>         Contact("This is first contact")
>         (1..13).map {  Contact("test $it") }
>         Contact("This is last contact")
> }
> ```

https://github.com/wireapp/wire-android-reloaded/pull/410/files#r820912381

I think the correct calls are `add(item)` and `addAll(items)`:

```kotlin
buildList<Contact> {
        add(Contact("This is first contact"))
        addAll((1..13).map {  Contact("test $it") })
        add(Contact("This is last contact"))
}
```

vitorhugods (changes requested):
Some of the changes and comments from #410 were not addressed.

PR #427, opened 2022-03-09
fix: refactor navigation on User Profile Screen and fix avatar upload bugs [AR-1204][AR-1210][AR-1211]

gongracr:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1204" title="AR-1204" target="_blank">SubtaskAR-1204</a>  Refactor navigation between UserProfile and Avatar Picker screens
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Currently the navigation between the `UserProfileScreen` and the `AvatarPickerScreen` was made via the `UserProfileViewModel` instead of using its own View Model.
- The way we had previously to share the picked image on the gallery/camera was also broken, hence not allowing to update the Avatar Icon on the `UserProfileScreen`. Also, the Camera option, was taking pictures on preview mode, hence pixelating the image too much.
- The User Profile didn't have any persistence load/save mechanisms, making the whole process of picking a new image, only available during the ongoing session.

### Solutions

- Added a new `AvatarPickerViewModel` to deal with the avatar picker logic.
- Added a new file provider to allow the Android Camera save the taken picture on a specific internal path.
- Delegated the process of uploading the picture to the `AvatarPickerViewModel`.
- Thanks to this [PR](https://github.com/wireapp/kalium/pull/243) I could refactored how the avatar image is retrieved and loaded (via the `GetPublicAssetUseCase`) and persisted to memory once a new picture has been chosen/taken (via the `UploadUserAvatarUseCase`) on the corresponding View Models.
- Added [Coil library](https://coil-kt.github.io/coil/) to handle image loading on the app.

### Attachments (Optional)
https://user-images.githubusercontent.com/2468164/157436359-c72e4228-969a-489c-a6ec-b7d58932d3ea.mov

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPickerViewModel.kt
```diff
@@ -0,0 +1,62 @@
...
+            (getUserAvatar(avatarAssetId) as PublicAssetResult.Success).asset
+        } catch (e: Exception) {
+            null
+        }
+
+    fun uploadNewPickedAvatarAndBack(chosenImgUri: Uri, context: Context) {
+        viewModelScope.launch {
+            withContext(Dispatchers.IO) {
+                val data = chosenImgUri.toByteArray(context)
+                uploadUserAvatar(mimeType = "image/jpg", imageData = data)
```
yamilmedina, line 55: I think for now it's ok doing it fixed to "image/jpg", but later on (for other types of assets) we could manage to extract the actual mimetype from the Uri, maybe as an Extension Function on `FileUtil` class. 
Using: 
- https://developer.android.com/reference/android/webkit/MimeTypeMap#getFileExtensionFromUrl(java.lang.String)
- https://developer.android.com/reference/android/webkit/MimeTypeMap#getMimeTypeFromExtension(java.lang.String)

yamilmedina (dismissed):
Awesome work! 👏  a pain in the a$$ working the FileProvider, but now it's working.

Just a small task before merging, taking care of this detekt issues, just a suppress will do it:
https://github.com/wireapp/wire-android-reloaded/runs/5482156311?check_suite_focus=true 

exceptions - 40min debt
	TooGenericExceptionCaught - [e] at /home/runner/work/wire-android-reloaded/wire-android-reloaded/app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt:95:18
	TooGenericExceptionCaught - [e] at /home/runner/work/wire-android-reloaded/wire-android-reloaded/app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPickerViewModel.kt:49:18
style - 10min debt
	MagicNumber - [toByteArray] at /home/runner/work/wire-android-reloaded/wire-android-reloaded/app/src/main/kotlin/com/wire/android/util/FileUtil.kt:33:84

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -44,20 +44,18 @@ import io.github.esentsov.PackagePrivate
 enum class NavigationItem(
     @PackagePrivate
     internal val primaryRoute: String,
-    private val canonicalRoute: String,
+    private val canonicalRoute: String = primaryRoute,
     val arguments: List<NamedNavArgument> = emptyList(),
```
saleniuk, line 48: I think it can be removed because it isn't used anywhere now as we declare arguments directly in `getRouteWithArgs` and we don't have any non-string arguments - what do you think?

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -31,32 +29,69 @@ import kotlinx.coroutines.withContext
...
+        showLoadingAvatar(true)
+
+        viewModelScope.launch {
+            getSelf().collect { selfUser ->
+                with(selfUser) {
+                    // Load user avatar raw image data
+                    val avatarAssetRaw = completePicture?.let { getAvatarAsByteArrayOrNull(it) }
+
+                    // Update avatar asset id on user data store
+                    completePicture?.apply { dataStore.updateUserAvatarAssetId(this) }
```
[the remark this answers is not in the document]
gongracr: The idea was simply to cache the asset id on the data store and that way save one more call to the `getSelf` api and only make a call to the `GetPublicAssetUseCase` which will return it from memory in case it had been also cached previously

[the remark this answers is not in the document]
saleniuk: I just feel like maybe we should rather cache it on the repository level, it would be easier to use - what do you think?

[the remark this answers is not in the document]
gongracr: But this is already happening with `downloadPublicAsset` method, isn't it? But we still need to keep somewhere a reference/id to the specific asset we cached previously and that we want to retrieve again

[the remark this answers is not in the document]
gongracr: Or you meant, also adding methods to save/load the avatar asset id, or directly the self user avatar?

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -165,8 +174,7 @@ class UserProfileViewModel @Inject constructor(
         }
     }
 
-    // !! TODO: this method is made only to pass the mock bitmap, later on we will not need it !!
-    fun mockMethodForAvatar(bitmap: Bitmap) {
-        userProfileState = userProfileState.copy(avatarBitmap = bitmap)
+    companion object {
+        private const val LOAD_AVATAR_TIMEOUT_MS = 5000L
```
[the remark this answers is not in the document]
gongracr: For now I'd prefer to keep it temporarily in this file until we get a common view on how to handle core/network failures in a unified way.  Also, moving it to `integers.xml` would require to pass a Context instance to the viewmodel method, which I'm not really fond of. Would that still be ok?

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
```diff
@@ -77,14 +44,19 @@ fun UserProfileAvatar(
...
+
         Image(
-            bitmap = avatarBitmap.asImageBitmap(),
+            painter = rememberAsyncImagePainter(model = avatarResource),
             contentDescription = stringResource(R.string.content_description_user_avatar),
-            contentScale = ContentScale.Crop,
             modifier = Modifier
                 .padding(dimensions().userAvatarStatusBorderSize)
-                .clip(CircleShape)
+                .background(Color.Companion.Black, CircleShape)
```
trOnk12, line 56: could you at least change it to Color.Black :P you do not need calling Companion there

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
```diff
@@ -42,9 +42,11 @@ fun HomeTopBar(
     val scrollDownState = viewModel.scrollDownFlow.collectAsState(false)
     val firstLineElevation = if (!isSearchable || scrollDownState.value) dimensions().topBarElevationHeight else 0.dp
 
-    Box(
-        Modifier.background(Color.Transparent)
-    ) {
+    LaunchedEffect(viewModel) {
```
trOnk12, line 45: hmmm... what did you try to achieve with it using viewModel as a key ?

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
```diff
@@ -42,9 +42,11 @@ fun HomeTopBar(
...
+        viewModel.loadUserAvatar()
+    }
+
+    Box {
```
trOnk12, line 49: I changed a lot this HomeTopBar function in my PR so I guess I am going to refactor it anyway, but I think it would make more sense to just pass bytearray here and hoist loading it up to the parent

app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
```diff
@@ -36,6 +48,14 @@ class HomeViewModel
 
     suspend fun navigateToUserProfile() = navigateTo(NavigationItem.UserProfile, MY_USER_PROFILE_SUBROUTE)
 
+    fun loadUserAvatar() {
+        viewModelScope.launch {
+            dataStore.avatarAssetId.first()?.let {
+                userAvatar = (getPublicAssetUseCase(it) as PublicAssetResult.Success).asset
+            }
+        }
+    }
```
trOnk12, line 57: are we handeling here errors ? you just cast it to  PublicAssetResult.Success, which is abit dangerous I think here ;p

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -31,32 +29,69 @@ import kotlinx.coroutines.withContext
...
+
+    private suspend fun fetchSelfUser() {
+        // Show progress bar on initial load
+        showLoadingAvatar(true)
+
+        viewModelScope.launch {
+            getSelf().collect { selfUser ->
+                with(selfUser) {
+                    // Load user avatar raw image data
+                    val avatarAssetRaw = completePicture?.let { getAvatarAsByteArrayOrNull(it) }
```
trOnk12, line 53: we are not handeling any failures here

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -31,32 +29,69 @@ import kotlinx.coroutines.withContext
...
+                        status = UserStatus.AVAILABLE,
+                        fullName = name.orEmpty(),
+                        userName = handle.orEmpty(),
+                        teamName = team,
+
+                        // TODO: remove mocked team when other accounts functionality is ready
+                        otherAccounts = listOf(
+                            OtherAccount("someId", "", name.orEmpty(), "Wire Swiss GmbH"),
+                            OtherAccount("someId", "", "B. A. Baracus", "The A-Team"),
+                        )
```
trOnk12, line 73: In my opinion a better user experience would be if we load the avatar seperetly as it may take more time because we are calling additional usecase for that, we could display user info immediatly though, what do you think ?

app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPicker.kt
```diff
@@ -9,56 +9,59 @@ import androidx.compose.foundation.layout.padding
...
         state = state,
         onCloseClick = {
-            when (val avatarImageState = state.picturePickerFlow.bitmapState) {
-                is BitmapState.BitmapPicked -> viewModel.changeUserAvatar(avatarImageState.bitmap, shouldNavigateBack = true)
-                is BitmapState.InitialBitmap -> viewModel.navigateBack()
+            if (state.avatarPickerFlow.pictureState is PictureState.Picked) {
+                viewModel.uploadNewPickedAvatarAndBack(state.avatarPickerFlow.pictureState.avatarUri, context)
+            } else {
+                viewModel.navigateBack()
             }
```
trOnk12, line 56: I think we talked about this on the call, maybe I did not explain it correctly to you but here on the designs : 

image

the loading of the image is happening actually on the UserProfileScreen. I also discussed this approach with Vitor I think you took and the issues here are : 

if loading takes long time, it looks like screen freezes, we could add a spinner on AvatarPicker to indicate loading, but then lets say either user has no internet or it takes like 1 minutes, we have no option to go back to the previous screeen while loading as the back button is actually starting the upload.

trOnk12 (commented):
Generally, good job 🥇left some comments and questions 

also the picture on the profile screen looks really weird : 

image

app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
gongracr: Good catch. Adding a dummy try/catch clause for now. In any case this smart casts will be soon refactored once we decide on a better wrapping for these use cases results

app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPicker.kt
gongracr: I know that the designs indicate that we should do the upload on the user profile screen, but actually from the technical point of view, it would make more sense to do the api call to upload the avatar on the picker screen. 

In any case, my idea was to just merge these changes now and fix the picker screen as it is not functional atm. Then once Wolfgang is back, we will need to have a talk with him, and decide on what's the best approach on where to upload it, and error handling, since the current design also doesn't allow the user to reset the picture to the original one once she/he has picked a new one. What do you think?

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
gongracr: If the avatar asset raw is null, then the default fallback picture will be shown

trOnk12: that is true, but we are calling : getPublicAssetUseCase(avatarAssetId) inside getAvatarAsByteArrayOrNull() so there could some failure from the use case and we are not handeling but we are just displaying the default picture and not show any snackbar or anything, is it something we want to do ?

app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPicker.kt
trOnk12: sure, I am fine with it, just was thinking you did notice it

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
gongracr: That is indeed a good point. However I think 99% of the cases it will take more or less the same time to load both, as avatar is meant to change rarely and therefore the subsequent calls to this use case will be rather immediate as the avatar will be pulled directly from the assets database. I could still change the logic a bit, to load the asset on a separate coroutine, if you think it's indeed necessary and the current avatar loading times are too high though.

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
gongracr: The error here will be when downloading the avatar picture, not when trying to upload it as the design specs suggest, so not sure what should I show here, maybe some snack bar saying "There was an error while fetching the user info, check your internet connection" ? But it is also true that I didn't add the snackbar when the uploads fails so will do it right away.

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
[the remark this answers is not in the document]
saleniuk: Yeah, can't we also cache this asset id in the repository? 😄

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
trOnk12: oki doki, thanks for explanation, your call on that one :P I am fine with both

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
trOnk12: I do not think it can hurt and may improve a user experience abit, but your call if you think it is worth it and will not take much time

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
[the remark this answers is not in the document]
gongracr: In any case I will mark it with a TODO: to adapt to this once we revisit this for the File sharing feature, where we will most likely create another db table to store the assets ids

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
gongracr: Added some inline comment to clarify it

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
gongracr: Good catch. Added

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -31,28 +29,69 @@ import kotlinx.coroutines.withContext
...
+                // Update avatar asset id on user data store
+                // TODO: obtain the asset id through a useCase once we also store assets ids
+                dataStore.updateUserAvatarAssetId(avatarAssetId)
+            } catch (e: ClassCastException) {
+                // Show error snackbar if avatar download fails
+                showErrorMessage()
+            } finally {
+                showLoadingAvatar(false)
             }
         }
```
trOnk12, line 96: I just now remember that while I was working on it the use cases do not uses the DIspatcher.IO so I would suggest to use it here instead so that we do not freeze the UI

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -31,28 +29,69 @@ import kotlinx.coroutines.withContext
...
+    }
+
+    private fun showLoadingAvatar(show: Boolean) {
+        userProfileState = userProfileState.copy(isAvatarLoading = show)
+    }
+
+    private fun updateUserAvatar(avatarAssetId: UserAssetId) {
+        // We try to download the user avatar on a separate thread so that we don't block the display of the user's info
+        viewModelScope.launch {
+            try {
```
trOnk12, line 81: I woudl suggest to add here withContext(Dispatchers.IO)

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -165,8 +178,7 @@ class UserProfileViewModel @Inject constructor(
         }
     }
 
-    // !! TODO: this method is made only to pass the mock bitmap, later on we will not need it !!
-    fun mockMethodForAvatar(bitmap: Bitmap) {
-        userProfileState = userProfileState.copy(avatarBitmap = bitmap)
+    companion object {
+        const val GENERIC_DOWNLOAD_USER_INFO_ERROR = 10
```
trOnk12, line 182: how about passing Int value that do not have much meaning, we pass a sealed class or an enum ?

trOnk12 (approved):
small comments but looks good to me ! goob job 👍 🥇

trOnk12 (approved):
well done 🍰

trOnk12 (approved):
even better 💪

PR #428, opened 2022-03-09
chore: initialize kalium logger as debug mode

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Initialize Kalium Logger in DEBUG mode due to being easier to identify current or new issues that may rise in staging build.

PS: current initialization mode for logger is `disabled` as this will probably be held logic by a feature flag or something we can control from outside the code.

----
##### References


app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -38,7 +40,18 @@ class CoreLogicModule {
...
-        return CoreLogic(applicationContext = context, rootProteusDirectoryPath = proteusPath, clientLabel = deviceLabel)
+
+        val coreLogic = CoreLogic(
+            applicationContext = context,
+            rootProteusDirectoryPath = proteusPath,
+            clientLabel = deviceLabel
+        )
+
+        CoreLogger.setLoggingLevel(
+            level = KaliumLogLevel.DEBUG
```
saleniuk, line 51: Is it a temporary solution? I think that we shouldn't allow debug logs in release version, so maybe we should check if it's `BuildConfig.DEBUG`?

trOnk12: personally I do no think it is a right place to call  CoreLogger.setLoggingLevel(...) this method should simply provide a CoreLogic component, with this it is doing also doing additional stuff, also I am not sure hwo often coreLogicProvider will be called but every time it is being called it is setting a logging level, normally I am used to do this stuff inside an Application, do not know how hard it is to achieve in current architecture ?

alexandreferris: @saleniuk for now we can use `BuildConfig.DEBUG` but in the future, the best approach would be having this controlled/managed by a feature flag or something we can control outside of doing the manual change into code.

I thought on using directly debug as the build is still in staging and it would be easier to identify issues.

alexandreferris: @trOnk12 first it was done in `WireApplication` but as it only handles the worker it wasn't a good approach, so thought on leaving it in the module as it is when it's built.

We can also add it to `WireActivity` as it is the main/first activity that is launched to the user, thus setting it only once the app is open. Let me know what you think

trOnk12: isn't the WireApplication being used as our App launches 🤔  we could just override OnCreate() there and initialize the logger or I am missing something ?

vitorhugods: `WireApplication` is overriding the `WorkerFactory` thing. And putting the logger in that method would be weird, indeed.

But putting it into `onCreate` could work.

I also felt weird when suggesting to set the log inside the Hilt provider.
But initializing it in a different place (Application), where there's little to now interaction with CoreLogic is also weird.

trOnk12: why dou think doing ``` CoreLogger.setLoggingLevel(
            level = KaliumLogLevel.DEBUG ``` inside onCreate of WireApplication would be weird ?

trOnk12: I do not know, it is a bit weird to me to Configure something that has nothing to do with  CoreLogic() on the @Provides annotated method, we are also not injecting it inside the CoreLogic thingy, neither we are using a Context proviede to the coreLogicProvider() maybe I am missing something but normally I initialized stuff like that inside the Application OnCreate() method and it worked, there is also the 

https://developer.android.com/topic/libraries/app-startup

but I did not dive deep into it yet :P

vitorhugods: > why dou think doing `CoreLogger.setLoggingLevel( level = KaliumLogLevel.DEBUG` inside onCreate of WireApplication would be weird ?

Not weird per se, but the logger is definitely related to CoreLogic. So putting it all together would be nice. A single place to configure how CoreLogic works.

But I guess we all see that putting it in Application's `onCreate` is alright and wouldn't be wrong to put it there. So I agree moving there too.

alexandreferris: I have then override the onCreate on `WireApplication.kt` and set the logging level there verifying if the buildConfig is DEBUG for now.

trOnk12 (approved):
lgtm 🍰 , thanks for changing it 🥇

PR #434, opened 2022-03-11
fix: exif rotation image avatar (AR-1212)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1212" title="AR-1212" target="_blank">SubtaskAR-1212</a>  Uploaded profile image is displayed on landscape mode
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1212

- This fixes the issue when taking a picture and gets rotated, now we are using the `Exif` info to get the correct orientation
- Also fixes aspect ratio for displaying the avatar, now using `FillWidth` to preserve aspect ratio
- Included a possible solution to reload the image preview when error occurs 

### Attachments (Optional)

https://user-images.githubusercontent.com/5806454/157950747-366f0a8c-e678-47aa-b3d5-317b97f93362.mp4

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPickerViewModel.kt
```diff
@@ -50,11 +50,9 @@ class AvatarPickerViewModel @Inject constructor(
     fun uploadNewPickedAvatarAndBack(imgUri: Uri, context: Context) {
         viewModelScope.launch {
             withContext(Dispatchers.IO) {
+                postProcessCapturedAvatar(imgUri, context)
                 val data = imgUri.toByteArray(context)
-                val extension = MimeTypeMap.getFileExtensionFromUrl(imgUri.path)
-                val mimeType = context.contentResolver.getType(imgUri)
-                    ?: MimeTypeMap.getSingleton().getMimeTypeFromExtension(extension)
-                    ?: "image/$extension"
+                val mimeType = imgUri.getMimeType(context) ?: "image/jpg"
```
saleniuk, line 55: I believe it should be `image/jpeg`: https://www.w3.org/Graphics/JPEG/
`image/jpg` should also work in general, but if we can I think we should stick to the first one

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+ * @param context
+ */
+fun postProcessCapturedAvatar(uri: Uri, context: Context) {
+    try {
+        val rawImage = uri.toByteArray(context)
+        val avatarBitmap = rawImage.toBitmap() ?: return
+
+        // Rotate if needed
+        val exifInterface = context.contentResolver.openInputStream(uri).use { inputStream ->
+            return@use inputStream?.let { ExifInterface(it) }
```
gongracr, line 53: Just out of curiosity, why do we need to add the `return@use` statement here? wouldn't it suffice just with the `.let {}` clause as it will return the object between brackets?

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -107,6 +108,7 @@ object Libraries {
     const val composeConstraintLayout   = "androidx.constraintlayout:constraintlayout-compose:${Versions.composeConstraint}"
     const val composeRuntimeLiveData    = "androidx.compose.runtime:runtime-livedata:${Versions.compose}"
     const val dataStore                 = "androidx.datastore:datastore-preferences:${Versions.dataStore}"
+    const val exifInterface                      = "androidx.exifinterface:exifinterface:${Versions.exif}"
```
gongracr, line 111: despite sounding a bit picky/pixel perfect nerdie, can we level it up to the rest of dependencies to make it more obvious at first sight from which library comes? 🤓

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+        // NOOP: None post process op performed
+    }
+}
+
+/**
+ * Compress image to save some disk space and memory
+ */
+private fun compressImage(imageBitmap: Bitmap): ByteArray? {
+    val byteArrayOutputStream = ByteArrayOutputStream()
+    imageBitmap.compress(Bitmap.CompressFormat.JPEG, 90, byteArrayOutputStream)
```
gongracr, line 72: Do we want to compress it even more, say 50/60% and extract this to a variable that can be reused when sharing media files later?

gongracr (changes requested):
Just small comments but overall seems good. However it seems that when selecting a landscape picture from the gallery, the BulletHomeImagePreview doesn't work as expected as the canvas gets cropped by the original image height. Check the picture below
Captura de pantalla 2022-03-14 a las 10 51 47

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    return try {
+        val rotated = Bitmap.createBitmap(this, 0, 0, this.width, this.height, matrix, true)
+        this.recycle()
+        rotated
+    } catch (exception: Exception) {
+        this
+    }
+}
```
trOnk12, line 35: kind of wonder what Normal orientation is here, without experience with ExiInterface :P maybe we could include the meaning of "normal orientation" info inside the method name ?

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+import androidx.exifinterface.media.ExifInterface
+import java.io.ByteArrayOutputStream
+
+/**
+ * Rotates the image to a [ExifInterface.ORIENTATION_NORMAL] in case it's rotated with a different orientation
+ *
+ * @param exif Exif interface for of the image to rotate
+ * @return Bitmap the rotated bitmap or the same in case there is no rotation performed
+ */
+private fun Bitmap.rotateImageToNormalOrientation(exif: ExifInterface?): Bitmap {
```
trOnk12, line 17: what happens when exif is null ?

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+    val orientation = exif?.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
+    val matrix = Matrix()
+    when (orientation) {
+        ExifInterface.ORIENTATION_NORMAL -> return this
```
trOnk12, line 21: do we need to do this : 

 ExifInterface.ORIENTATION_NORMAL -> return this
 
if after everything we do "else" on the value and return this any way ? or has it to do with a nullable oreiantion value ? maybe more clear will be a null check if(orienation == null) and otherwise use a when ? so that is it more implicit what is going on here

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+        ExifInterface.ORIENTATION_ROTATE_180 -> matrix.postRotate(180f)
+        ExifInterface.ORIENTATION_ROTATE_90 -> matrix.postRotate(90f)
+        ExifInterface.ORIENTATION_ROTATE_270 -> matrix.postRotate(270f)
+        else -> return this
+    }
+
+    return try {
+        val rotated = Bitmap.createBitmap(this, 0, 0, this.width, this.height, matrix, true)
```
trOnk12, line 29: maybe add named parameters to createBitmap ? no really know what is this, 0,0 , matrix, true

trOnk12 (commented):
It is probably due to my inexperience with Exif files, but would it be perhaps possible to somewhere include the info what is Exif, what is ExifInterface ? perhaps it is an overkill and the knowledge about Exif should be expected when working with those classes :P not sure, just my impression when reading

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
```diff
@@ -56,7 +56,7 @@ fun UserProfileAvatar(
                 .background(Color.Black, CircleShape)
                 .size(size)
                 .clip(CircleShape),
-            contentScale = ContentScale.FillBounds,
+            contentScale = ContentScale.FillWidth,
```
gongracr, line 59: According to the latest design requirements, this will need to be changed programmatically depending on the image to upload exif information:
_Depending on whether the image is in portrait or landscape mode, given by its Exif information, we should apply a content scale filter of FillWidth or FillHeight_

app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPickerViewModel.kt
yamilmedina: Thanks, totally agree, we should stick to the standard !

app/src/main/kotlin/com/wire/android/ui/common/imagepreview/BulletHoleImagePreview.kt
```diff
@@ -41,13 +41,14 @@ fun BulletHoleImagePreview(imageUri: Uri, contentDescription: String) {
...
-                contentScale = ContentScale.Crop,
+            AsyncImage(
+                model = imageUri,
+                contentScale = ContentScale.FillWidth,
                 contentDescription = contentDescription,
-                modifier = Modifier.fillMaxSize()
+                modifier = Modifier.fillMaxSize(),
+                onSuccess = { it.result.drawable.apply { this.invalidateSelf() } },
+                onError = { it.result.drawable?.invalidateSelf() }, // handle error when not possible to load preview
+                error = rememberAsyncImagePainter(model = imageUri) // retry, fallback to load image when error
```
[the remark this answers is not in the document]
yamilmedina: Yes, indeed 👀 . I was having some attempts to reload the image on failure, since the coil component wasn't behaving well loading, sometimes a blank screen.

Anyway, I removed the `onError` and `onSuccess` blocks and maintained the `error` with a local fallback. We'll have to handle this in a future PR. 

Now the preview loader it's working better with no errors, using a Bitmap instead of a ByteArray 😖

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
yamilmedina: Yeah, I had to research myself a couple of things regarding this `ExifInterface` (Image metadata) 🔬  
The normal orientation means the way we took the picture, for example, so if we took a picture in portrait, should not be rotated by the camera software. This will get that info accordingly to the metadata.

I added a Kdoc to the function, but didn't manage to change the naming since it will depend, Normal means either Portrait or Landscape.

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
yamilmedina: It will do nothing, will return the same bitmap without transformation. I considered this process shouldn't be breaking for the user, but maybe We can get some definitions on what to do if post-processing fails and implement that in a future PR :D

app/src/main/kotlin/com/wire/android/util/ImageUtil.kt
yamilmedina: I would like to do it, but We can't do that since this is a Java class and no named parameters are allowed 🙈 
There are just (x, y) coordinates to start creating the first pixel for the bitmap

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
yamilmedina: I spend a couple of hours trying to implement this dynamic scale filter, but had to implement another solution using `ContentScale.Crop` (This will Scale the source uniformly (maintaining the source's aspect ratio)). 

Having the Exif data loaded it's a bit overkill and personally have concerns about the performance, since we have to open `InputStream`s, or open Files to extract this metadata. I was thinking on how this will affect performance, for instance in a group conversation loading every avatar, having to calculate the Exif data for every user.

Using `ContentScale.Crop` solves the issue for both Landscape and Portrait, achieving images to maintain aspect ratio 🚀  and we are doing the Exif compute just once, only when we save/upload the avatar

gongracr (approved):
Looking great after addressing the comments, GJ  🥁 ✨

PR #438, opened 2022-03-16
feat: add msg sharing options ui (AR-1086)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1086" title="AR-1086" target="_blank">SubtaskAR-1086</a>  Implement attachment file type options menu component
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
https://wearezeta.atlassian.net/browse/AR-1086
This PR includes the UI part for sharing attachment options in the message composer
- Created sharing component menu to reuse and create all the options
- Extracted to a new file, so we can isolate the component and use a dedicated ViewModel for its logic later on.

### Attachments (Optional)

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/common/AttachmentButton.kt
```diff
@@ -0,0 +1,78 @@
...
+            modifier = Modifier
+                .size(dimensions().attachmentButtonSize)
+                .background(MaterialTheme.wireColorScheme.primaryButtonEnabled, CircleShape)
+                .padding(dimensions().userAvatarStatusBorderSize)
+                .clickable { onClick() }
+        ) {
+            Image(
+                painter = painterResource(icon),
+                contentDescription = stringResource(R.string.content_description_user_avatar),
+                contentScale = ContentScale.FillBounds,
```
saleniuk, line 53: I'm not sure if we want `FillBounds` - if the image has different aspect ratio then it will be distorted

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -0,0 +1,78 @@
...
+
+@OptIn(
+    ExperimentalComposeUiApi::class,
+    ExperimentalFoundationApi::class
+)
+@Composable
+fun AttachmentOptionsComponent() {
+    val viewModel: AttachmentOptionsViewModel = hiltViewModel()
+    LazyVerticalGrid(
+        cells = GridCells.Fixed(4),
```
saleniuk, line 29: I feel that this value can change depending on device type and screen orientation so maybe it should be moved to `wireDimensions` or even we should create another one `wireIntegers` as it's not actually a dimension 😅

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
```diff
@@ -82,6 +85,7 @@ private fun MessageComposer(
     onSendButtonClicked: () -> Unit
 ) {
     val focusManager = LocalFocusManager.current
+    var additionalInfoDisplayed by remember { mutableStateOf(false) }
```
trOnk12, line 88: to keep it consistent and clean, please try to move this value to MessageComposerInnerState and propagte this event to MessageComposer

trOnk12 (commented):
the collapsing of the share options has no animation which in my opinion looks a bit weird :P if you do it right you get the animation almost for free in Jetpack Compose, so maybe it is worth to take a look at it in this PR ? also a small comment about the remember object that is not inside the inner message compose state

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
trOnk12: I agree on that one :P it is abit wishfull thinking in my opinion that it will looks correct on all screens ;p maybe there is a way to keep placing those buttons on a row untill it has space and then jump into the next line and keep placing those buttons if there are some left ? that way it will be more responsive but then you theoritacally could have a case on some small screen devices for example where you got 2 button on a row but have lets say 3 columns in total :P 

otherwise we could use some custom Layout measuring the screen width and dividing it by lets say 120.dp for each grid and then decide based on that how many rows in a column we want to use :P which kinds of does this I think : 

 cells = GridCells.Adaptive(minSize = 128.dp)

app/src/main/kotlin/com/wire/android/ui/common/AttachmentButton.kt
```diff
@@ -0,0 +1,78 @@
...
+            modifier = Modifier.requiredSizeIn(minWidth = 40.dp, maxWidth = 60.dp)
+        )
+    }
+}
+
+@Preview(showBackground = true)
+@Composable
+fun PreviewAttachmentButton() {
+    AttachmentButton("Attach File", R.drawable.ic_attach_file) {
+        Log.d("AttachmentButton", ">> Clicked")
```
gongracr, line 76: Do we need to add a log on the preview composable? If so, let's use the `appLogger` instead.
```suggestion
        appLogger.d(">> Clicked")
```

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -0,0 +1,78 @@
...
+        contentPadding = PaddingValues(32.dp),
```
gongracr, line 30: Let's also extract this padding to Dimensions

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -0,0 +1,78 @@
...
+                stringResource(R.string.attachment_voice_message), R.drawable.ic_mic_on, modifier = Modifier.padding(top = 32.dp)
+            ) {
+                Log.d("AttachmentButton", "voice message clicked")
+            }
+        }
+        item {
+            AttachmentButton(
+                stringResource(R.string.attachment_share_location), R.drawable.ic_location, modifier = Modifier.padding(top = 32.dp)
+            ) {
+                Log.d("AttachmentButton", "share location clicked")
```
gongracr, line 68: Can we substitute all `Log.d()` calls for `appLogger.d()` ?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -0,0 +1,78 @@
...
+        }
+        item {
+            AttachmentButton(stringResource(R.string.attachment_record_video), R.drawable.ic_video_icon) {
+                Log.d("AttachmentButton", "take video clicked")
+            }
+        }
+        // Second row options
+        item {
+            AttachmentButton(
+                stringResource(R.string.attachment_voice_message), R.drawable.ic_mic_on, modifier = Modifier.padding(top = 32.dp)
```
gongracr, line 59: more magic numbers

app/src/main/kotlin/com/wire/android/ui/common/AttachmentButton.kt
yamilmedina: Applied Inside scale, to be safe

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: Thanks, I ended up removing the loggers and replaced it by `TODO` comments, as we'll have to implement those actions in the near near future

app/src/main/kotlin/com/wire/android/ui/common/AttachmentButton.kt
```diff
@@ -0,0 +1,74 @@
...
+) {
+    Column(
+        modifier = modifier.width(dimensions().attachmentButtonSize),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Box(
+            modifier = Modifier
+                .size(dimensions().attachmentButtonSize)
+                .background(MaterialTheme.wireColorScheme.primaryButtonEnabled, CircleShape)
+                .padding(dimensions().userAvatarStatusBorderSize)
```
borichellow, line 45: `dimensions().spacing2x` looks more logical here, IMO

borichellow (commented):
Also, i didn't find in code: what if attachmentOptions are opened and user clicks system back-button? 
I think it would be nice to close attachmentOptions in such case

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
```diff
@@ -341,13 +368,20 @@ private fun MessageComposeActions() {
 }
 
 @Composable
-private fun AdditionalOptionButton() {
+private fun AdditionalOptionButton(messageComposerInnerState: MessageComposerInnerState) {
```
trOnk12, line 371: in my opinion instead of passing messageComposerInnerState: MessageComposerInnerState as a whole it would make more sense to pass booleans, that way we can see from function declaration directly what it can do .

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.wire.android.ui.common.AttachmentButton
+import com.wire.android.ui.common.dimensions
+
+@OptIn(
+    ExperimentalComposeUiApi::class,
+    ExperimentalFoundationApi::class
+)
+@Composable
+fun AttachmentOptionsComponent() {
+    val viewModel: AttachmentOptionsViewModel = hiltViewModel()
```
trOnk12, line 28: looks like this viewmodel does nothing ?

trOnk12:
looks much better <3 could you attach our sexy animations with this component ?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
yamilmedina: I'll leave it as it is right now, as it's the straight forward way to mutate the inner state from inside the button, without having to pass a lambda to it. If I see a window of time, I'll give a look in the next PR's for this component

yamilmedina:
> looks much better <3 could you attach our sexy animations with this component ?

@trOnk12 thanks <3 I've updated the video in the PR description =)

yamilmedina:
> Also, i didn't find in code: what if attachmentOptions are opened and user clicks system back-button? I think it would be nice to close attachmentOptions in such case

Thanks @borichellow for the comments, indeed would be nice to have that behavior for the whole message composer, We can take a look later on maybe in another PR, since we'll have to apply the same logic to the collapsible button. 
As for now, I think we can have the function as it is right now =)

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposer.kt
trOnk12: Pls add TODO then so we do not forget

yamilmedina: Ended up implementing solution passing the lambda and the boolean =)

gongracr (approved):
LGTM 👏 🍇

PR #441, opened 2022-03-21
feat: message sharing attachments permissions request (AR-1098)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1098" title="AR-1098" target="_blank">SubtaskAR-1098</a>  Implement runtime permissions flow for file reading & writing
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1098

Permissions for all the options displayed as possible attachments, this includes:
- Storage access
- Camera access
- Location access
- Mic access

### Attachments (Optional)

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [X] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [X] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #442, opened 2022-03-22
feat: sharing image from gallery (AR-1097)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-1097" title="AR-1097" target="_blank">SubtaskAR-1097</a>  Implement sharing image from gallery behavior
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1097

This PR depends on #441 and takes care of the gallery and file sharing logic (selecting from the file browser/gallery) and passing the rawdata to the usecase (viewmodel) for this private asset to be uploaded. 
- Handles basic error via snackbar
- Created a class to pass this raw asset into the viewmodel, to later on pass it to the usecase (we still have to create this usecase in kalium)
- Reused some work done in the avatar upload, like transforming Uris to ByteArrays

### Attachments (Optional)

Screen Shot 2022-03-22 at 17 28 36

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -9,27 +9,50 @@ import androidx.compose.foundation.layout.PaddingValues
...
 import com.wire.android.util.permission.rememberRecordAudioRequestFlow
 import com.wire.android.util.permission.rememberTakePictureFlow
 
-@OptIn(ExperimentalComposeUiApi::class, ExperimentalFoundationApi::class)
+@OptIn(ExperimentalComposeUiApi::class, ExperimentalFoundationApi::class, ExperimentalMaterial3Api::class)
 @Composable
-fun AttachmentOptionsComponent() {
-    val attachmentOptions = buildAttachmentOptionItems()
+fun AttachmentOptionsComponent(onSendAttachment: (AttachmentPart?) -> Unit, snackbarHostState: SnackbarHostState) {
+    val viewModel: AttachmentOptionsViewModel = hiltViewModel()
```
trOnk12, line 39: from what I see this viewmodel is handeling the UI logic only, the problem that I see with using a hiltViewModel inside a @Composable is that theoritcally when we would have this function called two times inside a another @Composable, so lets say some screen would use this AttachmentOptionsComponent in two different places, those two instances of this composable will share common state as the hiltViewModel() uses a NavBackStackEntry as a host for the viewmodel, i know that for our case this will probably not happen :P 

another problem I see with this approach is that lets say we show the AttachmentOptionsComponent() we fire prepareAttachment() but then user closes the AttachmentOptionsComponent()  but before closing fires this coroutine , then it will not be cancelled, I do not know if that is something we want ? 
```

   viewModelScope.launch {
            attachmentState = try {
                val attachment =
                    AttachmentPart(attachmentUri.getMimeType(context).orDefault(DEFAULT_FILE_MIME_TYPE), attachmentUri.toByteArray(context))
                AttachmentState.Picked(attachment)
            } catch (e: IOException) {
                AttachmentState.Error
            }
        }
```

if you refactored the code so that is uses a pure kotlin Class like MessageComposer for example and uses a remember { } retuning that pure Kotlin Class, you will have a proper coroutine cancelation and also each call to AttachmentOptionsComponent will create its own state object managing its own state and do not share it with other calls to AttachmentOptionsComponent() I even wonder if it makes sense to create a seperate plain Kotlin Class to manage this UI component , since this is going to be used only inside the MessageComposer ? So I think it even makes sense to move all this logic to MessageComposerInnerState class, let me know what you tihnk of it

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -9,27 +9,50 @@ import androidx.compose.foundation.layout.PaddingValues
...
 import com.wire.android.util.permission.rememberOpenGalleryFlow
 import com.wire.android.util.permission.rememberRecordAudioRequestFlow
 import com.wire.android.util.permission.rememberTakePictureFlow
 
-@OptIn(ExperimentalComposeUiApi::class, ExperimentalFoundationApi::class)
+@OptIn(ExperimentalComposeUiApi::class, ExperimentalFoundationApi::class, ExperimentalMaterial3Api::class)
 @Composable
-fun AttachmentOptionsComponent() {
-    val attachmentOptions = buildAttachmentOptionItems()
+fun AttachmentOptionsComponent(onSendAttachment: (AttachmentPart?) -> Unit, snackbarHostState: SnackbarHostState) {
```
trOnk12, line 38: since we are not handeling the error inside the AttachmentOptionsComponent() I think it makes moer sense to hoist the error event up to the parent, then requiring SnackbarHostState() it gives us more flexibility to decide how we want to handle the error on the parent component and does not require the parent to use SnackbarHostState() because maybe it doesn't have to and maybe we decide later to handle the error differently

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewState.kt
```diff
@@ -1,10 +1,36 @@
...
     val messages: List<Message> = emptyList(),
-    val messageText : TextFieldValue = TextFieldValue("")
+    val messageText: TextFieldValue = TextFieldValue("")
 )
+
+/**
+ * Represents an attachment part of a message to be sent
+ */
+@Parcelize
+data class AttachmentPart(val mimeType: String, val rawContent: ByteArray) : Parcelable {
```
trOnk12, line 18: I am abit confused about the name of this class :P AttachmentPart makes me think it is a part of something but part of what :P maybe just Attachment would be less confusing ? Dunno just a personal opinion I guess ;p

trOnk12 (commented):
generally great job <3 I just have some doubts about using the viewmodel and passing a snackbarhoststate

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewState.kt
gongracr: How about renaming it to `AttachmentParcel` or `AttachmentBundle` ?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -9,27 +9,50 @@ import androidx.compose.foundation.layout.PaddingValues
...
+        is AttachmentState.Picked -> onSendAttachment(state.attachmentPart)
+        is AttachmentState.Error -> {
+            // FIXME. later on expand to other possible errors
+            val errorMessage = stringResource(R.string.error_unknown_message)
+            LaunchedEffect(state) {
+                snackbarHostState.showSnackbar(errorMessage)
+                viewModel.resetViewState()
+            }
+        }
+    }
```
gongracr, line 54: Can we extract this to a separated function and call it inside the `init { }` block?

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewState.kt
```diff
@@ -1,10 +1,36 @@
...
     val conversationName: String = "",
     val messages: List<Message> = emptyList(),
-    val messageText : TextFieldValue = TextFieldValue("")
+    val messageText: TextFieldValue = TextFieldValue("")
 )
+
+/**
+ * Represents an attachment part of a message to be sent
+ */
+@Parcelize
```
gongracr, line 17: Why do we need to declare it as `@Parcelize` ?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -48,22 +71,22 @@ fun AttachmentOptionsComponent() {
...
-    },
-    onPermissionDenied = { /* TODO: Implement denied permission rationale */ }
-)
+private fun GalleryFlow(viewModel: AttachmentOptionsViewModel): UseStorageRequestFlow {
+    val context = LocalContext.current
+    return rememberOpenGalleryFlow(
+        onGalleryItemPicked = { pickedPictureUri -> viewModel.prepareAttachment(context, pickedPictureUri) },
+        onPermissionDenied = { /* TODO: Implement denied permission rationale */ }
+    )
+}
```
gongracr, line 89: I'm not a huge fan of passing any `context` instance to view models as they are heavy objects, and viewmodels lifecycle may be too long sometimes. Maybe we could just do all the transformations from Uri to ByteArray within the bounds of the composable item (or even better extract it to other utils method) and just pass the attachment raw data to the view model?

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewState.kt
yamilmedina: I don't think it's needed per se, but as in this class we are going to be handling probably big `ByteArray`, using the `Parcelable` implementation theoretically will have a better performance than plain serialization

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: Wow! long but insightful read, thanks 👏  I just want to point out that this ViewModel itself handle the state of the view, but the main idea behind the creation was using a ViewModelScope coroutine to do the "heavy" IO work, cause transforming a `Uri` to a `ByteArray` fits better outside UI context.

Having said that, I agree with the point that actually this will fit better inside the `MessageComposerInnerState` class, cause in some degree this it's also part of composing a message :D 

I will try that approach, fitting this Attachment States inside `MessageComposerInnerState` and maybe just mark the `prepareAttachment` func as suspend to Launch it from the Composable component.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: I'm refactoring this states and moving away from the viewmodel approach, so I think this will be outdated code and it'll change a lot

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: Totally agree! Just took a shortcut seeing other viewmodel using android components, but we have to start moving away from this approach to start testing as you mentioned :D 

Anyway I'm refactoring this and moving away from the viewmodel approach, so I think this will be outdated code

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: @trOnk12 already implemented the suggested solution I think now looks better in fact without the ViewModel 🚀

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: This survived the refactor, but I only extracted the logic to a separate function, because t it's not possible to move it to an `init block` since this is not a class, just a file with composables

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
```diff
@@ -9,27 +9,39 @@ import androidx.compose.foundation.layout.PaddingValues
...
+@OptIn(ExperimentalComposeUiApi::class, ExperimentalFoundationApi::class, ExperimentalMaterial3Api::class)
 @Composable
-fun AttachmentOptionsComponent() {
-    val attachmentOptions = buildAttachmentOptionItems()
+fun AttachmentOptionsComponent(
+    attachmentInnerState: AttachmentInnerState,
+    onSendAttachment: (AttachmentBundle?) -> Unit,
+    onError: (String) -> Unit
+) {
+    val attachmentOptions = buildAttachmentOptionItems(attachmentInnerState)
```
trOnk12, line 42: what if we move this line inside LazyVerticalGrid ? on the place where we are iterating over attachmentOptions as we are not using val attachmentOptions anywhere I do not think we evern need this local variable

trOnk12, line 42: how about instead of passing attachmentInnerState to buildAttachmentOptionItems we pass an lambda like : filePicked : (Uri) -> {} and call buildAttachmentOptionItems like buildAttachmentOptionItems{ onFilePicked = 
uri -> attachmentInnerState.pickAttachment(context, pickedPictureUri }   ? that way it is clear while you are reading what you want to do with the the picked file,  you do not have to jump to buildAttachmentOptionItems first and then to FileBrowserFlow to see what we are going to do with that  attachmentInnerState, let me know what you think

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInnerState.kt
```diff
@@ -1,28 +1,39 @@
...
 
 @Composable
 fun rememberMessageComposerInnerState(
     defaultMessageText: String = "",
-    defaultMessageComposeInputState: MessageComposeInputState = MessageComposeInputState.Enabled
+    defaultMessageComposeInputState: MessageComposeInputState = MessageComposeInputState.Enabled,
+    defaultAttachmentInnerState: AttachmentInnerState = AttachmentInnerState()
 ): MessageComposerInnerState {
 
     return remember {
```
trOnk12, line 24: I think instead of passing each time context to fun pickAttachment(context: Context, attachmentUri: Uri)  you can move this line defaultAttachmentInnerState: AttachmentInnerState = AttachmentInnerState() away from the rememberMessageComposerInnerState argument and do it inside the body like this 

{
val context = LocalContext.current
val attachmentInnerState = AttachmentInnerState(context)
}

So tthat AttachmentInnerState holds the context, but I am not sure how safe it is to do that way , if the context gets garbaged collected in the mean time and we could have a memory leak here, I do not think honestly as I think there is only single instance of context as this is application context

trOnk12 (commented):
impressive work 👏I posted some small comments

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/attachment/AttachmentOptions.kt
yamilmedina: I'll leave it as it is, because sadly I think it's not possible since this `buildAttachmentOptionItems` function needs to be created inside a composable context

yamilmedina: I think indeed will help in not passing around or getting the context each time for every sharing flow.
This changes with the other from below on the Inner state, makes life easier when calling the function 👍

trOnk12: ahhh makes sense, sure no worries

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInnerState.kt
yamilmedina: I don't think we should have a problem with that `context`, I like the idea cause now the context it's getting assigned only once and not for every sharing flow

trOnk12 (approved):
good job 🥇  💯

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewState.kt
```diff
@@ -1,10 +1,14 @@
...
+import kotlinx.parcelize.Parcelize
 
 data class ConversationViewState(
     val conversationName: String = "",
     val messages: List<Message> = emptyList(),
-    val messageText : TextFieldValue = TextFieldValue("")
+    val messageText: TextFieldValue = TextFieldValue("")
 )
+
+
```
gongracr, line 14:
```suggestion
```

gongracr (dismissed):
Great effort with all the image sharing UI! 👏 👨‍🎨 🖼️

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInnerState.kt
```diff
@@ -60,6 +71,33 @@ class MessageComposerInnerState(
...
+    fun pickAttachment(context: Context, attachmentUri: Uri) {
+        attachmentState = try {
+            val attachment =
+                AttachmentBundle(
+                    attachmentUri.getMimeType(context).orDefault(DEFAULT_FILE_MIME_TYPE),
+                    attachmentUri.toByteArray(context)
+                )
+            AttachmentState.Picked(attachment)
+        } catch (e: IOException) {
+            AttachmentState.Error
```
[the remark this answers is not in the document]
yamilmedina: Just pushed the changes to have a log in case of failure 👍

gongracr (approved):

PR #445, opened 2022-03-23
chore: upload APK to internal lane on Google Play for commits on `develop` branch too

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Current builds were only uploading the Internal lane on Google Play for commits on `main` branch.

### Solutions

Since the Jenkinsfile trackName is hardcoded to `internal`, with this PR, every commit on `main`, `release` or `develop` branches will trigger an upload of the app's release APK to Google Play , so that it can be tested by design team at any time.

----

----
##### References

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,9 +8,9 @@ def defineFlavor() {
     def branchName = env.BRANCH_NAME
     if (branchName == "main") {
         return 'Internal'
-    } else if(branchName == "develop") {
+    } else if (branchName == "develop") {
         return 'Dev'
-    } else if(branchName == "release") {
+    } else if (branchName == "release") {
         return 'Public'
     }
     return 'Dev'
@@ -314,7 +314,7 @@ pipeline {
           }
           stage('Playstore') {
             when {
-              expression { env.trackName != 'None' && env.flavor != 'Dev' && env.CHANGE_ID == null }
+              expression { env.trackName != 'None' && env.CHANGE_ID == null }
             }
             steps {
               echo 'Checking folder before playstore upload'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -314,7 +314,7 @@ pipeline {
           }
           stage('Playstore') {
             when {
-              expression { env.trackName != 'None' && env.flavor != 'Dev' && env.CHANGE_ID == null }
+              expression { env.trackName != 'None' && env.CHANGE_ID == null }
```
MohamadJaara, line 317: This will make it upload to play store on every branch 
```groovy
def defineFlavor() {
    //check if the pipeline has the custom flavor env variable set
    def overwrite = env.CUSTOM_FLAVOR
    if(overwrite != null) {
        return overwrite
    }
    def branchName = env.BRANCH_NAME
    if (branchName == "main") {
        return 'Internal'
    } else if(branchName == "develop") {
    } else if (branchName == "develop") {
        return 'Dev'
    } else if(branchName == "release") {
    } else if (branchName == "release") {
        return 'Public'
    }
    return 'Dev' // 'Dev' here is the defualt flaver name 
}
```

MohamadJaara (changes requested):
I was checking Jenkins, and we already upload an apk to S3, and we can add a step to comment the S3 link to the jera if it's liked 
this will be a much better option
And updating the internal play store is a bad idea (what I was thinking when I suggested this)

gongracr:
Let's then favour the least frictious option and try to give access to the artifacts uploaded to the S3 buckets to those interested

PR #446, opened 2022-03-24
feat: token refresh

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

accommodate the changes from https://github.com/wireapp/kalium/pull/315

----
##### References


app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -124,62 +126,60 @@ class UseCaseModule {
...
         @KaliumCoreLogic coreLogic: CoreLogic,
-        @CurrentSession currentSession: AuthSession
+        @CurrentAccount currentAccount: String
     ): SendTextMessageUseCase =
-        coreLogic.getSessionScope(currentSession).messages.sendTextMessage
+        coreLogic.getSessionScope(currentAccount).messages.sendTextMessage
+
+    @ViewModelScoped
+    @Provides
+    fun logoutUseCaseProvider(@KaliumCoreLogic coreLogic: CoreLogic, @CurrentAccount currentAccount: String): LogoutUseCase =
```
alexandreferris, line 182: As we are now passing the current account / user ID could this be renamed to currentAccountID? so it makes more sense, as for my understanding `currentAccount` would be an `Account` class and not only the ID as string 🤔 

If so, then the same for the parameters changed above

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
```diff
@@ -107,6 +109,7 @@ class UserProfileViewModel @Inject constructor(
         // TODO
         viewModelScope.launch {
             dataStore.clear() // TODO this should be moved to some service that will clear all the data in the app
+            logoutUseCase()
```
alexandreferris, line 112: Could this bring any problems if we are calling after `dataStore.clear()`? Like could the logoutUseCase use anything from the dataStore when really logging the user out?

app/src/main/kotlin/com/wire/android/ui/userprofile/image/AvatarPickerViewModel.kt
```diff
@@ -59,7 +59,7 @@ class AvatarPickerViewModel @Inject constructor(
             withContext(Dispatchers.IO) {
                 val data = imgUri.toByteArray(context)
                 val mimeType = imgUri.getMimeType(context) ?: DEFAULT_IMAGE_MIME_TYPE
-                val result = uploadUserAvatar(mimeType = mimeType, imageData = data)
+                val result = uploadUserAvatar( imageData = data)
```
alexandreferris, line 62:
```suggestion
                val result = uploadUserAvatar(imageData = data)
```

aaaand 🙌🏻  no more erros with unmatching parameter

app/src/main/kotlin/com/wire/android/ui/userprofile/UserProfileViewModel.kt
MohamadJaara: The app Datastore now is used for some properties like showing suppressing warnings for changing user status, and its implementation in the app is not optimal  since it does not take multiple users into consecration.
 Back to your question now kalium needs no data from the app, it's the other way around so deleting app data before kalium data should not be an issue.

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
alexandreferris: We can do this in a later stage

alexandreferris (approved):

PR #507, opened 2022-04-13
feat: display avatars (AR-897) (AR-1181)

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-897" title="AR-897" target="_blank">SubtaskAR-897</a>  Update GetConversationsUseCase to include an URI to the conversation avatar for 1:1 Conversations
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We're not currently displaying the avatars of self and other users in a lot of screens

### Solutions

Implement `WireSessionImageLoader`, which wraps Coil inside of it, and exposes a `paint()` function to use with Compose.
This `paint()` function can take in any public asset (we can later expand this to private assets too).

Internally, the loader uses `Coil` and implements a custom `Fetcher` that takes an `AssetId` and uses Kalium to fetch the data.

I've replaced the placeholders with this wherever I could find.

As mentioned before, we can expand this to load private assets too and use this to load MessageImages for example.

#### How to Test

Open any screen, basically 😋 

### Attachments

Screenshot_1649871261
Screenshot_1649871149
Screenshot_1649870070
Screenshot_1649871362

----

----
##### References


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-507/1/) **failed**.

vitorhugods:
rebuild

app/src/main/kotlin/com/wire/android/ui/common/CurrentSessionViewModel.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import androidx.lifecycle.ViewModel
+import com.wire.android.util.ui.WireSessionImageLoader
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+
+@HiltViewModel
+class CurrentSessionViewModel @Inject constructor(
+    val wireSessionImageLoader: WireSessionImageLoader
+): ViewModel()
```
vitorhugods, line 11: It's a dummy ViewModel so we can inject the `ImageLoader` into composables.

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-507/4/) **succeeded**.

- [com.wire.android-vv0.0.1(3090820)(3090820)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(3090820)(3090820)-dev-release.apk)

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -143,7 +144,10 @@ class ConversationListViewModel @Inject constructor(
                 val otherUser = details.otherUser
                 GeneralConversation(
                     ConversationType.PrivateConversation(
-                        userInfo = UserInfo("", UserStatus.NONE), //TODO Get actual status and avatar
+                        userInfo = UserInfo(
+                            otherUser.completePicture?.let { UserAvatarAsset(it) },
```
gongracr, line 148: So we load here in the conversation avatar row the complete picture but the preview one in the HomeTopBar?

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
```diff
@@ -18,24 +18,25 @@ import androidx.compose.ui.platform.LocalContext
...
-// TODO: in the future we would not pass here null, but while developing there may be users having no avatar asset
-    avatarAssetByteArray: ByteArray? = null,
+    userAvatarAsset: UserAvatarAsset? = null,
     status: UserStatus = UserStatus.NONE,
     isClickable: Boolean = false,
     size: Dp = MaterialTheme.wireDimensions.userAvatarDefaultSize,
     modifier: Modifier = Modifier,
     onClick: (() -> Unit)? = null
 ) {
+    val imageLoader = hiltViewModel<CurrentSessionViewModel>().wireSessionImageLoader
```
trOnk12, line 38: nice workaround :P however I can see a problem with this, from what I know hiltViewModel() gets a ViewModel instance by scoping it to NavBackStackEntry, so if we use it in different places in our app, each composable() or NavHost will create its own instance of this ViewModel, which will create its own instance of WireSessionImageLoader().

If this work-around is the only way we can use this ImageLoader inside UserProfileAvatar, perhaps we can make WireSessionImageLoader a Singleton. 

I never did such things in Coil like you did here , which I am impressed by :P but by quickly looking into the docs it looks like you can set the ImageLoader instance with this : 

```
  @JvmStatic
    @Synchronized
    fun setImageLoader(imageLoader: ImageLoader) {
        this.imageLoaderFactory = null
        this.imageLoader = imageLoader
    }
```
    
it looks like for example this method : 
```

@Composable
fun rememberAsyncImagePainter(
    model: Any?,
    transform: (State) -> State = DefaultTransform,
    onState: ((State) -> Unit)? = null,
    contentScale: ContentScale = ContentScale.Fit,
    filterQuality: FilterQuality = DefaultFilterQuality,
) = rememberAsyncImagePainter(
    model = model,
    imageLoader = LocalImageLoader.current,
    transform = transform,
    onState = onState,
    contentScale = contentScale,
    filterQuality = filterQuality
)    
```

makes use of LocalImageLoader.current, which in return looks like takes this singleton instance from Coil which you can set with the method above :p 

what I want to say, maybe it is worth to investigate if we can do it so that we just set this ImageLoader once and then coil does the rest for us so that we can keep using rememberAsyncImagePainter() for example without the work-around above, like we can configure Retrofit for example by adding some interceptor to the instance of Retrofit 

if there is no way around it, I am also fine with this work-around, no such a big deal and if it works correctly why not just keep it, it does not bring that much of confusion

vitorhugods: The main issue is that we cannot have a generic static image loader.

It needs to be scoped down to the user/session level.

If we switch from one session to another, the image loader should change in order to be able to load the assets properly (different backends, authentication, or even private assets).

So, it's impossible to get away with a singleton. Unless we start managing the state of this singleton, and keep updating the instance on the go, which sounds a bit weird, manually managing these moving pieces on each session change.

app/src/main/kotlin/com/wire/android/ui/userprofile/common/UserProfileInfo.kt
```diff
@@ -52,9 +53,9 @@ fun UserProfileInfo(
     ) {
         UserProfileAvatar(
             onClick = onUserProfileClick,
-            isClickable = if (editableState is EditableState.NotEditable) false else !isLoading,
+            isClickable = true,
```
gongracr, line 56: Wouldn't this make also other user profile avatars clickable?

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
vitorhugods: I added some explanation on the scope and the problems this tries to solve [here](https://github.com/wireapp/wire-android-reloaded/blob/app/src/main/kotlin/com/wire/android/util/ui/WireSessionImageLoader.kt/b03cf0dae88df723e6e2569df6d05f6784065975).

app/src/main/kotlin/com/wire/android/util/ui/AssetImageFetcher.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.wire.android.model.UserAvatarAsset
+import com.wire.android.util.toBitmap
+import com.wire.kalium.logic.feature.asset.GetAvatarAssetUseCase
+import com.wire.kalium.logic.feature.asset.PublicAssetResult
+
+class AssetImageFetcher(
+    val data: UserAvatarAsset,
+    val getAvatarAsset: GetAvatarAssetUseCase,
+    val resources: Resources,
+    val imageLoader: ImageLoader
```
gongracr, line 20: are we planning to use this or can we remove it?

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
vitorhugods: One example:

User A is logged in, we set the static instance of the ImageLoader and it's able to load assets (private, public, etc) for User A.

User B logs in on the same app, what happens? I wanted to leave DI to handle this, and keep the behaviour consistent with all the other UseCases and ViewModels.
If we could have a custom scope, that we initiate and destroy on a Session basis, that can work too.

gongracr (approved):
Awesome work! Just some tiny comments that don't really block anything. 🍔🍟🥤

app/src/main/kotlin/com/wire/android/ui/calling/OngoingCallScreen.kt
```diff
@@ -27,6 +29,7 @@ import com.wire.android.ui.common.topappbar.NavigationIconType
 import com.wire.android.ui.common.topappbar.WireCenterAlignedTopAppBar
 import com.wire.android.ui.theme.wireColorScheme
 import com.wire.android.ui.theme.wireDimensions
+import com.wire.android.util.ui.WireSessionImageLoader
```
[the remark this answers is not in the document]
vitorhugods: Yes. 🥲

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
vitorhugods: Mistakes were made 😬

I'll use the preview for these tiny guys.

app/src/main/kotlin/com/wire/android/ui/userprofile/common/UserProfileInfo.kt
vitorhugods: Ooohh, I made a mistake here too.

It was supposed to remove the `isLoading` part. As the ViewModel has no control about the avatar loading anymore.


Or even leave it the way it was, and just consider if the profile is being loaded.

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
gongracr: It seems the link to the explanation on the scope is broken :/

trOnk12: I do not think I have enough concept to understand what you are saying ;p but if DI let us inject those ImageLoaders provide the proper use case to them then it should not be a problem if the user relogs because he reopens the app and reinitializes the dependency injection

app/src/main/kotlin/com/wire/android/util/ui/AssetImageFetcher.kt
vitorhugods: I'm not fully sure yet.

It depends on whether the custom-fetched images are Cached by Coil or not. My gut feeling is that no, they aren't.

But I don't want to solve caching in this PR. I'd leave it for a next one.

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
trOnk12: from what I see all we want to do is pass the instance of GetAvatarAssetUseCase to the ImageLoader, in theory I think it should be possible to configure it for example inside the @Provides method since the Coil instance has a method that allows us to set this ImageLoader, but it is all a theory at this point ;P

vitorhugods: > It seems the link to the explanation on the scope is broken :/

I'm on mobile, it seems the App doesn't let me get a proper link, but I tried to link to the documentation on `WireSessionImageLoader`.

vitorhugods: > from what I see all we want to do is pass the instance of GetAvatarAssetUseCase to the ImageLoader, in theory I think it should be possible to configure it for example inside the @Provides method since the Coil instance has a method that allows us to set this ImageLoader, but it is all a theory at this point ;P

DI is not initialized again.

`Application#onCreate` won't be called again when we just switch from one user to another.

I think only the activity(es) are recreated. So we can scope the ImageLoader to the main Activity for example.


Still, I guess one improvement we can have in the future is some custom scope, where we can start/stop in order to manipulate DI.

trOnk12: then again in theory we could inject for example the ImageLoader.Factory into the onCreate of Activityu and configure the Coil by setting the ImageLoader on the Coil instance or something like that, again this is all in theory do not know if it is possible by doing so, if not please stop me telling you bullshit :P

vitorhugods: Then we need to configure Coil in multiple places, DI and Activity for setting the instance.

I think we can just leave in DI. Together with everything else, instead of needing to actively care about changing the singleton instance.

We definitely need our own DI component/scope for the future, something we don't have yet.
Once we have that, we can have a single loader that is only recreated if the current user session changes.

Also, calling directly the coil functions is not great if we ever want to replace it with some other solution.

trOnk12: I do not understand why we would need to configure the Coil in multiple places, you can just do it once on Activity creation, regards calling coil functions I do not really know why we would switch to other solutions hmmm if we do it is a matter of changing one method inside UserProfileAvatar

trOnk12: I tried to do something like this : 

```
    @Inject
    lateinit var wireSessionImageLoader: WireSessionImageLoader

    val viewModel: WireActivityViewModel by viewModels()

        installSplashScreen()
        viewModel.handleDeepLink(intent)

        Coil.setImageLoader(wireSessionImageLoader.coilImageLoader)
```

but indeed there is an issue with some stuff scoped to the ViewModelComponent, so for now I think, you approach is the best we got

trOnk12 (approved):
looks good ! :) at the moment the best we got :) i still think we could omit this work-around in the future by confugiring the coil just once and let the custom ImageLoader be consumed by Coil, but one step at a time ! :)

app/src/main/kotlin/com/wire/android/ui/common/CurrentSessionViewModel.kt
vitorhugods: Right now, even if we use `WireActivityViewModel`, it will instantiate a `WireActivityViewModel` for each composable using it. So it's... not much of an improvement.

I'll see about creating a custom Dagger component/scope next week.

This way we will have a solid scope for each user session, and we can have a singleton image loader inside of it.

app/src/main/kotlin/com/wire/android/ui/common/UserProfileAvatar.kt
vitorhugods: Next week: custom dagger component/scope.

AndroidBob:
Build [6](https://10.10.124.134/job/wire-android-reloaded/job/PR-507/6/) **failed**.

AndroidBob:
Build [9](https://10.10.124.134/job/wire-android-reloaded/job/PR-507/9/) **succeeded**.

- [com.wire.android-vv0.0.1(3096716)(3096716)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(3096716)(3096716)-dev-release.apk)

PR #566, opened 2022-05-06
fix: asset size errors [AR-1476]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When user tries to upload large asset files, the app crashes with an `OutOfMemoryError`. This is due to fact that we pass around and manipulate the asset data in a ByteArray form, so when performing some operation that require these byte arrays to be copied, such exceptions are thrown. 

### Solutions

In order to solve this problem, one could decide between two solutions:
1. **The dirty and quick one**: Launching the app with a large memory heap flag + setting up an asset send limit. This will permit users send assets that weigh up to 100MB.  This would however cause that the garbage collector runs less often than with a normal heap limit, increasing the time until we notice about certain memory leaks, and making the debug of these more complicated.
2. **The clean but laborious one**: Instead of handling the raw data on the shape of ByteArrays, we could use files or inputStream like methods to make the upload, download, encryption and decryption processes. The main problem is that Ktor doesn't provide such methods to certain targets like iOS and JS. Therefore we would need to use a different library for this. A strong candidate would be (Okio)[https://square.github.io/okio/], as it seems it offers already a set of helper methods supporting Files and data stream for all platforms. 

For the moment, and due to the pressing circumstances to deliver an MVP, I'm proposing to go with option **1.** 
However, a future refactor to accommodate for option 2. will be much needed once we are done with the first prototype of the app.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
@@ -26,6 +26,8 @@ import com.wire.android.ui.common.bottomsheet.MenuItemIcon
 import com.wire.android.ui.common.bottomsheet.MenuModalSheetLayout
 import com.wire.android.ui.common.dimensions
 import com.wire.android.ui.common.snackbar.SwipeDismissSnackbarHost
+import com.wire.android.ui.home.conversations.ConversationErrors.ERROR_MAX_IMAGE_SIZE
+import com.wire.android.ui.home.conversations.ConversationErrors.ERROR_SENDING_IMAGE
 import com.wire.android.ui.home.conversations.delete.DeleteMessageDialog
 import com.wire.android.ui.home.conversations.mock.getMockedMessages
 import com.wire.android.ui.home.conversations.model.AttachmentBundle
@@ -50,7 +52,8 @@ fun ConversationScreen(conversationViewModel: ConversationViewModel) {
         onImageFullScreenMode = { conversationViewModel.navigateToGallery(it) },
         onBackButtonClick = conversationViewModel::navigateBack,
         onDeleteMessage = conversationViewModel::showDeleteMessageDialog,
-        onCallStart = audioPermissionCheck::launch
+        onCallStart = audioPermissionCheck::launch,
+        onError = conversationViewModel::onError
     )
     DeleteMessageDialog(conversationViewModel = conversationViewModel)
 }
@@ -74,7 +77,8 @@ private fun ConversationScreen(
     onImageFullScreenMode: (String) -> Unit,
     onBackButtonClick: () -> Unit,
     onDeleteMessage: (String, Boolean) -> Unit,
-    onCallStart: () -> Unit
+    onCallStart: () -> Unit,
+    onError: (ConversationErrors) -> Unit
 ) {
     val conversationScreenState = rememberConversationScreenState()
     val scope = rememberCoroutineScope()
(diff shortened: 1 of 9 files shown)


AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-566/3/) **succeeded**.

- [com.wire.android-vv0.0.1(3287770)(3287770)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(3287770)(3287770)-dev-release.apk)

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
```diff
@@ -141,6 +143,15 @@ private fun ConversationScreen(
     }
 }
 
+@Composable
+fun getErrorMessage(errorCode: ConversationErrors) =
+    when (errorCode) {
+        ErrorMaxAssetSize -> stringResource(R.string.error_conversation_max_asset_size_limit)
+        ErrorMaxImageSize -> stringResource(R.string.error_conversation_max_image_size_limit)
+        ErrorSendingSize -> stringResource(R.string.error_conversation_sending_image)
```
yamilmedina, line 151: I'm wondering if this `ErrorSendingSize` should be named to `ErrorSendingAsset` since this is the actual response from the server 🤔

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationViewModel.kt
```diff
@@ -141,29 +145,42 @@ class ConversationViewModel @Inject constructor(
...
+                                val (imgWidth, imgHeight) = extractImageParams(attachmentBundle.rawContent)
+                                val result = sendImageMessage(
+                                    conversationId = conversationId,
+                                    imageRawData = attachmentBundle.rawContent,
+                                    imageName = attachmentBundle.fileName,
+                                    imgWidth = imgWidth,
+                                    imgHeight = imgHeight
+                                )
+                                if (result is SendImageMessageResult.Failure) {
+                                    onError(ErrorSendingSize)
```
yamilmedina, line 159: This is similar as the comment about this class naming, we should use something more generic like `ErrorSendingAsset`

yamilmedina (commented):
I like the approach, we are not in a safe place putting some limits to the assets the users can send 👏 🍷  Just added some re-naming suggestions

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
MohamadJaara: The asset limit exists even in the BE, the only problem I see with not going with solution 2 is DB migration later will not be easy.

AndroidBob:
Build [5](https://10.10.124.134/job/wire-android-reloaded/job/PR-566/5/) **succeeded**.

- [com.wire.android-vv0.0.1(3295035)(3295035)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(3295035)(3295035)-dev-release.apk)

MohamadJaara:
> I like the approach, we are not in a safe place putting some limits to the assets.

The asset limit exists even in the BE, the only problem I see with not going with solution 2 is DB migration later will not be easy.
I suggest NOT to go through this change unless we have a migration strategy in place.

AndroidBob:
Build [7](https://10.10.124.134/job/wire-android-reloaded/job/PR-566/7/) **failed**.

yamilmedina (approved):
Great work as always compa! 👏  🎸

PR #675, opened 2022-06-16
chore: Back Navigation improve

borichellow:
# What's new in this PR?

- change handling `NavigationManager.navigateBack()`
- added new `BackStackMode.REMOVE_CURRENT` it removes the current item from the navigation backStack before navigating to the new one.

### Issues

In cases when `NavController.popBackStack()` is called and the current screen is the only item in backStack, then empty screen is displayed. 
That issue could appear when the app was opened by `PendingIntent` from the notification and `NavController.popBackStack()` is called (NOT android back button).
Example scenario: 
- kill the app on device1
- call from device2 to device1
- device1 shows IncomingCallScreen (it's not yet implemented but will be soon :) )
- cancel a call on device2, or decline a call on device1

As a result: 
empty screen is displayed (kinda default compose-navigation screen, before even `startDestination` is set)

### Causes (Optional)

As texed before: when we call `NavController.popBackStack()` and there is no backstack in NavController then empty screen is shown. 

### Solutions

To solve that we check if `NavController.popBackStack()`  popped stack at least once (`popBackStack()` returns `true` in that case) and if not - we assume that was the very first screen and we need to close the whole app (`finish()` the WireActivity).

The other problem occurs after that solution: some of the screens (IncomingCallScreen e.g.) does something like: 
step1:`navigateBack` 
step2: `navigateToSomeScreen` 
because we need to remove a current screen from a backStack and navigate to another screen. 

But with the solution described above, we'll get the whole app closed on step1, if the current screen was a very first (for example IncomingCallScreen was opened from the PendingIntent in notifications).

So to solve that secondary issue we add `BackStackMode.REMOVE_CURRENT`, which does all the `navigateBack+navigateTo` magic by itself.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
@@ -23,21 +23,37 @@ internal fun navigateToItem(
                     }
                 }
             }
-            BackStackMode.NONE -> {}
+            BackStackMode.REMOVE_CURRENT -> {
+                navController.run {
+                    println("cyka 0 ${backQueue.map { it.destination.route }}")
+                    backQueue.lastOrNull { it.destination.route != null }?.let { entry ->
+                        println("cyka 1 ${entry.destination.route}")
+                        val inclusive = true
+                        val startId = entry.destination.id
+                        popBackStack(startId, inclusive)
+                    }
+                }
+            }
+            BackStackMode.NONE -> {
+            }
         }
         launchSingleTop = true
         restoreState = true
     }
 }
 
-internal fun NavController.popWithArguments(arguments: Map<String, Any>?) {
+/**
+ * @return true if the stack was popped at least once and the user has been navigated to another destination,
+ * false otherwise
+ */
+internal fun NavController.popWithArguments(arguments: Map<String, Any>?): Boolean {
     previousBackStackEntry?.let {
         arguments?.forEach { (key, value) ->
             appLogger.d("Destination is ${it.destination}")
             it.savedStateHandle[key] = value
         }
     }
-    popBackStack()
+    return popBackStack()
 }
 
 @ExperimentalMaterial3Api
--- a/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
@@ -25,9 +25,7 @@ internal fun navigateToItem(
             }
             BackStackMode.REMOVE_CURRENT -> {
                 navController.run {
-                    println("cyka 0 ${backQueue.map { it.destination.route }}")
                     backQueue.lastOrNull { it.destination.route != null }?.let { entry ->
-                        println("cyka 1 ${entry.destination.route}")
                         val inclusive = true
                         val startId = entry.destination.id
                         popBackStack(startId, inclusive)
--- a/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
@@ -32,8 +32,7 @@ internal fun navigateToItem(
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
```diff
@@ -23,21 +23,37 @@ internal fun navigateToItem(
                     }
                 }
             }
-            BackStackMode.NONE -> {}
+            BackStackMode.REMOVE_CURRENT -> {
+                navController.run {
+                    println("cyka 0 ${backQueue.map { it.destination.route }}")
```
mchenani: Do we need this log? I we do, please use the appLogger.

borichellow: removed it, sorry ;P

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -83,16 +83,18 @@ class WireActivity : AppCompatActivity() {
...
-                    keyboardController?.hide()
                     navigateToItem(navController, command)
                 }
                 .launchIn(scope)
 
             navigationManager.navigateBack
                 .onEach {
-                    keyboardController?.hide()
-                    navController.popWithArguments(it)
+                    if (!navController.popWithArguments(it)) finish()
```
mchenani, line 92: A question, do we have anything that might be affected by finishing the activity by `finish()`?

mchenani: @vitorhugods do you have any comment?

vitorhugods: Killing the activity shouldn't do much IMHO. Same as the user killing the app by swiping it up.

If there are new events, a push will arrive, sync will be performed and we're back online with notifications being displayed.

app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
```diff
@@ -23,21 +23,35 @@ internal fun navigateToItem(
...
+                navController.run {
+                    backQueue.lastOrNull { it.destination.route != null }?.let { entry ->
+                        val inclusive = true
+                        val startId = entry.destination.id
+                        popBackStack(startId, inclusive)
+                    }
+                }
+            }
+            BackStackMode.NONE -> {
+            }
```
gongracr, line 36:
```suggestion
            BackStackMode.NONE -> {}
```

app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
```diff
@@ -23,21 +23,35 @@ internal fun navigateToItem(
...
         }
         launchSingleTop = true
         restoreState = true
     }
 }
 
-internal fun NavController.popWithArguments(arguments: Map<String, Any>?) {
+/**
+ * @return true if the stack was popped at least once and the user has been navigated to another destination,
```
gongracr, line 44:
```suggestion
 * @return true if the stack was popped at least once and the user navigated to another destination,
```

ohassine (approved):
I like the improvement 👌 well done ✅

gongracr (approved):
Looks good to me! I think we discussed having this extra flag when adding the `popWithArguments()` but not sure why it never went through. I hope this also fixes some other navigation issues keeping screen arguments that we had in the past too! 🍸

ohassine:
Could you apply this improvement also on `InitiatingCallViewModel` ?

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-675/4/) **succeeded**.

- [com.wire.android-vv0.0.1(3649475)(3649475)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(3649475)(3649475)-dev-release.apk)

PR #698, opened 2022-06-22
fix: back navigation with arguments

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There is an issue with navigating back with arguments, passing such arguments to the ViewModel and clearing them when not needed anymore.

### Solutions

Created SavedStateViewModel - ViewModel which already has SavedStateHandle parameter by default. Implemented hiltSavedStateViewModel method which takes NavBackStackEntry as an argument to pass the arguments to the ViewModel's SavedStateHandle. Every navigation (back and forward) updates the back navigation arguments of the current or previous entry. Added external functions for easier getting navigation arguments from SavedStateHandle.

#### How to Test

Currently it's used in two situations:
- removing image from media gallery screen, it should remove message after going back to conversation screen
- ignoring connection, it should show snackbar after going back to home screen

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/navigation/SavedStateViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/SavedStateViewModel.kt
@@ -15,3 +15,8 @@ inline fun <reified T : SavedStateViewModel> hiltSavedStateViewModel(navBackStac
             savedStateHandle[EXTRA_BACK_NAVIGATION_ARGUMENTS] = it.savedStateHandle.get<Map<String, Any>>(EXTRA_BACK_NAVIGATION_ARGUMENTS)
         }
     }
+
+fun <V: Any> SavedStateHandle.getBackNavArgs() = get<Map<String, V>>(EXTRA_BACK_NAVIGATION_ARGUMENTS) ?: mapOf()
+
+@Suppress("UNCHECKED_CAST")
+fun <V: Any> SavedStateHandle.getBackNavArg(key: String): V? = getBackNavArgs<Any>()[key] as? V
--- a/app/src/main/kotlin/com/wire/android/navigation/SavedStateViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/SavedStateViewModel.kt
@@ -1,0 +1,17 @@
+package com.wire.android.navigation
+
+import androidx.compose.runtime.Composable
+import androidx.hilt.navigation.compose.hiltViewModel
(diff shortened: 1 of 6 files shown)


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-698/1/) **succeeded**.

- [com.wire.android-vv0.0.1(3695321)(3695321)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(3695321)(3695321)-dev-release.apk)

app/src/main/kotlin/com/wire/android/navigation/SavedStateViewModel.kt
```diff
@@ -0,0 +1,22 @@
+package com.wire.android.navigation
+
+import androidx.compose.runtime.Composable
+import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import androidx.navigation.NavBackStackEntry
+
+open class SavedStateViewModel(open val savedStateHandle: SavedStateHandle) : ViewModel()
```
yamilmedina, line 9: Does this class is meant to be inherited from every `ViewModel` that needs to handle a `SavedStateHandle` ?
As a suggestion, maybe we could add a doc/comment, so in the future we know when to use it :D

yamilmedina (approved):
Hey amazing work ! I just left a small question/suggestion, so we can take advantage of this new class in the future 👏  🚀

Garzas (approved):
For me this is something new how now navigation works in android compose 😅

app/src/main/kotlin/com/wire/android/navigation/SavedStateViewModel.kt
saleniuk: For every `ViewModel` that wants to receive back navigation arguments, so there can be also regular `ViewModel` objects that have `SavedStateHandle`, they just won't get the back navigation arguments, so for consistency and safety it would be better to just use `SavedStateViewModel`. 😌 
I was trying to do it without providing a new class, but I really didn't want to use reflection or any other sorcery. 😄 
I added a commend and updated this [doc about Jetpack Compose & Navigation](https://wearezeta.atlassian.net/wiki/spaces/AR/pages/547458852/Jetpack+Compose+Navigation).

PR #886, opened 2022-08-23
feat: handle offline status when starting a call (AR-2145)

gongracr:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-2145" title="AR-2145" target="_blank">SubtaskAR-2145</a>  Calling  - Popup to signal offline therefore functionality is not available
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When initiating a call while the client was not connected to the internet, we were still launching the `InitiatingCallScreen` which didn't make sense. 

### Solutions

We will be showing a dialog informing about the lack of connectivity every time the user tries to start a call w/o a stable internet connection.

### Dependencies (Optional)

Needs releases with:
https://github.com/wireapp/kalium/pull/827

#### Test Coverage (Optional)

### Attachments (Optional)
[UPDATED]

https://user-images.githubusercontent.com/2468164/186407938-780764bb-ffbd-4d37-a4c2-30a626e3bf62.mov

----

----

----
##### References


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-886/1/) **failed**.

ohassine:
I think we don't have to show the outgoing screen if we don't have an Internet ?

For the string what about using strings from old app ?

Wire 2022-08-23 at 2_52 PM

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-886/2/) **succeeded**.

- [com.wire.android-vv0.0.1(4230647)(4230647)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(4230647)(4230647)-dev-release.apk)

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallScreen.kt
```diff
@@ -52,15 +58,16 @@ fun InitiatingCallScreen(
...
+    errorState: InitiatingCallState.InitiatingCallError,
+    onErrorStartingCall: () -> Unit,
     callState: CallState,
     toggleMute: () -> Unit,
     toggleSpeaker: () -> Unit,
     toggleVideo: () -> Unit,
     onHangUpCall: () -> Unit,
     onVideoPreviewCreated: (view: View) -> Unit,
-    onSelfClearVideoPreview: () -> Unit
+    onSelfClearVideoPreview: () -> Unit,
```
ohassine, line 69: we don't need this comma  💁‍♂️

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
```diff
@@ -116,27 +123,37 @@ class InitiatingCallViewModel @Inject constructor(
...
-            conversationType = conversationType
-        )
-        callRinger.ring(R.raw.ringing_from_me)
+    internal suspend fun initiateCall() {
+        val conversationType = retrieveConversationType()
+        withContext(dispatchers.main()) {
+            val result = startCall(
+                conversationId = conversationId,
+                conversationType = conversationType
+            )
```
ohassine, line 132: `startCall` should not be run on main

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
```diff
@@ -116,27 +123,37 @@ class InitiatingCallViewModel @Inject constructor(
...
-            launch {
-                navigateBack()
-                callRinger.stop()
-            }
-        }
+    fun hangUpCall() = viewModelScope.launch(dispatchers.io()) {
+        endCall(conversationId)
+        callRinger.stop()
+        navigateBack()
+    }
```
ohassine, line 153: Why did you remove all the coroutines ? We need to run them in parallel..
This way you will see some delay before the screen is closed. `endCall` will take some time before it's done

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
```diff
@@ -20,28 +25,31 @@ import com.wire.kalium.logic.feature.call.usecase.ObserveEstablishedCallsUseCase
...
     qualifiedIdMapper: QualifiedIdMapper,
+    private val navigationManager: NavigationManager,
     private val conversationDetails: ObserveConversationDetailsUseCase,
     private val observeEstablishedCalls: ObserveEstablishedCallsUseCase,
     private val startCall: StartCallUseCase,
     private val endCall: EndCallUseCase,
     private val isLastCallClosed: IsLastCallClosedUseCase,
-    private val callRinger: CallRinger
+    private val callRinger: CallRinger,
+    private val dispatchers: DispatcherProvider
```
ohassine, line 48: I think it's better to change the dispatcher in the suspend function it self and not in the viewModel(eg. usecase). This way we don't have to worry about what Dispatcher to use for what type of work.

**This responsibility lies in the class that does the work.**

ohassine (changes requested):
Seeing the calling UI when we don't have an Internet connection is confusing, it gives me the feeling that the call already started and then the Internet stopped which not the case.

We should display the popup before starting the call I think.

Also in this way, by starting the call we are triggering unnecessary work like initializing AVS and invoking some calling methods

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallScreen.kt
vitorhugods: I think we never agreed on dangling commas, but it's not that bad. It may look a bit uglier now, but it helps in the future when we add more items to the list:

grafik

Credits: https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
gongracr: io? or the default thread then?

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
gongracr: you were launching 3 threads here, 1 containing another 2. Could we do it then with just 1 extra one to end the call and leave the current onging thread to run the tasks to stop the ring and navigate back?

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
gongracr: Hmmm not sure if I follow. I do use the dispatchers to change the context on the corresponding method inside the view model. 

If you mean that the suspend function calling the view model should hold the responsibility (`InitiateCallScreen` for instance) you are implying that the view layer knows already the kind of logic that is being performed on the VM (whether if it is a IO operation or not) which I'm not sure it is the case.

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-886/4/) **failed**.

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
ohassine: I am lunching 3 coroutines 1 parent and 2 children. This will not affect the app performance anymore since coroutines are cheap compared to threads. The most important thing here is that they should run in parallel

app/src/main/kotlin/com/wire/android/ui/calling/initiating/InitiatingCallViewModel.kt
ohassine: I mean we don't need to change dispatcher here, we should do it in the suspend function itself(usecase for example)
All that we need to do here is starting a new coroutine and call the functions

gongracr: addressed

AndroidBob:
Build [6](https://10.10.124.134/job/wire-android-reloaded/job/PR-886/6/) **failed**.

AndroidBob:
Build [8](https://10.10.124.134/job/wire-android-reloaded/job/PR-886/8/) **failed**.

AndroidBob:
Build [9](https://10.10.124.134/job/wire-android-reloaded/job/PR-886/9/) **succeeded**.

- [com.wire.android-vv0.0.1(4244435)(4244435)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(4244435)(4244435)-dev-release.apk)

PR #915, opened 2022-08-31
fix: Conversation options doesn't open after sending app to background (AR-2177)

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-2177" title="AR-2177" target="_blank">BugAR-2177</a>  + button on conv details doesn't expand after sending app background
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

When clicking on add + button, it expands the menu. If I send back to the background, open the app again but click the app icon (not recent apps), the button gets selected, but menu is not open

### Causes (Optional)

When app just returns from the background there is some moment when "available screen" space includes Android Navigation buttons space (and we remember that as a `fullScreenHeight`). So when re-composition is called it find out that the screen size is 24.dp smaller than the remembered one and assumes that keyboard height is 24.dp :) 

### Solutions

made the keyboard height be at least `DEFAULT_KEYBOARD_TOP_SCREEN_OFFSET` 
also added some checks: if the keyboard height is already known, then do not update it

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
@@ -222,9 +222,12 @@ private fun ConversationScreen(
                     }
 
                     // if the currentScreenHeight is smaller than the initial fullScreenHeight
+                    // and we don't know the keyboard height yet
                     // calculated at the first composition of the ConversationScreen, then we know the keyboard size
-                    if (currentScreenHeight < fullScreenHeight) {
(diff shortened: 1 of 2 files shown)


trOnk12 (approved):
great job 🥇

AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-915/1/) **succeeded**.

- [com.wire.android-vv0.0.1(4297836)(4297836)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(4297836)(4297836)-dev-release.apk)

PR #1021, opened 2022-09-22
chore: Add GetIncomingCallsOnceUseCase

borichellow:
# What's new in this PR?

use `GetIncomingCallsOnceUseCase` to get incomingCalls only once and do not observe it in case of PushNotification

### DIFF
--- a/app/src/main/kotlin/com/wire/android/notification/WireNotificationManager.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/WireNotificationManager.kt
@@ -148,8 +148,8 @@ class WireNotificationManager @Inject constructor(
                 appLogger.d("$tag; Getting incoming calls for")
                 coreLogic.getSessionScope(userId)
                     .calls
-                    .getIncomingCalls()
-                    .first().also {
+                    .getIncomingCallsOnce()
+                    .also {
                         appLogger.d("$tag; Incoming calls result list size = ${it.size}")
                     }
             }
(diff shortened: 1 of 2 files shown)


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1021/1/) **failed**.

app/src/main/kotlin/com/wire/android/notification/WireNotificationManager.kt
```diff
@@ -148,8 +148,8 @@ class WireNotificationManager @Inject constructor(
                 appLogger.d("$tag; Getting incoming calls for")
                 coreLogic.getSessionScope(userId)
                     .calls
-                    .getIncomingCalls()
-                    .first().also {
+                    .getIncomingCallsOnce()
+                    .also {
```
ohassine, line 152: But `.first()` isn't designed to do that ? Get only one item, which is the first received, and cancel the flow

borichellow: this PR i kinda "testing", something like "lets try". We have some problems on notifications and looks like it stucks in this `fetchAndShowCallNotificationsOnce`, so I decided to try that :)

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1021/2/) **failed**.

PR #1064, opened 2022-10-04
feat: search conversations (AR-2538)

trOnk12:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-2538" title="AR-2538" target="_blank">StoryAR-2538</a>  Filter conversation list
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- We did not support searching the conversations. The first iteration is just a filter through our existing conversation

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

https://user-images.githubusercontent.com/15973138/194102959-522e0c9e-2bd7-4c2d-b716-73bab81736b6.mp4

----

----

----
##### References


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/1/) **failed**.

AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/1/) **failed**.

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
```diff
@@ -45,185 +44,211 @@ import com.wire.android.ui.common.dimensions
...
-            CollapsingTopBarScaffold(
-                snapOnFling = false,
-                keepElevationWhenCollapsed = true,
-                topBarHeader = homeTopBar,
-                snackbarHost = {
-                    SwipeDismissSnackbarHost(
-                        hostState = snackbarHostState,
-                        modifier = Modifier.fillMaxWidth()
+    with(homeScreenState) {
+        with(homeState) {
```
saleniuk: Does it make sense to have two `with`s one right after another?

app/src/main/kotlin/com/wire/android/navigation/HomeNavigation.kt
```diff
@@ -70,23 +46,24 @@ enum class HomeNavigationItem(
...
-    val content: (HomeUIState) -> (@Composable (NavBackStackEntry) -> Unit)
+    val content: (HomeScreenState) -> (@Composable (NavBackStackEntry) -> Unit)
 ) {
     Conversations(
         route = HomeDestinationsRoutes.conversations,
         title = R.string.conversations_screen_title,
         tabName = R.string.conversations_all_tab_title,
         icon = R.drawable.ic_conversation,
-        isSearchable = false, // TODO
+        isSearchable = true, // TODO
```
saleniuk: We can remove this TODO finally 😄

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
```diff
@@ -45,185 +44,211 @@ import com.wire.android.ui.common.dimensions
...
+                                },
+                                topBarCollapsing = {
+                                    if (currentNavigationItem.isSearchable)
+                                        SearchTopBar(
+                                            isSearchActive = searchBarState.isSearchActive,
+                                            searchBarHint = stringResource(
+                                                R.string.search_bar_hint,
+                                                stringResource(id = title).lowercase()
+                                            ),
+                                            searchQuery = searchBarState.searchQuery,
```
saleniuk: Shouldn't we open another screen dedicated for the search and not search/filter on this screen? I mean, I thought the idea was to make a dummy clickable search field here which just opens another screen, did the plan change?

app/src/main/kotlin/com/wire/android/ui/home/HomeScreenState.kt
```diff
@@ -67,34 +70,48 @@ class HomeUIState(
         homeBottomSheetContent = content
     }
 
+    fun closeDrawer() {
+        coroutineScope.launch {
```
saleniuk: Should the state have the coroutine scope? Should the state execute such actions or just keep the current state? Maybe actions like this one or `navigateTo` should be handled by the ViewModel

app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
```diff
@@ -116,24 +99,37 @@ class HomeViewModel @Inject constructor(
...
+//
+//    fun searchConversation(searchQuery: TextFieldValue) {
+//        val textQueryChanged = searchQueryTextFieldFlow.value.text != searchQuery.text
+//        // we set the state with a searchQuery, immediately to update the UI first
+//        viewModelScope.launch {
+//            searchQueryTextFieldFlow.emit(searchQuery)
+//
+//            if (textQueryChanged) mutableSearchQueryFlow.emit(searchQuery.text)
+//        }
+//    }
```
saleniuk: Should this be commented? If so then add a few words explaining why it is this way for others not to remove it accidentally 😉

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListState.kt
```diff
@@ -5,12 +5,13 @@ import com.wire.android.ui.home.conversationslist.model.ConversationItem
 
 data class ConversationListState(
     val conversations: Map<ConversationFolder, List<ConversationItem>> = emptyMap(),
+    val hasNoConversations: Boolean = false,
+    val conversationSearchResult: Map<ConversationFolder, List<ConversationItem>> = emptyMap(),
```
saleniuk, line 9: Can't we just reuse the `conversations`? Do we need to pass two independent large maps when we search?

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
```diff
@@ -30,153 +39,242 @@ import com.wire.android.ui.home.conversationslist.model.GroupDialogState
...
+    LaunchedEffect(requestInProgress) {
+        if (!requestInProgress) {
+            leaveGroupDialogState.dismiss()
+            deleteGroupDialogState.dismiss()
+            blockUserDialogState.dismiss()
+            unblockUserDialogState.dismiss()
+        }
+    }
+
+    return remember(initialConversationItemType) {
```
saleniuk: So this will change only if `initialConversationItemType` changes - is that how it is supposed to work?
I'm worried that creating such huge "remember states" is more error-prone (you can easily miss some cases when it should be updated and when remembered) and maybe it's better to combine states using `derivedState` - this way we get less recompositions.

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/all/AllConversationScreen.kt
```diff
@@ -147,14 +147,15 @@ fun ConversationListEmptyStateScreen() {
...
+//@Composable
+//fun ComposablePreview() {
+//    AllConversationScreen(mapOf(), {}, {}, {}, {}, {}, false, emptySearchResult)
+//}
+//
+//@Preview
+//@Composable
+//fun ConversationListEmptyStateScreenPreview() {
+//    AllConversationScreen(mapOf(), {}, {}, {}, {}, {}, true, emptySearchResult)
+//}
```
saleniuk: Please provide some previews, it's easier and faster to understand the composable if you can see how it looks like

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/search/SearchConversationScreen.kt
```diff
@@ -0,0 +1,129 @@
...
+    onOpenConversationNotificationsSettings: (ConversationItem) -> Unit,
+    onJoinCall: (ConversationId) -> Unit
+) {
+    Box(Modifier.fillMaxSize()) {
+        if (conversationSearchResult.values.isEmpty()) {
+            EmptySearchResult(onOpenNewConversation)
+        } else {
+            val context = LocalContext.current
+
+            LazyColumn(
```
saleniuk: I think this code of `LazyColumn` and its items should be extracted and reused on multiple screens instead of duplicating the code

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
trOnk12: I see no reason not to do it :thinking

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
trOnk12: There is another screen being rendered if you mean NavigationItem i do not think it makes sense in this case because we would need to fetch all the conversation inside the viewmodel for that navigationitem here we can reuse what we already have, however on next iteration we need to perform sql query and it will make sense to use a navigationitem then I think

app/src/main/kotlin/com/wire/android/ui/home/HomeScreenState.kt
trOnk12: I though we agreed not to do UI stuff inside ViewModel also remberCoroutineScope creates a scope associated with Composable which is exactly what we want I guess. The way I see state is that it is a object handling UI related stuff

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
trOnk12: Hmmm I honestly do not think this is Error prone if we know what we are doing :p and yes it should be only remembered then also I am not sure how derived state would be better here :thinking

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListState.kt
trOnk12: We need to hold somewhere alll the conversations to perform a filter on the conversation in this iteration, so I guess there is not really an alternative to do it ?

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
saleniuk: Both objects are mixed because both are accessed via `this` and you don't know which method and field come from which one, for instance if you use `hashCode()` then you need to remember that you're inside the inner `with` and if you want to differentiate them by using `this@with.hashCode()` then you get "There is more than one label with such a name in this scope", so yeah, it works but it's confusing and can be error-prone.
We can leave it as it is now, I just prefer the code to be more readable and easier to understand. 😉

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
saleniuk: We can create a cache to reuse the data on both screens, but ok, it's fine now.

app/src/main/kotlin/com/wire/android/ui/home/HomeScreenState.kt
saleniuk: Okay, but is the navigation only an UI action? I just don't like the idea of state handling some actions where it should be a data class only containing the data.

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationRouter.kt
saleniuk: Yep, "if we know what we are doing", you can also implement way more complicated stuff and say "I know what I'm doing" but it doesn't make everyone else also know what's happening there and why 😉 
`derivedState` recomposes only if the calculated outcome value is changed, here it happens when the input `initialConversationItemType` is changed, so don't we want it to be updated when `requestInProgress` is changed for instance?

app/src/main/kotlin/com/wire/android/ui/home/HomeScreenState.kt
trOnk12: closing a drawer is not a navigation, and why are you not liking it ? 🤔 

we have a Composable that has a state so indeed it has some data, but it also can do UI things like closing a drawer, closing a bottom sheet and stuff, I mean we need to call somewhere rememberBottomSheetState for example inside our Composable and why not wrap the thing that rememberBottomSheetState returns and put it into a "global" state object that manipulates this UI object, maybe the name is not right for a class doing it, but I see not cons to not do that

app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
trOnk12: I agree, I used only the first with

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/2/) **failed**.

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListState.kt
trOnk12: @saleniuk  In the end I decided to keep it the way it is here, because when I was navigating back from "empty" search result very quickly there was a empty conversation screen for a second on the HomeScreen where you it tells you to click on "New" button to creata a conversation, this will be iterated pretty soon with a SQL query so there will be a need for a variable holding the search results I think

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/3/) **failed**.

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/4/) **failed**.

saleniuk (approved):

AndroidBob:
Build [5](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/5/) **failed**.

AndroidBob:
Build [6](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/6/) **failed**.

AndroidBob:
Build [7](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/7/) **succeeded**.

- [com.wire.android-vv0.0.1(4677497)(4677497)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(4677497)(4677497)-dev-release.apk)

AndroidBob:
Build [8](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/8/) **failed**.

AndroidBob:
Build [10](https://10.10.124.134/job/wire-android-reloaded/job/PR-1064/10/) **succeeded**.

- [com.wire.android-vv0.0.1(4677996)(4677996)-dev-release.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/dev/release/com.wire.android-vv0.0.1(4677996)(4677996)-dev-release.apk)

PR #1437, opened 2023-02-08
feat: create a guest link - Part 1 : Move guest access config to separate screen (AR-2706)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-2706" title="AR-2706" target="_blank">StoryAR-2706</a>  Admin: create a guest link
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Changing guest access config from conversation details should be done in separate screen.

### Solutions

I created a new screen and moved all the logic the new viewModel

#### Test Coverage (Optional)

#### How to Test

- Go to conversation details of a group
- Tap on Guest item

### Attachments (Optional)

https://user-images.githubusercontent.com/18124919/217604564-0e2d42ac-5cce-45cb-833f-671cd05937ab.mp4

----

----

----
##### References


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1437/1/) **failed**.

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-1437/4/) **succeeded**.

- [com.wire.android-vv4.0.0(5691229)(5691229)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.0(5691229)(5691229)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessViewModel.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    val conversationId: QualifiedID = qualifiedIdMapper.fromStringToQualifiedID(
+        checkNotNull(savedStateHandle.get<String>(EXTRA_CONVERSATION_ID)) {
+            "No conversationId was provided via savedStateHandle to EditGuestAccessViewModel"
+        }
+    )
+    private val isGuestAccessAllowed: Boolean = savedStateHandle.get<String>(EXTRA_EDIT_GUEST_ACCESS_IS_GUEST_ACCESS_ALLOWED)!!.toBoolean()
+    private val isServicesAllowed: Boolean = savedStateHandle.get<String>(EXTRA_EDIT_GUEST_ACCESS_IS_SERVICES_ALLOWED)!!.toBoolean()
+    private val isUpdatingGuestAccessAllowed: Boolean =
+        savedStateHandle.get<String>(EXTRA_EDIT_GUEST_ACCESS_IS_UPDATING_GUEST_ACCESS_ALLOWED)!!.toBoolean()
```
saleniuk: Can't we just collect the Flows in this ViewModel the same way like it's done in `GroupConversationDetailsViewModel.observeConversationDetails()`? I think it would be better than passing these value as navigation params

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -342,6 +344,35 @@ enum class NavigationItem(
...
+                "&$EXTRA_EDIT_GUEST_ACCESS_IS_UPDATING_GUEST_ACCESS_ALLOWED={$EXTRA_EDIT_GUEST_ACCESS_IS_UPDATING_GUEST_ACCESS_ALLOWED}",
+        content = { EditGuestAccessScreen() },
+        animationConfig = NavigationAnimationConfig.CustomAnimation(smoothSlideInFromRight(), smoothSlideOutFromLeft())
+    ) {
+        override fun getRouteWithArgs(arguments: List<Any>): String {
+            val conversationIdString: String = arguments.filterIsInstance<ConversationId>().firstOrNull()?.toString()
+                ?: "{$EXTRA_CONVERSATION_ID}"
+
+            val isGuestAccessAllowed: String = arguments.filterIsInstance<Boolean>().firstOrNull()?.toString()
+                ?: "{$EXTRA_EDIT_GUEST_ACCESS_IS_GUEST_ACCESS_ALLOWED}"
```
saleniuk: What will happen if the parameter won't be passed and `"{$EXTRA_EDIT_GUEST_ACCESS_IS_GUEST_ACCESS_ALLOWED}"` will be returned? 
In the ViewModel it looks like this:
```
savedStateHandle.get<String>(EXTRA_EDIT_GUEST_ACCESS_IS_GUEST_ACCESS_ALLOWED)!!.toBoolean()
```
Won't that crash then because of the mapping to Boolean failure?

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
```diff
@@ -342,6 +344,35 @@ enum class NavigationItem(
...
+
+            val isServicesAllowed: String = arguments.filterIsInstance<Boolean>().firstOrNull()?.toString()
```
saleniuk: Probably both will get the same value because both `isGuestAccessAllowed` and `isServicesAllowed` look for the Boolean and use `firstOrNull()` so both will get the same first found Boolean param. Doing it this way, one by one, is okay when each param is of a different type, otherwise for each next param of the same type we would need to use `getOrNull(X)` and always keep track of the params order.

It's better and easier to create a dedicated data class for these params:
```
data class EditGuestAccessParams(
    val conversationIdString: String,
    val isGuestAccessAllowed: String,
    val isServicesAllowed: String,
    val isUpdatingGuestAccessAllowed
)
```
pass the whole model as an argument and retrieve:
```
val editGuestAccessParams: EditGuestAccessParams? = arguments.filterIsInstance<EditGuestAccessParams>().firstOrNull()
```

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessState.kt
```diff
@@ -0,0 +1,36 @@
...
+package com.wire.android.ui.home.conversations.details.editguestaccess
+
+import com.wire.kalium.logic.CoreFailure
+
+data class EditGuestAccessState(
+    val isGuestAccessAllowed: Boolean = false,
+    val isUpdatingGuestAccessAllowed: Boolean = false,
+    val changeGuestOptionConfirmationRequired: Boolean = false,
+    val isUpdating: Boolean = false,
+    val error: Error = Error.None,
```
saleniuk: Do we handle this error somehow?

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
ohassine: Oh yes, it should  be `lastOrNull()` for service. That will work I would say

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessViewModel.kt
ohassine: yes we can, but why fetching the same value twice from database ? we already get it from previous screen

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
ohassine: yes correct it will crash if there is no guest value

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessState.kt
ohassine: Yes not used

AndroidBob:
Build [6](https://10.10.124.134/job/wire-android-reloaded/job/PR-1437/6/) **succeeded**.

- [com.wire.android-vv4.0.0(5750113)(5750113)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.0(5750113)(5750113)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessViewModel.kt
saleniuk: Because it's less complicated and less error-prone, as you can see that you already had troubles with passing and retrieving navigation params 😄 Also, it wouldn't require to pass all these params when opening this screen - it can retrieve these values itself and we don't have to worry about it 😉 
We could extract the whole `observeConversationDetails` to a separate use case and make use of `.shareIn()` with replay so it works like a cache and don't have to query the DB every time, when passing `SharingStarted.WhileSubscribed()` we can set a timeout so it would still replay the same value for the next screen's ViewModel

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
saleniuk: So it requires these params, we can ensure that we have them here, by either `checkNotNull` or `first` instead of `firstOrNull`, no need to try to open the screen if we know that it won't be successful 🤔

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
saleniuk: Nope, because you actually pass three `Boolean` type params, so now it clashes with the next one - both `isServicesAllowed` and `isUpdatingGuestAccessAllowed` get the same `lastOrNull` value.
I still suggest that if we have to pass all these params then we should use a dedicated data class for that.

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessState.kt
saleniuk: Maybe we should add a TODO then, to not forget about implementing that? 😄

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessViewModel.kt
ohassine: Okay..But the main point of passing data via navigation params is to avoid flickering, even with your approach the screen will flicker because screen state is initialised with default values then updated.

Also I don't see any raison to observe something not observable, `isGuestAccessAllowed` and `isServicesAllowed` do not change in real time if there is a change in a different client.

app/src/main/kotlin/com/wire/android/navigation/NavigationItem.kt
ohassine: yeah got your point, will use a data class for it then

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessState.kt
ohassine: It's to remove, I handled error in a different way

AndroidBob:
Build [8](https://10.10.124.134/job/wire-android-reloaded/job/PR-1437/8/) **succeeded**.

- [com.wire.android-vv4.0.0(5758082)(5758082)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.0(5758082)(5758082)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessViewModel.kt
saleniuk: If we use it the right way with the replay parameter set, then it'll get the proper value as the initial one already, so there should be no flickering. 😎 I believe we already had this discussion on collective and opted out of passing more data as navigation params because flickering was fixed by using right dispatchers. 😉 
And you're right, these value cannot be changed in the meantime by something else, but we still however for some reason observe them in `GroupConversationDetailsViewModel`, we could also pass data to this screen as navigation params, so why don't we do that? 😄 
It makes the code and already complex navigation more complicated in my opinion and makes this new screen dependent on the parent screen that opens it. Only crucial data (like IDs) that's used to retrieve all other data needed on a particular screen should be passed as a navigation params, you can read about it here: https://developer.android.com/jetpack/compose/navigation#retrieving-complex-data

ohassine: As discussed in 1:1, I kept passing values  over nav parms and I am now observing the conversation details.

AndroidBob:
Build [10](https://10.10.124.134/job/wire-android-reloaded/job/PR-1437/10/) **succeeded**.

- [com.wire.android-vv4.0.0(5769122)(5769122)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.0(5769122)(5769122)-staging-compat.apk)

saleniuk (approved):

PR #1450, opened 2023-02-14
fix: Missed notifications on spamming (AR-2867)

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-2867" title="AR-2867" target="_blank">BugAR-2867</a>  Notifications don't show all notifications
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

When we receive new messages really quick (spamming with 2-3 messages per second), some notifications are missed.

### Causes (Optional)

When the push comes we launch the coroutine job for syncing and observing the notifications and calls + there is an algorithm to check if such a job is already running and do not start a new if yes. 
The problem was in call observation: because of some limitations in AVS  we need some time to get an Event about a call from our BE, then go to AVS and ask if such a call is active yet. Because of it we have a workaround that observes calls for 8 seconds. 

So when somebody spams us, new pushes come, but the new sync job is not started as the app thinks that it's already running, cause the 8-seconds workaround is a part of that job. So what do we get: job is running, but the sync is not => losing the notifications. 

### Solutions

Extract 8-seconds workaround for calls from the sync-job to the separate job.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/notification/WireNotificationManager.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/WireNotificationManager.kt
@@ -74,9 +74,29 @@ class WireNotificationManager @Inject constructor(
 
     private val scope = CoroutineScope(SupervisorJob() + dispatcherProvider.default())
     private val fetchOnceMutex = Mutex()
-    private val fetchOnceJobs = hashMapOf<String, Job>()
+    private val fetchOnceJobs = hashMapOf<UserId, Job>()
+    private val fetchCallsOnceJobs = hashMapOf<UserId, Job>()
     private val observingWhileRunningJobs = hashMapOf<UserId, ObservingJobs>()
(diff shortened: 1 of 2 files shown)


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1450/1/) **failed**.

vitorhugods (approved):
🕵🏼 NICE

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1450/2/) **succeeded**.

- [com.wire.android-vv4.0.1(5741365)(5741365)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.1(5741365)(5741365)-staging-compat.apk)

PR #1534, opened 2023-03-07
fix: German translations for adding, removing and leaving groups [AR-3190][AR-3192]

gongracr:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3190" title="AR-3190" target="_blank">BugAR-3190</a>  Removing, adding a user, and self user leaving group have wrong translations in german
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Translations were broken for German locale. They were not handling correctly the cases where self user was removing others, or being removed by other. Same for being added to groups. Therefore, a deeper refactor was needed to be written mapping each case to the full sentence as the German grammar wasn't allowing partial translations of smaller chunks of the phrase.

#### How to Test

Add, remove, users to groups, with different accounts, and check the right strings are shown in English and German.

----
##### References


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1534/1/) **failed**.

app/src/main/kotlin/com/wire/android/mapper/MessagePreviewContentMapper.kt
```diff
@@ -106,101 +111,117 @@ fun MessagePreview.uiLastMessageContent(): UILastMessageContent {
...
-                                    R.plurals.last_message_people,
-                                    membersAddedContent.otherUserIdList.size,
-                                    membersAddedContent.otherUserIdList.size
-                                )
-                            )
-                            .plus(UIText.StringResource(R.string.last_message_to_conversation))
-                    )
+                    val previewMessageContent = when {
+                        // This case would never be applicable. If self added self, this will be a MemberJoined
+                        isSelfMessage && isSelfAdded -> {
```
saleniuk, line 155: `isSelfMessage` means that I created it, `isSelfAdded` means that I'm the one that's being added, but what if there are multiple people added by this single system message, not only me? 🤔

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1534/2/) **succeeded**.

- [com.wire.android-vv4.0.2(5930942)(5930942)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.2(5930942)(5930942)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/mapper/MessagePreviewContentMapper.kt
gongracr: As the comment reads, this case is not really reachable, because self user can't be the originator/trigger of the adding message and still add himself. But Kubaz still wanted to add it for security.
Also maybe I can rename those fields to something clearer, like `isSelfUserOriginated` and `isSelfUserAdded` or something similar.

saleniuk: What about joining a group by an invitation link? Isn't that considered as a user adding himself? If not then probably it's not possible, it's just that the proto model gives the possibility for the author to add himself and some others 😄

gongracr: Nope, there is an extra event called `WithUser.MemberJoined` which will be delegated to a different mapped class. In any case this was just a way to "catch" this undesired scenario and display it in a slightly more logic preview message.

saleniuk (approved):

PR #1550, opened 2023-03-15
feat: self deletion of assets and text messages

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Add support to display the time left for a self-deleting message

### Solutions

- I implemented a self deletion timer, that calculates the time left as well as formats the String that informs the user, how much time is actually left. The timer is launched whenever the user sees the message , that is when the messages is composed, the timer keeps ticking due the recursive way of how LaunchedEffect works, the key is the timeLeft, timeLeft changes after some delay that is calculated through interval(), after is changes LaunchedEffect is retriggered

### Attachments (Optional)
Screenshot_20230321_165543

----

----

----
##### References


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/1/) **failed**.

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/2/) **failed**.

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/3/) **succeeded**.

- [com.wire.android-vv4.1.1(6068849)(6068849)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6068849)(6068849)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/mapper/MessageMapper.kt
```diff
@@ -155,7 +171,10 @@ class MessageMapper @Inject constructor(
             )
 
         message.status == Message.Status.FAILED -> MessageStatus.SendFailure
-        message.status == Message.Status.FAILED_REMOTELY -> MessageStatus.SendRemotelyFailure(message.conversationId.domain)
+        message.status == Message.Status.FAILED_REMOTELY -> MessageStatus.SendRemotelyFailure(
+            message.conversationId.domain
+        )
```
gongracr: why this strange formatting?

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
gongracr, line 1: Shouldn't we name this file `SelfDeletionTimer` for coherence?

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
```diff
@@ -0,0 +1,266 @@
...
+            }
+
+            fun alphaBackgroundColor(): Float {
+                val totalTimeLeftRatio = timeLeft / expireAfter
+
+                return if (totalTimeLeftRatio >= TIME_LEFT_RATIO_BOUNDARY_FOR_ALMOST_TIME_LEFT_ELAPSED_ALPHA) {
+                    PAST_RATIO_BOUNDARY_FOR_ALMOST_TIME_LEFT_ALPHA_VALUE
+                } else {
+                    BEFORE_RATIO_BOUNDARY_FOR_ALMOST_TIME_LEFT_ALPHA_VALUE
+                }
```
gongracr: Shouldn't be this the opposite? if we are above 75% of the remaining time, then the alpha should be 1.0, right ?
So either change the alpha value of the constants or the sign of the comparison.

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
```diff
@@ -0,0 +1,266 @@
...
+                expirableTimer = expirableTimer,
+                onSelfDeletingMessageRead = { onStartMessageSelfDeletion(message) },
+                downloadStatus = messageContent.downloadStatus
+            )
+
+            is UIMessageContent.ImageMessage -> startAssetDeletion(
+                expirableTimer = expirableTimer,
+                onSelfDeletingMessageRead = { onStartMessageSelfDeletion(message) },
+                downloadStatus = messageContent.downloadStatus
+            )
```
gongracr, line 270: Can we just group all of these together?
```suggestion
            is UIMessageContent.AssetMessage, is UIMessageContent.ImageMessage, is UIMessageContent.AudioAssetMessage -> startAssetDeletion(
                expirableTimer = expirableTimer,
                onSelfDeletingMessageRead = { onStartMessageSelfDeletion(message) },
                downloadStatus = messageContent.downloadStatus
            )
```

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
```diff
@@ -0,0 +1,266 @@
...
+private fun startAssetDeletion(
+    expirableTimer: SelfDeletionTimer.SelfDeletionTimerState.Expirable,
+    onSelfDeletingMessageRead: () -> Unit,
+    downloadStatus: Message.DownloadStatus
+) {
+    LaunchedEffect(downloadStatus) {
+        if (downloadStatus == Message.DownloadStatus.SAVED_EXTERNALLY || downloadStatus == Message.DownloadStatus.SAVED_INTERNALLY) {
+            onSelfDeletingMessageRead()
+        }
+    }
```
gongracr, line 301: Isn't this condition the same than the one below? why not moving the  `onSelfDeletingMessageRead()` method to the LaunchedEffect below and just have one?

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
```diff
@@ -86,106 +93,163 @@ fun MessageItem(
...
-private fun Modifier.customizeMessageBackground(
-    message: UIMessage
-) = run {
-    if (message.sendingFailed || message.receivingFailed) {
-        background(MaterialTheme.wireColorScheme.messageErrorBackgroundColor)
-    } else this
+fun MessageExpireLabel(messageContent: UIMessageContent?, timeLeft: String) {
+    when (messageContent) {
+        is UIMessageContent.TextMessage -> {
+            StatusBox(statusText = "Self-deleting message •  $timeLeft")
```
gongracr: Let's localise this text pls.

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/ConversationMessagesViewModel.kt
```diff
@@ -105,6 +106,15 @@ class ConversationMessagesViewModel @Inject constructor(
...
         observeAudioPlayerState()
+        observeSelfDeletingMessages()
+    }
+
+    private fun observeSelfDeletingMessages() {
+        viewModelScope.launch {
+//            selfDeletingMessagesObserver.observableSelfDeletingMessageState.collect {
+//                println("state :$it")
+//            }
+        }
```
gongracr: Let's just delete this chunk if this is only for testing purposes.

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/ConversationMessagesViewModel.kt
```diff
@@ -117,6 +127,10 @@ class ConversationMessagesViewModel @Inject constructor(
         }
     }
 
+    fun observeSelfDeletingMessageProgress(conversationId: QualifiedID, messageId: String) {
+//        selfDeletingMessagesObserver.observeSelfDeletingMessageProgress(conversationId, messageId)
+    }
+
```
gongracr: Same here

app/src/test/kotlin/com/wire/android/mapper/MessageMapperTest.kt
```diff
@@ -92,8 +91,16 @@ class MessageMapperTest {
...
+        val message3 = arrangement.testMessage(
+            senderUserId = userId1,
+            editStatus = Message.EditStatus.Edited(now),
+            date = now
+        )
+        val message4 = arrangement.testMessage(
+            senderUserId = userId1,
+            visibility = Message.Visibility.DELETED,
+            date = now
+        )
```
gongracr: Are you using a different formatting style than the rest of the project? Because this shouldn't truncate by default 🤔

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
```diff
@@ -86,106 +93,163 @@ fun MessageItem(
...
+
+        is UIMessageContent.ImageMessage -> {
+            StatusBox(
+                statusText = "Self-deleting message" +
+                        " ${if (messageContent.downloadStatus == Message.DownloadStatus.SAVED_INTERNALLY) "•$timeLeft" else ""}"
+            )
+        }
+
+        else -> {}
+    }
```
gongracr: Maybe we could refactor all these items into another sealed class like `UIAssetMessageContent` so that they share at least the `assetId`, `downloadStatus` and `uploadStatus`. 
That way we can group them into easier `when` clauses.

Also please take care of localising the texts.

gongracr (changes requested):
Awesome job as usual 🦾 ! I just added some suggestions to simplify some code and a question to clear about the logic of showing the alpha of a self deleted message.

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -199,7 +202,10 @@ class SessionModule {
 
     @ViewModelScoped
     @Provides
-    fun provideCurrentAccountUserDataStore(@CurrentAccount currentAccount: UserId, userDataStoreProvider: UserDataStoreProvider) =
+    fun provideCurrentAccountUserDataStore(
+        @CurrentAccount currentAccount: UserId,
+        userDataStoreProvider: UserDataStoreProvider
+    ) =
         userDataStoreProvider.getOrCreate(currentAccount)
```
vitorhugods:
```suggestion
    fun provideCurrentAccountUserDataStore(
        @CurrentAccount currentAccount: UserId,
        userDataStoreProvider: UserDataStoreProvider
    ) = userDataStoreProvider.getOrCreate(currentAccount)
```

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
```diff
@@ -208,22 +214,34 @@ class SessionModule {
 class ConnectionModule {
     @ViewModelScoped
     @Provides
-    fun provideSendConnectionRequestUseCase(@KaliumCoreLogic coreLogic: CoreLogic, @CurrentAccount currentAccount: UserId) =
+    fun provideSendConnectionRequestUseCase(
+        @KaliumCoreLogic coreLogic: CoreLogic,
+        @CurrentAccount currentAccount: UserId
+    ) =
         coreLogic.getSessionScope(currentAccount).connection.sendConnectionRequest
```
vitorhugods:
```suggestion
    fun provideSendConnectionRequestUseCase(
        @KaliumCoreLogic coreLogic: CoreLogic,
        @CurrentAccount currentAccount: UserId
    ) = coreLogic.getSessionScope(currentAccount).connection.sendConnectionRequest
```

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
vitorhugods, line 1: All this formatting in general? 🤔

`Auto-correct with detekt` can be used, but please take time to clean up because it also left some mess behind 😅 

Detekt won't complain if it's formatted like this:

```kotlin
    fun provideCurrentAccountUserDataStore(
        @CurrentAccount currentAccount: UserId,
        userDataStoreProvider: UserDataStoreProvider
    ) = userDataStoreProvider.getOrCreate(currentAccount)
```

It left the `= userDataStoreProvider.getOrCreate(currentAccount)` on the next line because it was like this before. But it's fine to keep in the same line.

app/src/main/kotlin/com/wire/android/mapper/MessageMapper.kt
```diff
@@ -114,12 +114,28 @@ class MessageMapper @Inject constructor(
...
+                    it.expireAfter,
+                    it.selfDeletionStatus
+                )
+            } ?: ExpirationStatus.NotExpirable
+        } else {
+            ExpirationStatus.NotExpirable
+        }
+
+        return expirationStatus
+    }
```
vitorhugods: Nitpick. A bit more readable, perhaps?

```suggestion
    private fun provideExpirationData(message: Message.Standalone): ExpirationStatus {
        val expirationData = (message as? Message.Regular)?.expirationData
        return if (expirationData == null) {
            ExpirationStatus.NotExpirable
        } else {
            ExpirationStatus.Expirable(
                expireAfter = expirationData.expireAfter,
                selfDeletionStatus = expirationData.selfDeletionStatus
            )
        }
    }
```

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
```diff
@@ -214,6 +214,7 @@ fun ConversationScreen(
         conversationMessages = conversationMessagesViewModel.infoMessage,
         conversationMessagesViewModel = conversationMessagesViewModel,
         onPingClicked = messageComposerViewModel::sendPing,
+        onSelfDeletingMessageRead = messageComposerViewModel::startSelfDeletion,
```
vitorhugods, line 213: Why `onSelfDeletingMessageRead`?

Couldn't it be just `onMessageRead` and let the `ViewModel` do its job?

Perhaps this could also be useful for literally marking messages as read as the user scrolls up/down.

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
```diff
@@ -0,0 +1,266 @@
...
+
+        class Expirable(
+            private val resources: Resources,
+            timeLeft: Duration,
+            private val expireAfter: Duration
+        ) : SelfDeletionTimerState() {
+            companion object {
+                private const val TIME_LEFT_RATIO_BOUNDARY_FOR_ALMOST_TIME_LEFT_ELAPSED_ALPHA = 0.75
+                private const val PAST_RATIO_BOUNDARY_FOR_ALMOST_TIME_LEFT_ALPHA_VALUE = 0F
+                private const val BEFORE_RATIO_BOUNDARY_FOR_ALMOST_TIME_LEFT_ALPHA_VALUE = 1F
```
vitorhugods: What are these supposed to mean exactly?

What is "almost time left"?

Care to please add some docs?

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
```diff
@@ -0,0 +1,266 @@
...
+                    resources.getQuantityString(
+                        R.plurals.seconds_left,
+                        timeLeft.inWholeSeconds.toInt(),
+                        timeLeft.inWholeSeconds.toInt()
+                    )
+
+                else -> throw IllegalStateException("Not possible state for a time left label")
+            }
+
+            fun updateInterval(): Duration {
```
vitorhugods, line 164:
```suggestion
            /**
             * Calculates when the next update should be done based on the [timeLeft]
             * For example, it's unnecessary to update on each second when the timer
             * is still marking 4 weeks left.
             * However, when there's less than a minute left, the timer should be
             * updated every second.
             * @return how long until the next timer update.
             */
            fun updateInterval(): Duration {
```

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
```diff
@@ -71,9 +76,11 @@ import com.wire.android.ui.home.conversations.model.messagetypes.image.ImageMess
 import com.wire.android.ui.theme.wireColorScheme
 import com.wire.android.ui.theme.wireTypography
 import com.wire.android.util.CustomTabsHelper
+import com.wire.kalium.logic.data.message.Message
 import com.wire.kalium.logic.data.user.UserId
 
 // TODO: a definite candidate for a refactor and cleanup
+@Suppress("ComplexMethod")
```
vitorhugods, line 83: Instead of suppressing the complexity, can't it be split into smaller composable functions?

It's always a pain to read diffs of big composable functions. Especially when a new layout is introduced and the whole indentation changes.

Perhaps the whole timer thing could be extracted to its own composable instead of living in the `MessageItem` directly.

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
vitorhugods: And have it accept formatting parameters, so we can plug in the `timeLeft` instead of concatenating the strings manually.

There will always be some language where it makes sense for the duration to be in different places of the string.

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
```diff
@@ -86,106 +93,163 @@ fun MessageItem(
...
 }
 
 @Composable
-private fun Modifier.customizeMessageBackground(
-    message: UIMessage
-) = run {
-    if (message.sendingFailed || message.receivingFailed) {
-        background(MaterialTheme.wireColorScheme.messageErrorBackgroundColor)
-    } else this
+fun MessageExpireLabel(messageContent: UIMessageContent?, timeLeft: String) {
```
vitorhugods, line 224: Maybe move expiration stuff to its own file?

app/src/test/kotlin/com/wire/android/mapper/MessageMapperTest.kt
vitorhugods: Maybe missing the detekt configuration file?

image

Kalium's [Readme](https://github.com/wireapp/kalium/tree/f3aea16e5f300db93c57d32cfb8f6c172339b516#detekt-rules)

kalium
vitorhugods, line 1: When creating the Epic's PR in Reloaded, make sure to add the Kalium's counterpart as dependency:

- https://github.com/wireapp/kalium/pull/1582

As this change in reference would update Reloaded to use a not-yet-merged version of Kalium.

vitorhugods (changes requested):
Some opportunity for small improvements here and there, and missing localis(z)ation.

But it seems nice so far.

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
trOnk12: I think it makes more sense to keep it the way it is, for now we are calling onSelfDeletingMessageRead when we see message for the first, the same goes for marking a message read so it would make sense here, but it could change in the future also for asset we only start self deletion when the user downloads the asset, so already here we would have problems

app/src/main/kotlin/com/wire/android/mapper/MessageMapper.kt
trOnk12: I did not know as? exists, pretty nice thingy, thanks for the tip

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
trOnk12: I think MessageExpiration suits better 🤔 as the timer is really happening inside MessageItem

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
trOnk12: no, this is correct if we are past the boundary for almost time left we give it alpha 0 otherwise 1

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
trOnk12: not possible unfortunetly :S needs a refactor in order to do this, I am going to leave as it is now, from what I remember Kubaz is going to refactor UiMessage ? If he does not lniclude this in his part I am going to submit a PR later on for this

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
trOnk12: I will take care of it in a refactor of this Composable later on as I think there could be more done to this Composable

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItem.kt
trOnk12: Will do in a seperate PR later on

app/src/test/kotlin/com/wire/android/mapper/MessageMapperTest.kt
trOnk12: fixed

app/src/main/kotlin/com/wire/android/di/CoreLogicModule.kt
trOnk12: weird :S I pulled the develop for this one not sure how this sneaked in

app/src/main/kotlin/com/wire/android/mapper/MessageMapper.kt
trOnk12: no idea :S thanks for the sharp eye

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
trOnk12: it is the same condition indeed, however the other LaunchedEffect takes timeLeft as a key as well, because we want to start the timer when the asset is downloaded, but also keep launching the LaunchedEffect when the timeLeft changes to keep the timer "ticking" so I do not see a way to move it there 🤔

trOnk12: I could move onSelfDeletingMessageRead() to the other one but it would be called each time the timer ticks :S

AndroidBob:
Build [8](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/8/) **succeeded**.

- [com.wire.android-vv4.1.1(6102688)(6102688)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6102688)(6102688)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
gongracr: I think you understood it the other way around:
Captura de pantalla 2023-03-28 a las 11 56 12
It reads, 
>The message background highlight starts fading **in** when 50% of the time that the message is to be displayed has already passed. It reaches its **full color** when 75% of the time has passed and **stays in full color** until the message is deleted.

also if you look at the alpha percentage in that Figma screenshot, it says that at the end, right before deletion, the alpha is 100%

trOnk12: I tested it and it works as it is supposed to :P so I do not think I misunderstood it, however you are right about fading from 50 to 75 % I did miss that, thanks ! :)

trOnk12: I think the confusion is coming from the fact that 100 % alpha is invisible so I would say the design specs suggest wrong values

trOnk12: hmmm ... I am going to take a second look at it, it is working fine but I understand the confusion now

AndroidBob:
Build [9](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/9/) **failed**.

AndroidBob:
Build [11](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/11/) **failed**.

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
vitorhugods: That's fair.

But in any case, the composable knows a bit too much.

It could be `onMessageContententRead` or similar. So it can be a generic lambda used to inform the ViewModel that some message and all of its content (included displaying of images) is _completely_ visible.

Or, the same lambda could take a parameter.

Like:

```
enum class ContentAvailability { 
    /**
     * Content is only half-available. For example, the asset message
     * is visible to the user, but the asset wasn't downloaded yet.
     */
    PARTIAL,

    /**
     * Content is only fully-available to the user. For example, the asset
     * message is visible and downloaded, so the user can see it.
     */
    COMPLETE
}

// For text messages
onMessageContentRead(COMPLETE)

// For assets, check if it's fully visible
onMessageContentRead(if(isAssetDownloaded) COMPLETE else PARTIAL)
```

This way:
- The view worries about displaying and informing the `ViewModel`
- The `ViewModel` can figure out what needs to be done for each kind of message (_e.g._ marking as read, and/or triggering self deletion, etc)

I'm happy for now, but I'd prefer to keep the view as dumb as possible regarding business logic and reduce the number of lambdas to make a simpler interface between V-VM

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
vitorhugods: Any reason for not documenting it?

I really don't think `updateInterval` is a name clear enough to explain its behaviour and what it returns.

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
vitorhugods: I guess I'll never know 😅

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
trOnk12: This class is a mess on it's own we could take your proposition into the refactoring because there are more places where we could extract the logic to viewmodel i feel like this needs a seperate PR with a seperate place for discussion

trOnk12: And deserves a seperate PR with a refactor because again the problem is  not only here:p

vitorhugods (commented):
Just a couple of questions still hanging

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
```diff
@@ -193,6 +193,18 @@ class SelfDeletionTimer(private val context: Context) {
...
+            * if the time elapsed ratio is between 0.50 and 0.75
+            * we want to change the value proportionally to how much
+            * time is left between 0.50 and 0.75, we doing that by dividing
+            * how much time is elapsed since half of the total expire after time by
+            * the "time slice" that fits between 0.5 and 0.75
+            * for example. [expireAfter] = 10 sec, timeElapsed = 6 sec
+            * quarterTimeLeftSlice = 2.5 sec, halfTimeSlice = 5 sec
+            * durationInBetweenHalfTimeAndQuarterSlice = 5 sec - 2.5 sec = 2.5 sec
+            * timeElapsedFromHalfTimeSlice = 6 sec - 5 sec = 1 sec
+            * alpha value is equal to the ratio = 1 / 2.5 = 0.4
```
vitorhugods: Instead of copying `0.5` and `0.75`, please use something like:

```kotlin
            * If the time elapsed ratio is between [LOW_TIME_ELAPSED_RATIO_BOUNDARY_FOR_PROPONTIAL_ALPHA_CHANGE] and [HALF_RATIO_VALUE]
            * we want to change the value proportionally to how much
            * time is left between [LOW_TIME_ELAPSED_RATIO_BOUNDARY_FOR_PROPONTIAL_ALPHA_CHANGE] and [HALF_RATIO_VALUE], we doing that by dividing

```

So if the values change, there's no need to update the docs. Also, it helps to understand what these values are for.

Keep in mind that I just copied/pasted some names in there.

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
trOnk12: the doc is there 🤔

vitorhugods: Indeed. Some weird GitHub behaviour I guess 😄

vitorhugods (approved):
🪄✨

AndroidBob:
Build [13](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/13/) **failed**.

AndroidBob:
Build [14](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/14/) **failed**.

gongracr (approved):
Awesome effort 🍄 💪 thx for the endless changes. Also make sure that this PR gets merged after the Kalium counterparts get merged to develop 🙏

AndroidBob:
Build [15](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/15/) **failed**.

AndroidBob:
Build [18](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/18/) **failed**.

AndroidBob:
Build [19](https://10.10.124.134/job/wire-android-reloaded/job/PR-1550/19/) **succeeded**.

- [com.wire.android-vv4.1.1(6120477)(6120477)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6120477)(6120477)-staging-compat.apk)

PR #1560, opened 2023-03-17
fix: infinite call vibration (AR-3214)

alexandreferris:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3214" title="AR-3214" target="_blank">BugAR-3214</a>  PlayTest 14.03 - Phone keeps vibrating when rejecting an incoming call
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Notification for incoming call would keep an infinite loop of vibration when caller start the call and cancel before the other user even gets the first calling signal.
The only way to stop it was opening the notification tray.

### Solutions

Add a delay of 300ms before cancelling the notification so it has time to properly set the notification vibration+sound before cancelling it (as it was being called around less than 100ms) and thus not doing exactly what it needed to do (cancel the notification vibration+sound).

Because from out POV it was like this:
`Create Notification -> Create Vibration object -> cancel notification -> start vibration -> infinite vibration`
with the delay we are giving a bit of time for the vibration in notification to properly start and then properly be cancelled.

#### How to Test

- Have app in vibration mode
- Have 3 accounts
- Login with acc1 and acc2
- make acc2 active
- call acc1 from acc3 and cancel before acc3 even receives the first call signal
- app should not be stuck in vibration

### DIFF
--- a/app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
@@ -55,6 +55,11 @@ class CallNotificationManager @Inject constructor(private val context: Context) 
 
     fun hideIncomingCallNotification() {
         appLogger.i("$TAG: hiding incoming call")
+
+        // This delay is just so when the user receives two calling signals one straight after the other [INCOMING -> CANCEL]
+        // Due to the signals being one after the other we are creating a notification when we are trying to cancel it, it wasn't properly
+        // cancelling vibration as probably when we were cancelling, the vibration object was still being created and started and thus
+        // never stopped.
         TimeUnit.MILLISECONDS.sleep(300L)
         notificationManager.cancel(NotificationConstants.CALL_INCOMING_NOTIFICATION_ID)
     }
--- a/app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
@@ -60,7 +60,7 @@ class CallNotificationManager @Inject constructor(private val context: Context) 
         // Due to the signals being one after the other we are creating a notification when we are trying to cancel it, it wasn't properly
         // cancelling vibration as probably when we were cancelling, the vibration object was still being created and started and thus
         // never stopped.
-        TimeUnit.MILLISECONDS.sleep(300L)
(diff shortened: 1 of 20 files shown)


borichellow (approved):
👍 worth a try IMO

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-1560/4/) **succeeded**.

- [com.wire.android-vv4.0.2(6008737)(6008737)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.0.2(6008737)(6008737)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
```diff
@@ -54,6 +55,12 @@ class CallNotificationManager @Inject constructor(private val context: Context)
 
     fun hideIncomingCallNotification() {
         appLogger.i("$TAG: hiding incoming call")
+
+        // This delay is just so when the user receives two calling signals one straight after the other [INCOMING -> CANCEL]
+        // Due to the signals being one after the other we are creating a notification when we are trying to cancel it, it wasn't properly
+        // cancelling vibration as probably when we were cancelling, the vibration object was still being created and started and thus
+        // never stopped.
+        TimeUnit.MILLISECONDS.sleep(CANCEL_CALL_NOTIFICATION_DELAY)
```
trOnk12, line 63: Sleep is putting a Thread to sleep right ? Perhaps we can do it on coroutine level and like suspend the coroutine instead of putting a Thread to sleep ?

trOnk12 (approved):

app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
alexandreferris: As you pointed out this could lead to a small performance issue as it holds a thread for 300ms, but in order for us to be able to test this today I will keep it as it is and then after that will try to work the fix with injecting a: `dispatcherProvider: DispatcherProvider` into the manager and have more time to test with a coroutine scope.

Thanks for pointing it out 🙇🏻

PR #1600, opened 2023-03-29
fix: crash when putting an ongoing call in background (AR-3260)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3260" title="AR-3260" target="_blank">BugAR-3260</a>  App crashing when being logged in with 2 accounts and in an active call.
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When there is an ongoing call and we put the in background, the app is sometimes crashing

### Causes (Optional)

OngoingCallService is started before creating the foreground notification.

```
03-29 17:07:33.005  7720  7720 E AndroidRuntime: android.app.ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{fbeead0 u0 com.wire.android.internal/com.wire.android.services.OngoingCallService}
03-29 17:07:33.005  7720  7720 E AndroidRuntime: 	at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:2152)
```

It looks like the system is taking some time to pass data via Intent to service.

UPDATE:
We are starting OnGoingCallService multiple times that's because ongoingCalls flow is emitting values multiple times.
Race condition could happen here, startForeground of second run could be executed before generating the notification.

### Solutions

This PR #1583 didn't fix the issue, so I am adding some delay here to give time to create the foreground notification before starting the service.

UPDATE:
The fix would be by adding distinctUntilChanged() when collecting the flow

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

Have a call.
Try to put the app in background(usually it crashes at the third attempt)

----

----

----
##### References

### DIFF
--- a/app/src/androidTest/java/com/wire/android/SelfDeletionTimerTest.kt
+++ b/app/src/androidTest/java/com/wire/android/SelfDeletionTimerTest.kt
@@ -1,0 +1,416 @@
+package com.wire.android
+
+import androidx.test.platform.app.InstrumentationRegistry
+import com.wire.android.ui.home.conversations.SelfDeletionTimer
+import com.wire.android.ui.home.conversations.model.ExpirationStatus
+import com.wire.kalium.logic.data.message.Message
+import org.junit.Test
+import kotlin.time.Duration.Companion.days
+import kotlin.time.Duration.Companion.hours
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+class SelfDeletionTimerTest {
+
+    private val selfDeletionTimer = SelfDeletionTimer(
+        context = InstrumentationRegistry.getInstrumentation().targetContext
+    )
+
+    @Test
+    fun givenTimeLeftIsAboveOneHour_whenGettingTheUpdateInterval_ThenIsEqualToMinutesLeftTillWholeHour() {
+        val selfDeletionTimer = selfDeletionTimer.fromExpirationStatus(
+            ExpirationStatus.Expirable(
+                expireAfter = 23.hours + 30.minutes,
+                selfDeletionStatus = Message.ExpirationData.SelfDeletionStatus.NotStarted
+            )
+        )
+        assert(selfDeletionTimer is SelfDeletionTimer.SelfDeletionTimerState.Expirable)
+        val interval = (selfDeletionTimer as SelfDeletionTimer.SelfDeletionTimerState.Expirable).updateInterval()
+        assert(interval == 30.minutes)
+    }
+
+    @Test
+    fun givenTimeLeftIsEqualToWholeHour_whenGettingTheUpdateInterval_ThenIsEqualToOneMinute() {
+        val selfDeletionTimer = selfDeletionTimer.fromExpirationStatus(
+            ExpirationStatus.Expirable(
+                expireAfter = 23.hours,
+                selfDeletionStatus = Message.ExpirationData.SelfDeletionStatus.NotStarted
+            )
+        )
+        assert(selfDeletionTimer is SelfDeletionTimer.SelfDeletionTimerState.Expirable)
+        val interval = (selfDeletionTimer as SelfDeletionTimer.SelfDeletionTimerState.Expirable).updateInterval()
+        assert(interval == 1.hours)
+    }
+
+    @Test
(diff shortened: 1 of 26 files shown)


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1600/1/) **failed**.

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1600/2/) **failed**.

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1600/3/) **succeeded**.

- [com.wire.android-vv4.1.1(6121292)(6121292)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6121292)(6121292)-staging-compat.apk)

borichellow:
As far as `startForeground` is called inside the `OngoingCallService` delay before starting the service wouldn't change anything. 

I found an answer for a similar problem on [redit](https://www.reddit.com/r/Vanced/comments/rrm9yb/android_12_get_very_often_error_messages_question/) According to it there is no 100% way to fix it, it's just bad android design. 

BUT 
I got an idea that needs to be checked: what if we call `startForeground` in `onCreate()`?  `onCreate` is definitely called earlier than `onStartCommand`. 
The only problem is that in onCreate we don't have the Intent to get all the call data. But we can try to put some placeholder data into the notification there and leave `generateForegroundNotification`  in `onStartCommand`. 

So it works like this: 
- Service start called
- once Service is created it calls `startForeground` with a "placeholder" notification
- once we get all the necessary for notification data (in `onStartCommand`) we call `startForeground` again with the same notification Id, but updated notification. 

Android will replace the existing "placeholder" notification with the new one (cause they have the same notification ID).

ohassine:
> As far as `startForeground` is called inside the `OngoingCallService` delay before starting the service wouldn't change anything.
> 
> I found an answer for a similar problem on [redit](https://www.reddit.com/r/Vanced/comments/rrm9yb/android_12_get_very_often_error_messages_question/) According to it there is no 100% way to fix it, it's just bad android design.
> 
> BUT I got an idea that needs to be checked: what if we call `startForeground` in `onCreate()`? `onCreate` is definitely called earlier than `onStartCommand`. The only problem is that in onCreate we don't have the Intent to get all the call data. But we can try to put some placeholder data into the notification there and leave `generateForegroundNotification` in `onStartCommand`.
> 
> So it works like this:
> 
> * Service start called
> * once Service is created it calls `startForeground` with a "placeholder" notification
> * once we get all the necessary for notification data (in `onStartCommand`) we call `startForeground` again with the same notification Id, but updated notification.
> 
> Android will replace the existing "placeholder" notification with the new one (cause they have the same notification ID).

Thanks for the review. 
The fix works here by adding a delay, but it is just a workaround that helped me finally find the root cause.

The issue is: 
We are starting `OnGoingCallService` multiple times that's because `ongoingCalls` flow is emitting values multiple times.
Race condition could happen here, startForeground of second run could be executed before generating the notification.

The fix would be by adding `distinctUntilChanged()` when collecting the flow

borichellow (approved):
nice catch 👍

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-1600/4/) **succeeded**.

- [com.wire.android-vv4.1.1(6129396)(6129396)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6129396)(6129396)-staging-compat.apk)

alexandreferris (approved):

AndroidBob:
Build [5](https://10.10.124.134/job/wire-android-reloaded/job/PR-1600/5/) **succeeded**.

- [com.wire.android-vv4.1.1(6130587)(6130587)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6130587)(6130587)-staging-compat.apk)

PR #1641, opened 2023-04-06
chore: small view model for connection button [AR-3252]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3252" title="AR-3252" target="_blank">TaskAR-3252</a>  Implement ViewModels for Smaller Composables
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Context
As our Android Reloaded Wire app continues to grow in complexity, we have encountered some challenges with maintaining large view models. This has led us to explore strategies for breaking down our view models into smaller, more manageable components.

### Decision 
We have decided to adopt the approach outlined in the article "Divide big view models to smaller" by Q42 Engineering, and implement it using the Resaca library for breaking down our view models into smaller components.

### Alternatives
We considered several alternative approaches, including using interfaces with functions for each particular composable, which were then passed in the composable tree. However, we ultimately decided that the approach outlined in the Q42 article and implemented with Resaca was the most suitable for our needs.

### Consequences
By dividing our large view models into smaller components, we expect to see improvements in several areas, including code organization, testability, maintainability, and reusability of the composables. Smaller view models will allow us to create more focused and targeted components, which can be reused across multiple screens and layouts, reducing the amount of duplicate code in our codebase. 

### Status
We have already begun implementing this approach in our Android Reloaded Wire app using the Resaca library for breaking down our view models into smaller components. We will continue to monitor its effectiveness and make adjustments as needed.


AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/2/) **failed**.

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/3/) **succeeded**.

- [com.wire.android-vv4.1.1(6181746)(6181746)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6181746)(6181746)-staging-compat.apk)

trOnk12 (commented):
Did not read through yet, but is there any reason this ADR is not included to confluence ?

app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
```diff
@@ -0,0 +1,199 @@
...
+    val viewModel: ConnectionActionButtonBaseViewModel = if (LocalInspectionMode.current) {
+        ConnectionActionButtonPreviewModel
+    } else {
+        hiltViewModelScoped<ConnectionActionButtonViewModel>(
+            key = "${ConnectionActionButtonViewModel.MY_ARGS_KEY}$userId",
+            defaultArguments = bundleOf(
+                EXTRA_USER_ID to userId.toString(),
+                EXTRA_USER_NAME to userName,
+                EXTRA_CONNECTION_STATE to connectionStatus.toString()
+            )
```
trOnk12, line 63: Did you check when the instance of ConnectionActionButtonViewModel gets cleared ? Like  we enter the screen having this ConnectionActionButton, we navigate back from the screen does this at this time gets cleared ? What if we scroll and more items do appear having this ConnectionActionButton, will they just spam more and more ViewModels ? Will they ever be cleared ?

trOnk12 (commented):
Great job ! :) It is awesome to see how your experience from Flutter comes in handy with Jetpack Compose and I am impressed of the effort that have put in to this subject. 

However, I am abit sceptical if we should be refactoring the code and taking this approach, before we have an answer how to handle displaying dialogs and bottomsheet, it is awesome that you handled the snackbar, but I think it is less complicated one, because we can display one globally, however with dialogs and bottomsheet I am afraid it will get abit more complicated because they are scoped to the "screen", but perhaps it will not be as hard as I think, anyways I think it is good idea to explore it first abit .

My concern is that if we do refactor following this approach, we may be forgetting sometthing that we could be doing in first place, while moving to this approach, if that would be the case it would mean that we would need again do refactor to adjust our code to work with bottomsheets and dialogs, if we explore that beforehand we would have a clear strategy and guideliness defined how to approach it from the very beginning and thus have less work in case we assumed something not correctly,

app/src/main/kotlin/com/wire/android/ui/common/snackbar/SnackBarViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    init {
+        viewModelScope.launch {
+            showSnackBar.observerSnackBarStatus.collect { result ->
+                result.message?.let {
+                    _snackBarMessage.emit(SnackBarState(message = it))
+                }
+            }
+        }
+    }
```
trOnk12, line 49: I wonder how this will work with multiple composable elements firing snackbar events 🤔 
Like before we had a snackbar per screen, navigating back would dismiss any not yet handled request to show a snackbar, with this approach I am concerned is, how it would work with smaller composables firing potentially more events to show a snackbar. If we navigate back those events are still processed and are not dropped since we have global handler. I guess that is not a problem since snackbar is a global thing that we display across all the screens, but perhaps there is more of a risk of a race-condition, for example, user clicks on something on Screen B, there is Composable element that emits a value to show a snackbar, we navigate back from screen B to screen A, the user clicks something that shows the snackbar, what if the event from snackbar B arrived after the one from screen A ? we would be showing a snackbar that should not be shown I am not sure if MutableSharedFlow allows this behaviour because perhaps it does guarantee the order, 

But i think it is worth to at least unit test it a bit, we were having this risk before this approach to :P but the chance of it happening were way less

Garzas:
> Did not read through yet, but is there any reason this ADR is not included to confluence ?

It will be after this will be merged :)

AndroidBob:
Build [5](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/5/) **succeeded**.

- [com.wire.android-vv4.1.1(6188768)(6188768)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6188768)(6188768)-staging-compat.apk)

AndroidBob:
Build [6](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/6/) **succeeded**.

- [com.wire.android-vv4.1.1(6232101)(6232101)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6232101)(6232101)-staging-compat.apk)

AndroidBob:
Build [7](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/7/) **succeeded**.

- [com.wire.android-vv4.1.1(6242360)(6242360)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.1.1(6242360)(6242360)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
Garzas: According to documentation: 

> Once you use the rememberScoped or viewModelScoped functions, the same object will be restored as long as the Composable is part of the composition, even if it temporarily leaves composition on configuration change (e.g. screen rotation, change to dark mode, etc.) or while being in the backstack.
> 
> For ViewModels, in addition to being forgotten when they're really not needed anymore, their coroutineScope will also be automatically canceled because ViewModel's onCleared method will be automatically called by this library.
> 
> 💡 Optional key: a key can be provided to the call, rememberScoped(key) { ... } or viewModelScoped(key) { ... }. This makes possible to forget an old object when there is new input data during a recomposition (e.g. a new input id for your ViewModel).
> 
> ⚠️ Note that ViewModels remembered with viewModelScoped should not be created using any of the Compose viewModel() or ViewModelProviders factories, otherwise they will be retained in the scope of the screen regardless of viewModelScoped. Also, if a ViewModel is remembered with rememberScoped its clean-up method won't be called, that's the reason to use viewModelScoped instead.

So when you go back it will be cleared but it will not be cleared when you will have it on back stack

app/src/main/kotlin/com/wire/android/ui/common/snackbar/SnackBarViewModel.kt
Garzas: I think we should not give option to fire many snackbars at once even without that solution 😄 But it's worth for checking or discuss how snackbar should behave when user close screen before seeing snackbar

AndroidBob:
Build [8](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/8/) **succeeded**.

- [com.wire.android-vv4.2.0(6294997)(6294997)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6294997)(6294997)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
trOnk12: I wonder what they mean with temporarily leaves composition, if we scroll through the list will the connection request be ongoing for that specific view item that requested it ? I think it will it would be weird if it was otherwise

AndroidBob:
Build [9](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/9/) **failed**.

AndroidBob:
Build [10](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/10/) **succeeded**.

- [com.wire.android-vv4.2.0(6301870)(6301870)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6301870)(6301870)-staging-compat.apk)

AndroidBob:
Build [11](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/11/) **succeeded**.

- [com.wire.android-vv4.2.0(6309713)(6309713)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6309713)(6309713)-staging-compat.apk)

AndroidBob:
Build [12](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/12/) **succeeded**.

- [com.wire.android-vv4.2.0(6335652)(6335652)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6335652)(6335652)-staging-compat.apk)

trOnk12 (approved):

app/src/main/kotlin/com/wire/android/ui/common/scaffold/WireScaffold.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }
+    Scaffold(
+        modifier = modifier,
+        topBar = topBar,
+        bottomBar = bottomBar,
+        snackbarHost = {
+            if (snackbarHost != null) {
+                snackbarHost()
```
borichellow, line 69: it means if `snackbarHost` is not null, then `snackbarHostState` is ignored and doesn't show the snackbars?

app/src/main/kotlin/com/wire/android/ui/common/snackbar/SnackBarViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+data class SnackBarState(
+    val showSnackBar: Boolean = false,
+    val message: UIText? = null
+)
+
+@HiltViewModel
+class SnackBarViewModel @Inject constructor(private val showSnackBar: ShowSnackBarUseCase) : ViewModel() {
```
borichellow, line 36: what is the point of having both `SnackBarViewModel` and `ShowSnackBarUseCase`? You can have just 1 `Singleton`. Didn't I get something?

app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButtonViewModel.kt
```diff
@@ -0,0 +1,212 @@
...
+class ConnectionActionButtonViewModel @Inject constructor(
+    private val navigationManager: NavigationManager,
+    private val dispatchers: DispatcherProvider,
+    private val sendConnectionRequest: SendConnectionRequestUseCase,
+    private val cancelConnectionRequest: CancelConnectionRequestUseCase,
+    private val acceptConnectionRequest: AcceptConnectionRequestUseCase,
+    private val ignoreConnectionRequest: IgnoreConnectionRequestUseCase,
+    private val unblockUser: UnblockUserUseCase,
+    private val getOrCreateOneToOneConversation: GetOrCreateOneToOneConversationUseCase,
+    private val showSnackBarUseCase: ShowSnackBarUseCase,
```
vitorhugods, line 85: I find it weird to have a `UseCase` called by `ViewModel` in order to display UI stuff.

Why would the `ViewModel` trigger a `SnackBar`? It should be the UI responsibility to handle UI, not the `ViewModel`.

Can't we have this handled by the Composables and use something like [`CompositionLocal`](https://developer.android.com/jetpack/compose/compositionlocal#intro) to pass SnackBar stuff when needed? _i.e._ when a Composable wants to display a `SnackBar`, it gets a `SnackBarManager` in its local composition tree in order to be displayed.

ViewModels should not worry about **HOW** things are displayed. They should worry about **WHAT** is displayed.

This pattern here is cementing the design to the ViewModel, something we do in some places (looking at Dialogs, specifically) but we should have avoided it since the beginning of the project.

If the ViewModel exposes a `Operation Failed` state. The Composable should figure out if that goes in a `Snackbar`, `Dialog`, `Toast`, or whatever are the design requirements.

app/src/main/kotlin/com/wire/android/ui/common/scaffold/WireScaffold.kt
Garzas: It's for compatibility when we have in some screen already implemented some snackbarHost

app/src/main/kotlin/com/wire/android/ui/common/snackbar/SnackBarViewModel.kt
Garzas: I just wanted to have separate ViewModel for WireScaffold to not use directly UseCase

AndroidBob:
Build [13](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/13/) **succeeded**.

- [com.wire.android-vv4.2.0(6406033)(6406033)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6406033)(6406033)-staging-compat.apk)

AndroidBob:
Build [14](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/14/) **succeeded**.

- [com.wire.android-vv4.2.0(6457878)(6457878)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6457878)(6457878)-staging-compat.apk)

AndroidBob:
Build [15](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/15/) **succeeded**.

- [com.wire.android-vv4.2.0(6483530)(6483530)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6483530)(6483530)-staging-compat.apk)

AndroidBob:
Build [16](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/16/) **succeeded**.

- [com.wire.android-vv4.2.0(6516830)(6516830)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6516830)(6516830)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButtonViewModel.kt
trOnk12: I think the only problem that may arrise if design decides to display "the message" ( because snackbar are only shown when we want to display a message, from my knowledge ) in some other form like toast or a dialog, in that case we could create another use case for toast and another one for dialog, that would imply that we also handle dialogs from the viewmodel ( at least showing them, because I remember @saleniuk  and @Garzas  had the idea to have the viewmodels designed for dialogs ) if this would be working fine and create no problems I do not think it creates some danger for us ? However, if we would not be able to create a use case allowing us to simply swap the use case inside the viewmodel, than I think it could be cubersome for the future changes, because all the sudden we need to change "flow of data" between the viewmodel to the Composable element to handle the dialog or a toast there. 

What else could change from the design would be the "action"to be performed in place of displaying a message, but that could be either navigation, which we are already doing inside the viewmodel, or changing some data or showing some additional stuff on the UI, but that could be a simple flag exposed inside the state in case of additiional stuff displayed or a matter of consuming data, which we are already doing both.

Perhaps it would be troublesome if we want to consume the use case in some kind of services or a worker, but how many times do we do that ? And if we do, we could adjust the viewmodel by then. 

Am I missing something here ?

trOnk12: Also if we let the component handle displaying of a dialog or a snackbar, it w ill not be a "stand-alone"component anymore it would have to have some dependency on a Scaffold snackbar state or either expose a lambda that would make it coupled to something from the outside world. 

Another thing that came to my mind is that the dialogs can also display snackbar/toast and another dialog. 

I personally thinl enclosing everything that makes the component not coupled to anything else is a better approach **IF** we are able to have use case for displaying dialogs/toast/snackbars and really think it through if we make that decision that we are not shooting us in foot and have the solutions for all the cases we can think of, otherwise this may be riscy, because we will do a refactor of a refactor of a refactor ;P

AndroidBob:
Build [17](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/17/) **succeeded**.

- [com.wire.android-vv4.2.0(6546031)(6546031)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6546031)(6546031)-staging-compat.apk)

AndroidBob:
Build [18](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/18/) **succeeded**.

- [com.wire.android-vv4.2.0(6552450)(6552450)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6552450)(6552450)-staging-compat.apk)

AndroidBob:
Build [19](https://10.10.124.134/job/wire-android-reloaded/job/PR-1641/19/) **succeeded**.

- [com.wire.android-vv4.2.0(6552727)(6552727)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6552727)(6552727)-staging-compat.apk)

Garzas:
Closed because of a new approach https://github.com/wireapp/wire-android-reloaded/pull/1810

PR #1768, opened 2023-05-09
fix(dev): NavBackStackEntry crash on HomeScreen [AR-3286]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3286" title="AR-3286" target="_blank">BugAR-3286</a>  Playstore crash: You cannot access the NavBackStackEntry's ViewModels after the NavBackStackEntry is destroyed
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sometimes we get a crash "You cannot access the NavBackStackEntry's ViewModels after the NavBackStackEntry is destroyed" with the stack trace leading to the `HomeScreen`.

### Causes (Optional)

We use `Scaffold`, `ModalBottomSheetLayout` and `ModalDrawer` on `HomeScreen`, all of them use `SubcomposeLayout` underneath (`BoxWithConstraints` also uses it). This layout breaks a little the usual order of actions for the composables by registering a separate sub-composition for that element which happens independently and after the parent's composition, sometimes even if the value that's required is already destroyed (`NavBackStackEntry` in this case).

More detailed explanation is here: https://wearezeta.atlassian.net/browse/ACOL-105

### Solutions

According to the Google's response:

> To fix this would require,
> 1. Ensuring all composition is performed in strict dependency order (parent first, sub-compositions after parents).
> 2. Communicate to the sub-compositions when they are to be disposed by the parent and ignore compositions in the round they are to be disposed.

The workaround in this case is to not allow recomposing the `NavHost` if we already know that the `NavBackStackEntry` has been already destroyed.

#### How to Test

Open the app, navigate to the specific conversation and then spam "back" arrow button from the top bar, it should close the app without any crash. That's the only way that I found to reproduce this crash.

### Notes (Optional)

- We should find a way to disallow the ability to click "back" arrow button multiple times, because right now it schedules a "back" navigation every time so the user can close the app by clicking multiple times - it should be possible to only go back one level from a particular screen.
- We should also consider reducing the number of nested navigations that we have, on `HomeScreen` `NavHost` is currently not needed at all so it only increases complexity without any major benefit.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
@@ -44,10 +44,12 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.graphics.RectangleShape
 import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.lifecycle.Lifecycle
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import com.wire.android.R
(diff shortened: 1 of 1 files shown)


MohamadJaara (approved):

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1768/2/) **succeeded**.

- [com.wire.android-vv4.2.0(6467099)(6467099)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6467099)(6467099)-staging-compat.apk)

tmspzz (approved):

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1768/3/) **succeeded**.

- [com.wire.android-vv4.2.0(6469293)(6469293)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6469293)(6469293)-staging-compat.apk)

PR #1774, opened 2023-05-10
fix: issue not identifying correctly already consumed intents [AR-3377] [AR-3387]

gongracr:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3377" title="AR-3377" target="_blank">BugAR-3377</a>  PlayTest 02.05 - After sharing an image, while opening a deeplink on the background there is share image screen
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently there is a buggy behavior in the app causing that if we open it via a navigation deeplink (message notification, or by importing media from other apps), when sending the app to background and resuming it, the initial navigation screen shown when handling the deeplink will pop up again.

### Solutions

Add an extra value on the intent once the user has handled the initial intent and check if it was already set when handling new intents. If so, we don't handle it again.

### Attachments (Optional)
| Before | After |
| <video src= "https://github.com/wireapp/wire-android-reloaded/assets/2468164/5bc2e8de-2782-4ace-8f80-db9b56b69d4a"/> | <video src= "https://github.com/wireapp/wire-android-reloaded/assets/2468164/7db5baa0-b543-470f-b324-da1fdb015c33" /> |

##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
@@ -190,9 +190,9 @@ class WireActivity : AppCompatActivity() {
 
         DisposableEffect(navController) {
             val updateScreenSettingsListener = NavController.OnDestinationChangedListener { controller, _, _ ->
-                    currentKeyboardController?.hide()
-                    updateScreenSettings(controller)
-                }
+                currentKeyboardController?.hide()
+                updateScreenSettings(controller)
+            }
             navController.addOnDestinationChangedListener(updateScreenSettingsListener)
             navController.addOnDestinationChangedListener(currentScreenManager)
 
@@ -403,11 +403,13 @@ class WireActivity : AppCompatActivity() {
         if (intent == null
             || intent.flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY != 0
             || savedInstanceState?.getBoolean(HANDLED_DEEPLINK_FLAG, false) == true
+            || intent.getBooleanExtra(HANDLED_DEEPLINK_FLAG, false)
         ) {
             return
+        } else {
+            viewModel.handleDeepLink(intent)
+            intent.putExtra(HANDLED_DEEPLINK_FLAG, true)
         }
-
-        viewModel.handleDeepLink(intent)
     }
 
     companion object {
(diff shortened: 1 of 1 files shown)


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1774/1/) **failed**.

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1774/2/) **succeeded**.

- [com.wire.android-vv4.2.0(6473850)(6473850)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6473850)(6473850)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -396,18 +396,21 @@ class WireActivity : AppCompatActivity() {
...
         intent: Intent?,
         savedInstanceState: Bundle? = null
     ) {
         if (intent == null
             || intent.flags and Intent.FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY != 0
             || savedInstanceState?.getBoolean(HANDLED_DEEPLINK_FLAG, false) == true
+            || intent.getBooleanExtra(HANDLED_DEEPLINK_FLAG, false)
         ) {
             return
+        } else {
```
borichellow, line 410: this `else` is not needed as far as there is `return` in prev. `if`

borichellow (approved):
great! Simple and elegant solution 👍

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
gongracr: Where would you add the removal of the consumed intent extras?

vitorhugods: I _think_ @borichellow means to just use the if as a guard clause:

```kotlin
if (isIntentNullOrConsumed) {
    return
}

// handle intent and mark as consumed
```

No need to maintain an extra level of nesting
```kotlin
if (isIntentNullOrConsumed) {
    return
} else {
    // handle intent and mark as consumed
}
```

vitorhugods (approved):
Nice solution!

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
gongracr: I think I will leave the else clause as a way of being more verbose/self-explanatory/redundant

PR #1795, opened 2023-05-16
fix: Notification observing on app hidden (AR-3369) (AR-3357)

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3369" title="AR-3369" target="_blank">BugAR-3369</a>  Fix notifications observing on app hidden
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

1. for the users that have turned on PersistentConnection, notifications were observed twice (in PersistentWebSocketService and in GlobalObserversManager)
2. when the app is killed notifications are delayed or even missed

### Causes (Optional)

1.  in `GlobalObserversManager` calls `observeNotificationsAndCallsWhileRunning` for all the accounts including the accs that have `PersistentWebSocket` turned on.
2. few reasons: 
 - forgot to merge [small improvements in kalium ](https://github.com/wireapp/kalium/pull/1741)
 - the algorithm of fetching notifications on push itself 

**When the push comes we do:** 
1. start observing the notifications (but only after WS-connection goes to LIVE) and show notifications 
2. call `connectionPolicyManager.handleConnectionOnPushNotification` which fetch all the pending events and disconnect once we LIVE
3. stop observing the notification

The problem is that `getNotificationsUseCase` doesn't emit anything until we LIVE (it's was done like that to not spam user by the notification while he's syncing after long absence and just show 1 big notification about all the missed events) 
AND `connectionPolicyManager.handleConnectionOnPushNotification` disconnects once we LIVE and the notification observation is canceled, so not time for getting the notification from DB and show it.

### Solutions

1. call `observeNotificationsAndCallsWhileRunning` only for the users that have PersistentConnection turned on, and for others - launch `PersistentWebSocketService`
2.  update the kalium and set some delay in `connectionPolicyManager.handleConnectionOnPushNotification` to stay alive for a while. Experimentally I found out that calls need about 500ms after connection is LIVE to fetch, handle and show notification for incoming call, so I put 1s delay to be sure. But this is definitely the place to investigate why such a big lag there and improve it.


AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/1/) **failed**.

AndroidBob:
Build [2](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/2/) **failed**.

app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
```diff
@@ -39,19 +32,7 @@ class GlobalObserversManager @Inject constructor(
...
             .collect { list ->
                 notificationChannelsManager.createUserNotificationChannels(list.map { it.first })
-                list.map {
-                    it.first.id
-                }.run {
-                    notificationManager.observeNotificationsAndCallsWhileRunning(this, scope) {
-                        openIncomingCall(it.conversationId)
-                    }
-                }
+                list.map { it.first.id }
```
ohassine, line 61: Do we need to map this list ?

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
```diff
@@ -451,6 +453,14 @@ class WireActivityViewModel @Inject constructor(
...
                     is ObservePersistentWebSocketConnectionStatusUseCase.Result.Success -> {
                         result.persistentWebSocketStatusListFlow.collect { statuses ->
+                            val usersToObserve = statuses
+                                .filter { !it.isPersistentWebSocketEnabled }
+                                .map { it.userId }
+
+                            notificationManager.observeNotificationsAndCallsWhileRunning(
+                                usersToObserve,
+                                this
+                            ) { call -> openIncomingCall(call.conversationId) }
```
ohassine: This for persistent websocket, sould we call observeNotificationsAndCallsPersistently() here ?

ohassine (commented):
Shouldn't we keep observing for messages and calls even the app is killed ? This way if we kill the app the activity got destroyed and we the app will stop observing

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
```diff
@@ -451,6 +453,14 @@ class WireActivityViewModel @Inject constructor(
 
                     is ObservePersistentWebSocketConnectionStatusUseCase.Result.Success -> {
                         result.persistentWebSocketStatusListFlow.collect { statuses ->
+                            val usersToObserve = statuses
+                                .filter { !it.isPersistentWebSocketEnabled }
+                                .map { it.userId }
+
+                            notificationManager.observeNotificationsAndCallsWhileRunning(
```
saleniuk: If the delay added in `handleConnectionOnPushNotification` makes it work then is this move from `GlobalObserversManager` to `WireActivityViewModel` necessary? I mean, now we have a single activity so maybe it's safe to assume "WireActivity is open == the app is running" but that can not always be the case and I don't quite understand how moving it here is a solution, could you explain why we need to move it here? 🤔 Also, won't that create a race condition with creating the notification channels because both are executed independently now?

app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
borichellow: forgot to remove, thanks :)

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
borichellow: this observation is for users that disabled PersistentWebSocket, so we want to observe their message only when the app is running.

borichellow:
> Shouldn't we keep observing for messages and calls even the app is killed ? This way if we kill the app the activity got destroyed and we the app will stop observing

The idea of these observations is that we have 2 options for users: 
1. Regular observation: app observes messages only while it's running, and when the app is killed and user receives some message we'll get FCM push, which triggers synchronization and user will get a notification. 
2. Persistent connection, which is run in `PersistentWebSocketService` in that case user is always connected to websocket and receives all the messages. 

`option 1` is better in a performance context, but it's totally un-usable for the devices that have no GoogleServices and can't receive FCMs, for such devices we have `option 2` which is worse in performance but guarantees user receives all the messages in time.

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
borichellow: moving it here doesn't actually solve any known issue. But the problem of having it in `GlobalObserversManager`  is a coroutine scope we use for observation. 
In `GlobalObserversManager` we use scope that lives longer, than ViewModelScope so that we may observe the messages longer than the app lives, which is redundant. 
Once the app is hidden/killed we disconnect from the websocket and if user receives anything, we'll get FCM push which triggers a new connection and new messages observation.

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/4/) **failed**.

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
borichellow: about independent notifications and channels creating: you're right, it's rare, but possible case, so I'll move it to ViewModel too.

saleniuk: > we may observe the messages longer than the app lives

You mean longer than the Activity lives, right? Because `GlobalObserversManager` is also a part of the app so if it lives, it means that the app also still lives. 😅 On the other hand, we can have gaps between the Activity is being closed, websocket being closed and the app starts receiving pushes and maybe handling notification twice in that case is still better than missing the notification until next one arrives 😅 Maybe I'm wrong but redundancy still shouldn't happen because in `observeMessageNotificationsOnceJob` we check if `observingWhileRunningJobs` is still running 🤔 

> about independent notifications and channels creating: you're right, it's rare, but possible case, so I'll move it to ViewModel too.

I may be dangerous, I think we already had that in the ViewModel and it was causing issues, because channels were created too late - for instance when your app updates in the background and you don't open the app but receive a push - until you open the app and create this ViewModel you don't have channels created. 🙈

AndroidBob:
Build [5](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/5/) **failed**.

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
borichellow: Yeah, you are right about ViewModel. 
So I moved the channelCreating and notificationObserving back to the GlobalObserversManager AND added there running `PersistentWebSocketService` so we don't do `observeNotificationsAndCallsWhileRunning` for the users that have `PersistentWebSocket` ON  :)

AndroidBob:
Build [8](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/8/) **succeeded**.

- [com.wire.android-vv4.2.0(6554149)(6554149)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6554149)(6554149)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
```diff
@@ -26,25 +30,50 @@ class GlobalObserversManager @Inject constructor(
...
+                            openIncomingCall(it.conversationId)
+                        }
+                    }
+
+                if (persistentStatuses.any { it.isPersistentWebSocketEnabled }) {
+                    if (!servicesManager.isPersistentWebSocketServiceRunning()) {
+                        servicesManager.startPersistentWebSocketService()
                     }
+                } else {
+                    servicesManager.stopPersistentWebSocketService()
```
ohassine: When the app is in background this will throw an exception `ForegroundServiceStartNotAllowedException` because you are trying to start a foreground service from background. That's why we put it in the `WireActivityViewModel`

app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
```diff
@@ -26,25 +30,50 @@ class GlobalObserversManager @Inject constructor(
...
 ) {
     private val scope = CoroutineScope(SupervisorJob() + dispatcherProvider.io())
 
     fun observe() {
-        scope.launch { observeAccountsToCreateChannels() }
+        scope.launch { observeAccountsToCreateChannelsAndObserveNotifications() }
     }
 
-    private suspend fun observeAccountsToCreateChannels() {
+    private suspend fun observeAccountsToCreateChannelsAndObserveNotifications() {
```
ohassine: Maybe a simpler name ? `setUpNotifications()`

app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
borichellow: fair point, thanks. 
So i moved the starting service back to ViewModel, but kept "filtering if not persistent" in the GlobalObservers, to not observe notifications twice for the users that enabled the persistante connection.

AndroidBob:
Build [10](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/10/) **succeeded**.

- [com.wire.android-vv4.2.0(6586890)(6586890)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6586890)(6586890)-staging-compat.apk)

AndroidBob:
Build [12](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/12/) **succeeded**.

- [com.wire.android-vv4.2.0(6594635)(6594635)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6594635)(6594635)-staging-compat.apk)

ohassine (approved):

alexandreferris (approved):

AndroidBob:
Build [13](https://10.10.124.134/job/wire-android-reloaded/job/PR-1795/13/) **succeeded**.

- [com.wire.android-vv4.2.0(6595213)(6595213)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6595213)(6595213)-staging-compat.apk)

PR #1797, opened 2023-05-17
feat: add self deletion mechanism for import screen [AR-3294]

gongracr:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3294" title="AR-3294" target="_blank">StoryAR-3294</a>  Send self deleting message (asset) from outside the app
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently we have no way to import media assets that respect the current self deletion team and conversation settings status. Therefore, I added this mechanism to display the current self deletion settings for conversations that have enforced timers or allow the user to change and persist it for conversations where these settings still not exist.

#### Test Coverage (Optional)

#### How to Test

Just import assets to Wire.

### Attachments

https://github.com/wireapp/wire-android-reloaded/assets/2468164/86a1b149-da38-4116-a5ce-12993965788d

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/ModalSheetHeaderItem.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/ModalSheetHeaderItem.kt
@@ -25,9 +25,11 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
+import androidx.compose.material3.LocalContentColor
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.Dp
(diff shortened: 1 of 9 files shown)


app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/ModalSheetHeaderItem.kt
```diff
@@ -40,30 +42,33 @@ fun ModalSheetHeaderItem(header: MenuModalSheetHeader = MenuModalSheetHeader.Gon
...
-                    bottom = header.customBottomPadding ?: dimensions().modalBottomSheetHeaderVerticalPadding
-                )
-            ) {
-                header.leadingIcon()
-                Spacer(modifier = Modifier.width(dimensions().spacing8x))
-                Text(
-                    text = header.title,
-                    style = MaterialTheme.wireTypography.title02
-                )
+            CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.secondary) {
```
gongracr, line 47: Not really in the scope of this PR, but this is fixing last remaining issue for dark mode on bottom sheet menus title not being readable enough.

alexandreferris (approved):

app/src/main/kotlin/com/wire/android/ui/sharing/ImportMediaAuthenticatedViewModel.kt
```diff
@@ -299,6 +345,26 @@ class ImportMediaAuthenticatedViewModel @Inject constructor(
         }
     }
 
+    fun onNewConversationPicked(conversationId: ConversationId) = viewModelScope.launch {
+        importMediaState = importMediaState.copy(
+            selfDeletingTimer = observeSelfDeletionSettingsForConversation(
+                conversationId = conversationId,
+                considerSelfUserSettings = true
+            ).first()
```
Garzas, line 354: I'm not convinced to use first here because in meantime when some setting will change we should update it

Garzas (changes requested):
One small question :)

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1797/3/) **succeeded**.

- [com.wire.android-vv4.2.0(6535088)(6535088)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6535088)(6535088)-staging-compat.apk)

Garzas (approved):

app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/ModalSheetHeaderItem.kt
saleniuk: I think there's a deeper problem and these text colors are just one of the bad effects of that. The issue is that we still use many material2 components but we only support theme from material3.
In this case, we set our custom `sheetBackgroundColor` and by default it tries to guess the correct `sheetContentColor` by comparing it to material2 color scheme, but we don't specify it, we only have material3 color scheme specified in our theme.
But even adding material2 theme or setting custom `sheetContentColor` won't work because it will set `LocalContentColor` for material2, so if we use `Text` from material3, its content won't use this color.
This can become a problem at any place where we still mix material2 and material3 components, so we should migrate to material3 only.
I can see that we still use these components from material2:
```
SwipeToDismiss
DropdownMenu
ModalBottomSheet
Icon
Divider
Scaffold
BottomSheetScaffold
ModalDrawer
```
Fortunately all of them are now also available on material3, we just need to upgrade it to 1.1 and migrate components.
https://developer.android.com/jetpack/compose/designsystems/material2-material3

gongracr: Thanks for the detailed explanation 🙏  For now, let's just go with this temporal fix. But I would be very glad to see these migation to material 3 components and simplify the way we handle these color fallbacks in a separated ticket/PR.

PR #1823, opened 2023-05-26
fix: ForgroundService crash (AR-3471)

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AR-3471" title="AR-3471" target="_blank">BugAR-3471</a>  Play store crash: ActivityThread.throwRemoteServiceException
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

[Crash from the play store](https://play.google.com/console/u/0/developers/7098984309886892484/app/4973241010395499500/vitals/crashes/d8241f2d6ddc185210f2b4d7f5452c52/details?days=28&versionCode=6518403&installedFrom=PLAY_STORE)

### Causes (Optional)

Probably crashes because we call `Context.startForegroundService(...)` and then call `Context.stopService(...)` before `Service.startForeground(...)` is called.

There’s an issue created: [Google Issue Tracker](https://issuetracker.google.com/issues/76112072) but it’s not fixed on their side as it’s probably intentional.

### Solutions

call `startForeground` in `Service.OnCreate` with some placeholder notification and then replace that notification with the right notification in `Service.onStartCommand`

### DIFF
--- a/app/src/main/kotlin/com/wire/android/notification/NotificationChannelsManager.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/NotificationChannelsManager.kt
@@ -64,10 +64,12 @@ class NotificationChannelsManager @Inject constructor(
             val groupId = createNotificationChannelGroup(user.id, user.handle ?: user.name ?: user.id.value)
 
             createIncomingCallsChannel(groupId, user.id)
-            createOngoingNotificationChannel(groupId, user.id)
             createMessagesNotificationChannel(user.id, groupId)
             createPingNotificationChannel(user.id, groupId)
         }
+
+        // OngoingCall is not user specific channel, but common for all users.
+        createOngoingNotificationChannel()
     }
 
     /**
@@ -113,15 +115,14 @@ class NotificationChannelsManager @Inject constructor(
         notificationManagerCompat.createNotificationChannel(notificationChannel)
     }
 
-    private fun createOngoingNotificationChannel(groupId: String, userId: UserId) {
-        val chanelId = NotificationConstants.getOngoingChannelId(userId)
+    private fun createOngoingNotificationChannel() {
+        val chanelId = NotificationConstants.ONGOING_CALL_CHANNEL_ID
         val notificationChannel = NotificationChannelCompat
             .Builder(chanelId, NotificationManagerCompat.IMPORTANCE_MAX)
             .setName(NotificationConstants.ONGOING_CALL_CHANNEL_NAME)
             .setVibrationEnabled(false)
             .setImportance(NotificationManagerCompat.IMPORTANCE_MAX)
             .setSound(null, null)
-            .setGroup(groupId)
             .build()
 
         notificationManagerCompat.createNotificationChannel(notificationChannel)
--- a/app/src/main/kotlin/com/wire/android/notification/NotificationChannelsManager.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/NotificationChannelsManager.kt
@@ -101,9 +101,9 @@ class NotificationChannelsManager @Inject constructor(
             .setUsage(getAudioAttributeUsageByOsLevel())
             .build()
 
-        val chanelId = NotificationConstants.getIncomingChannelId(userId)
+        val channelId = NotificationConstants.getIncomingChannelId(userId)
         val notificationChannel = NotificationChannelCompat
-            .Builder(chanelId, NotificationManagerCompat.IMPORTANCE_MAX)
+            .Builder(channelId, NotificationManagerCompat.IMPORTANCE_MAX)
             .setName(NotificationConstants.INCOMING_CALL_CHANNEL_NAME)
             .setImportance(NotificationManagerCompat.IMPORTANCE_MAX)
             .setSound(incomingCallSoundUri, audioAttributes)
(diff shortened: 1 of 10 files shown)


app/src/main/kotlin/com/wire/android/services/OngoingCallService.kt
```diff
@@ -84,12 +85,15 @@ class OngoingCallService : Service() {
...
+            if (!isStopCall) {
+                appLogger.w(
+                    "$TAG: stopSelf. Reason: some of the parameter is absent. " +
+                            "userIdString: ${userIdString?.obfuscateId()}, " +
+                            "conversationIdString: ${conversationIdString?.obfuscateId()}, " +
+                            "callName: $callName"
+                )
+            }
+            stopForeground(STOP_FOREGROUND_REMOVE)
             stopSelf()
```
alexandreferris: Shouldn't 
```
    stopForeground(STOP_FOREGROUND_REMOVE)
    stopSelf()
```

be called inside the if as well? 🤔  Just trying to understand why we having the if logic only for the logger

alexandreferris (commented):
Great work! Just a quick question

AndroidBob:
Build [1](https://10.10.124.134/job/wire-android-reloaded/job/PR-1823/1/) **succeeded**.

- [com.wire.android-vv4.2.0(6612571)(6612571)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6612571)(6612571)-staging-compat.apk)

ohassine:
I think it's better to find the root cause..We need to know why we are calling stop just after starting it.

app/src/main/kotlin/com/wire/android/services/OngoingCallService.kt
borichellow: we can't create the notification if there is no `userId`, or `conversationId`, or `callName`, it means something is wrong, so we should stop the service and stopForeground. 
But if the flag `isStopCall == true` means that we did it intentionally cause we want the service to stop, so there is no need to log "Something went wrong" message.

borichellow:
> I think it's better to find the root cause..We need to know why we are calling stop just after starting it.

Looks like this is actually the root of the problem: service should not crash when we want to finish it, no matter when it's called.

I think it may happen on some slow device, scenario is something like: 
1. start a call
2. hide the app
   2.1 app suppose to show the notification by `startForeground`
3. open the app again immediately 

So if the device is slow enough or by some power-saving rules the app has not enough time to startForeground it crash.

But I can't check it as I wasn't able to reproduce it on any of my devices.

ohassine:
> > I think it's better to find the root cause..We need to know why we are calling stop just after starting it.
> 
> Looks like this is actually the root of the problem: service should not crash when we want to finish it, no matter when it's called.
> 
> I think it may happen on some slow device, scenario is something like:
> 
> 1. start a call
> 2. hide the app
>    2.1 app suppose to show the notification by `startForeground`
> 3. open the app again immediately
> 
> So if the device is slow enough or by some power-saving rules the app has not enough time to startForeground it crash.
> 
> But I can't check it as I wasn't able to reproduce it on any of my devices.

In that case It should be an OS bug I think. My suspicion is that we are doing some work that is delaying `startForeground() `call. 
 We are calling `fromStringToQualifiedID` before invoking `startForeground()` and that mapper requires selfUserId that is retrieved from database. It could be the app is still mapping the userId and the service got stopped

saleniuk (approved):

AndroidBob:
Build [3](https://10.10.124.134/job/wire-android-reloaded/job/PR-1823/3/) **failed**.

AndroidBob:
Build [4](https://10.10.124.134/job/wire-android-reloaded/job/PR-1823/4/) **failed**.

borichellow:
> > > I think it's better to find the root cause..We need to know why we are calling stop just after starting it.
> > 
> > 
> > Looks like this is actually the root of the problem: service should not crash when we want to finish it, no matter when it's called.
> > I think it may happen on some slow device, scenario is something like:
> > 
> > 1. start a call
> > 2. hide the app
> >    2.1 app suppose to show the notification by `startForeground`
> > 3. open the app again immediately
> > 
> > So if the device is slow enough or by some power-saving rules the app has not enough time to startForeground it crash.
> > But I can't check it as I wasn't able to reproduce it on any of my devices.
> 
> In that case It should be an OS bug I think. My suspicion is that we are doing some work that is delaying `startForeground() `call. We are calling `fromStringToQualifiedID` before invoking `startForeground()` and that mapper requires selfUserId that is retrieved from database. It could be the app is still mapping the userId and the service got stopped

You reminded me the idea that I had some time ago but forgot to try. 
I re-do the PR, now we'll call `startForeground` in `Service.OnCreate` with some placeholder notification and then replace that placeholder with the right (filled) notification in `Service.onStartCommand`. So the notification will be there from the very beginning of the Service life.

borichellow:
After new update approval from @saleniuk is not valid anymore, so need 2 again :)

AndroidBob:
Build [5](https://10.10.124.134/job/wire-android-reloaded/job/PR-1823/5/) **succeeded**.

- [com.wire.android-vv4.2.0(6639959)(6639959)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6639959)(6639959)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/notification/NotificationChannelsManager.kt
```diff
@@ -113,15 +115,14 @@ class NotificationChannelsManager @Inject constructor(
         notificationManagerCompat.createNotificationChannel(notificationChannel)
     }
 
-    private fun createOngoingNotificationChannel(groupId: String, userId: UserId) {
-        val chanelId = NotificationConstants.getOngoingChannelId(userId)
+    private fun createOngoingNotificationChannel() {
+        val chanelId = NotificationConstants.ONGOING_CALL_CHANNEL_ID
```
ohassine: typo `channelId` *

ohassine (approved):

saleniuk (approved):
Sounds good, I hope it will ensure startForeground is always executed before any stopService/stopSelf 👌

AndroidBob:
Build [6](https://10.10.124.134/job/wire-android-reloaded/job/PR-1823/6/) **succeeded**.

- [com.wire.android-vv4.2.0(6646935)(6646935)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-vv4.2.0(6646935)(6646935)-staging-compat.apk)

PR #2057, opened 2023-08-03
fix: crash when stopping OngoingCallService [WPB-2320] [WPB-1836] [WPB-3457]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

The PR is big, but 80% of that are tests. 😅

### Issues

There are still multiple crashes related to stopping OngoingCallService: `ForegroundServiceDidNotStartInTimeException` or `Context.startForegroundService() did not then call Service.startForeground()`, depending on Android version.
Also, we have one service for all logged in users so there's a race condition because one account can close the service even if another one has an ongoing call.

### Causes (Optional)

The problem is that if we start service as a foreground one, then it has to call `startForeground` and show notification, if we stop such service before it does that, then the app crashes.

### Solutions

We already tried to call `startForeground` with some placeholder notification when creating service, before `onStartCommand` is called, but it's not enough and crashes still happen quite commonly, so to fix that the debounce is added to take 200 milliseconds to check whether starting this service is really necessary - if "stop service" happens within that time then it won't even start and prevent us from the crash. Also, stopping the service, if already started but before it's in the foreground, is done directly inside the service, so even if we call it to stop too soon, then it will still execute `onStartCommand`, show a placeholder notification for a split second and then safely close the service.
For the issue with multiple accounts, a common SharedFlow is provided which collects data for all accounts and then starts or stops the service if any of them has an ongoing call. The same is implemented for incoming call notification (we also show a single incoming call notification for all accounts). The service is started and then it handles keeping itself alive as long as there is an ongoing call for any of valid accounts.

#### Test Coverage (Optional)

----

----

----
##### References


AndroidBob:
Build [895](https://10.10.124.134/job/AR-build-pipeline/895/) **failed**.

AndroidBob:
Build [896](https://10.10.124.134/job/AR-build-pipeline/896/) **failed**.

AndroidBob:
Build [897](https://10.10.124.134/job/AR-build-pipeline/897/) **failed**.

AndroidBob:
Build [902](https://10.10.124.134/job/AR-build-pipeline/902/) **succeeded**.

- [com.wire.android-v4.3.0(7211524)(7211524)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.3.0(7211524)(7211524)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
```diff
@@ -106,7 +158,7 @@ class CallNotificationManager @Inject constructor(private val context: Context)
         return NotificationCompat.Builder(context, channelId)
             .setContentTitle(callName)
             .setContentText(context.getString(R.string.notification_ongoing_call_content))
-            .setPriority(NotificationCompat.PRIORITY_HIGH)
+            .setPriority(NotificationCompat.PRIORITY_DEFAULT)
```
ohassine, line 161: Why downgrading the priority here and in other places ?

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
```diff
@@ -25,26 +25,109 @@ import android.content.Context
...
+            appLogger.i("ServicesManager: stopping OngoingCallService for user:${userId.toLogString()}")
+            ongoingCallServiceForUsers.update { it.filter { it.userId != userId } }
+        }
+    }
+
+    fun stopOngoingCallServiceForAll() {
+        scope.launch {
+            appLogger.i("ServicesManager: stopping OngoingCallService for all users")
+            ongoingCallServiceForUsers.emit(emptyList())
+        }
```
ohassine: We actually can have only one ongoing call per app session, even if you switch to other user the current call will be ended. 
so we will need just `stopOngoingCallService()`

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
```diff
@@ -25,26 +25,109 @@ import android.content.Context
...
+
+    suspend fun currentlyOngoingCall(): Flow<OngoingCallData?> = ongoingCallServiceForUsers
+        .combine(
+            coreLogic.getGlobalScope().observeValidAccounts()
+        ) { ongoingCalls: List<OngoingCallData>, validAccounts: List<Pair<SelfUser, Team?>> ->
+            ongoingCalls.lastOrNull { ongoingCallData ->
+                // filter out calls that are for accounts that are not valid anymore
+                validAccounts.any { (selfUser, _) -> selfUser.id == ongoingCallData.userId }
+            }
+        }
```
ohassine: Is this used to check if we have ongoing calls ? we have `ObserveOngoingCallsUseCase` for that if it's the case

app/src/main/kotlin/com/wire/android/services/OngoingCallService.kt
```diff
@@ -74,61 +80,71 @@ class OngoingCallService : Service() {
...
+
+        fun newIntentToStop(context: Context): Intent =
             Intent(context, OngoingCallService::class.java).apply {
-                putExtra(EXTRA_USER_ID, userId)
-                putExtra(EXTRA_CONVERSATION_ID, conversationId)
-                putExtra(EXTRA_CALL_NAME, callName)
+                putExtra(EXTRA_STOP_SERVICE, true)
             }
+
+        var serviceState: AtomicReference<ServiceState> = AtomicReference(ServiceState.NOT_STARTED)
```
ohassine, line 161: Would be safer if we add `private set`

app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
saleniuk: We don't need this ongoing call notification to pop up on screen every time we put the app into the background as the user already is aware that the call is ongoing - it can be a little bit silent one so it shows but doesn't disturb the screen. Do you have different opinion on that?

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
saleniuk: I know, but checks for each account are done independently, so if we have users A and B, user A has an ongoing call and B does not, if the order in which app gets this info is first A then B, then the service will start and stop immediately, so this solution is to provide a buffer to gather data from all users instead of relying on only one of them - now if the app gets the info that user A needs a service, if then user B comes up and tells that doesn't need it, the service will still be running because we still keep the info that user A needs it.
"stop for user" means that we let the app know that this user doesn't need this service, "stop for all" means that we should mark all users as not needing the service and stop the service instantly.

Names of these functions are not the best, so I changed them to unify it with handling incoming calls (one method "handle" instead of "start" and "stop for user"), let me know if that's ok or maybe do you have any ideas for better names for them?

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
saleniuk: Yep, but `ObserveOngoingCallsUseCase` returns calls only for the specific user but we need to get info for all users, that's why there's this buffer, I renamed it to `ongoingCallsData`

app/src/main/kotlin/com/wire/android/services/OngoingCallService.kt
saleniuk: Right, forgot about that 👍

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
saleniuk: Yep, but `ObserveOngoingCallsUseCase` returns calls only for the specific user but we need to get info for all users, that's why there's this buffer, I renamed it to `ongoingCallsData`

AndroidBob:
Build [933](https://10.10.124.134/job/AR-build-pipeline/933/) **succeeded**.

- [com.wire.android-v4.3.0(7243892)(7243892)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.3.0(7243892)(7243892)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/notification/CallNotificationManager.kt
ohassine: Makes sense👍

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
ohassine: I got your point, but what's the issue if the service and stop immediately ? if it's a crash, the delay that we added shouldn't fix it

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
ohassine: Yes `ObserveOngoingCallsUseCase` will observe ongoing calls for the current selected user and that's what we need I would say, we can't have ongoing calls for other users

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
saleniuk: The crash happens because `stopService` happens too soon, before `startForeground` is executed, so adding a debounce of 200millis is usually enough for the app to execute it, and even if not, then I also implemented a mechanism to do it in right order, so maybe debounce is redundant but it will actually also reduce starting and stopping the service if not needed so the app won't use resources when not needed

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
saleniuk: Ok, but then, why did the app already check the ongoing calls for all valid users? I just wanted to keep the same logic, so maybe we already had it wrong. Are you sure that the app always switches the user when the call becomes ongoing so that we can rely on only checking it for the currently selected user? 🤔

ohassine: I didn't work on that part, maybe we check valid users to end the call on logout.
And yes we do switch the the user when accepting a call on non current user

AndroidBob:
Build [958](https://10.10.124.134/job/AR-build-pipeline/958/) **succeeded**.

- [com.wire.android-v4.3.0(7252673)(7252673)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.3.0(7252673)(7252673)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
saleniuk: Ok, it took me some time as I had to move some tests around and update them, but I changed and simplified it so that now it only observes ongoing calls for current user

AndroidBob:
Build [1007](https://10.10.124.134/job/AR-build-pipeline/1007/) **succeeded**.

- [com.wire.android-v4.3.0(7263818)(7263818)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.3.0(7263818)(7263818)-staging-compat.apk)

app/src/main/kotlin/com/wire/android/services/ServicesManager.kt
ohassine: Thank you for applying the changes

ohassine (approved):

alexandreferris (approved):

PR #2104, opened 2023-08-16
fix: security classification label [WPB-3715]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- fixed issue that classification label doesn't show when message input field is not open
- refactored security classification banner to have one single composable with view model where it can be reused in OtherUserProfile, OngoingCall and Conversation Screen.
- changed structure of showing banner in OngoingCallScreen by adding spacers to show calling options always in center with or without classification banner


AndroidBob:
Build [1119](https://10.10.124.134/job/AR-build-pipeline/1119/) **failed**.

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
```diff
@@ -310,29 +305,27 @@ private fun OngoingCallTopBar(
...
+    flipCamera: () -> Unit
 ) {
-    Column {
-        val topPadding = if (classificationType != SecurityClassificationType.NONE) {
-            dimensions().spacing8x
-        } else {
-            dimensions().spacing16x
-        }
+    Column(
+        modifier = Modifier.height(96.dp)
```
saleniuk: suggestion: to make sure it always fits, we should make `SecurityClassificationBanner` also have a strict height, right? And maybe we should extract this value to the `dimensions` to not be hardcoded directly in the code

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
```diff
@@ -310,29 +305,27 @@ private fun OngoingCallTopBar(
...
+    ) {
+        Spacer(modifier = Modifier.weight(1F))
         Row(
             horizontalArrangement = Arrangement.SpaceEvenly,
             verticalAlignment = Alignment.CenterVertically,
             modifier = Modifier
                 .fillMaxWidth()
-                .padding(top = topPadding)
+                .height(56.dp)
```
saleniuk: suggestion: same here - maybe take from `dimensions`

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationArgs.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.wire.kalium.logic.data.id.QualifiedID
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class SecurityClassificationArgs(
+    @SerialName("conversation_id") val conversationId: QualifiedID?,
+    @SerialName("user_id") val userId: QualifiedID?
+) : ScopedArgs {
+    override val key = "$ARGS_KEY:${conversationId?.toString().orEmpty()}${userId?.toString().orEmpty()}"
```
saleniuk: suggestion: maybe we should put some spacer in between conversationId and userId, like ":" or "_"

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    private var state by mutableStateOf(SecurityClassificationType.NONE)
+
+    override fun state(): SecurityClassificationType = state
+
+    init {
+        when {
+            conversationId != null -> fetchConversationClassificationType(conversationId)
+            userId != null -> fetchUserClassificationType(userId)
+            else -> appLogger.w("Either conversationId or userId should be provided to SecurityClassificationViewModel")
```
saleniuk: suggestion: maybe we should throw an exception if we don't have both conversationId and userId, because it's an illegal state

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationArgs.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import com.wire.android.di.ScopedArgs
+import com.wire.kalium.logic.data.id.QualifiedID
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class SecurityClassificationArgs(
+    @SerialName("conversation_id") val conversationId: QualifiedID?,
+    @SerialName("user_id") val userId: QualifiedID?
```
saleniuk, line 28: suggestion: it's required to have one of them, so maybe we should make an `Either` or disable default constructor and create custom ones to make sure we always create this class with one of them being not null?

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+        observeSecurityClassificationLabel.invoke(conversationId)
+            .collect { classificationType ->
+                state = classificationType
+            }
+    }
+
+    private fun fetchUserClassificationType(userId: UserId) = viewModelScope.launch {
+        state = getOtherUserSecurityClassificationLabel(userId)
+    }
+}
```
saleniuk: question: `getOtherUserSecurityClassificationLabel` just returns a single value and `observeSecurityClassificationLabel` returns a Flow - shouldn't both be unified? Either we get only first value for both cases or we observe for the value changes

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
```diff
@@ -310,29 +305,27 @@ private fun OngoingCallTopBar(
...
-    Column {
-        val topPadding = if (classificationType != SecurityClassificationType.NONE) {
-            dimensions().spacing8x
-        } else {
-            dimensions().spacing16x
-        }
+    Column(
+        modifier = Modifier.height(96.dp)
+    ) {
+        Spacer(modifier = Modifier.weight(1F))
```
saleniuk, line 322: question: won't this empty space be too big when the security banner is hidden?

default.json
```diff
@@ -14,7 +14,7 @@
             "logging_enabled": true,
             "application_is_private_build": true,
             "development_api_enabled": true,
-            "mls_support_enabled": true,
+            "mls_support_enabled": false,
```
saleniuk: question: is this change intentional? 😅

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationArgs.kt
saleniuk: question: also, what will happen if both are not null? 😄

default.json
Garzas: No, in federation squad we always forgot to change it back because without it we cannot use app 😅

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
Garzas: Thanks! I will use defaultSheetPeekHeight

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
Garzas: Ok

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationViewModel.kt
Garzas: I think warning is a better option because we are using it more for federation purposes and we may not catch exception before release 🤔

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationArgs.kt
Garzas: I think we cannot do that because both conversationId and userId have the same type 🤔 I can try with sealed classes 🤔

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
Garzas: I will discuss this with Wolfgang

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationViewModel.kt
Garzas: I can return Flow from GetOtherUserSecurityClassificationLabelUseCase, so new kalium PR 😄

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationArgs.kt
saleniuk: Ah, okay, then the other option could be to instead do something like:
```
val id: QualifiedID, // non-null
val type: Type.USER // Type.USER or Type.CONVERSATION
```
but it's fine to keep it as it is
😄

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationViewModel.kt
saleniuk: Ok, I guess it's not a big deal, because if both are null then this banner simply won't be visible, right? It's just that throwing an exception in that case would let us know that someone screwed up in the code faster 😄 
But it was just a suggestion 😉

app/src/main/kotlin/com/wire/android/ui/common/banner/SecurityClassificationViewModel.kt
saleniuk: I mean, it would be better to have Flow, but if it worked until now for the user with just taking single value, then we can keep it this way for now

AndroidBob:
Build [1158](https://10.10.124.134/job/AR-build-pipeline/1158/) **failed**.

AndroidBob:
Build [1159](https://10.10.124.134/job/AR-build-pipeline/1159/) **failed**.

AndroidBob:
Build [1160](https://10.10.124.134/job/AR-build-pipeline/1160/) **failed**.

saleniuk (approved):

AndroidBob:
Build [1168](https://10.10.124.134/job/AR-build-pipeline/1168/) **succeeded**.

- [com.wire.android-v4.4.0(7330851)(7330851)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.4.0(7330851)(7330851)-staging-compat.apk)

yamilmedina (approved):

AndroidBob:
Build [1171](https://10.10.124.134/job/AR-build-pipeline/1171/) **succeeded**.

- [com.wire.android-v4.4.0(7337769)(7337769)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.4.0(7337769)(7337769)-staging-compat.apk)

PR #2687, opened 2024-02-12
fix: NotificationChannelGroup crash (WPB-6233)

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6233" title="WPB-6233" target="_blank">BugWPB-6233</a>  [Android] Notification playstore crash with createNotificationChannel
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

crash:
https://play.google.com/console/u/2/developers/7098984309886892484/app/4973241010395499500/vitals/crashes/cb7ccba8c13d0a497df6edf45962bb4f/details?days=7&isUserPerceived=true

### Causes (Optional)

Logs says that app is trying to create a NotifcationChannel in a NotifcationChannelGroup that doesn't exist. This looks strange as we create Group first and only after it create channels in that group. 
So it's hard to reproduce the crash. 

From the logs in DataDog found out that at least some part of the crashes happens after logout. 
AND deleting the NotifcationChannelGroup is called from a few places (basically on logout) and launched in a separate CoroutineScope.

So there is a theory that when user logs out, `GlobalObserversManager.setUpNotifications` gets invalid data first (list of users with logged out user) and after some time - valid data. This makes the app create notification group and channels for the user who is logged out. And somewhere during this in another CoroutineScope `deleteNotificationChannelGroup`  is called. 

### Solutions

Move calling `deleteNotificationChannelGroup` into the same CoroutineScope to escape such conflicts. 
For that update `NotificationChannelsManager.createUserNotificationChannels`  to not just create Groups and Channels for "active" users, but also remove the Groups and Channels that don't belong to any "active" user.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
+++ b/app/src/main/kotlin/com/wire/android/GlobalObserversManager.kt
@@ -112,7 +112,6 @@ class GlobalObserversManager @Inject constructor(
             val callback: LogoutCallback = object : LogoutCallback {
                 override suspend fun invoke(userId: UserId, reason: LogoutReason) {
                     notificationManager.stopObservingOnLogout(userId)
-                    notificationChannelsManager.deleteChannelGroup(userId)
                     if (reason != LogoutReason.SELF_SOFT_LOGOUT) {
                         userDataStoreProvider.getOrCreate(userId).clear()
                     }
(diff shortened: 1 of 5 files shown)


saleniuk (approved):

AndroidBob:
Build [3064](https://10.10.124.134/job/AR-build-pipeline/3064/) **succeeded**.

- [com.wire.android-v4.6.0(8878435)(8878435)-staging-compat.apk](https://z-lohika.s3-eu-west-1.amazonaws.com/megazord/android/reloaded/staging/compat/com.wire.android-v4.6.0(8878435)(8878435)-staging-compat.apk)

MohamadJaara (approved):

PR #3026, opened 2024-05-22
fix: 1:1 calls can not be established on staging environment (WPB-9359)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9359" title="WPB-9359" target="_blank">BugWPB-9359</a>  [Android] 1:1 calls can not be established on staging environment
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When the user starts a call from android and answer the receiver answer, the call does not establish and stays on connecting state.

### Causes (Optional)

When we start the call, websocket closes immediately.
This happens after moving calling feature to a separate activity in separate task. The app thinks that it's on background when we start the call so it changes the connectivity policy to `DISCONNECT_AFTER_PENDING_EVENTS` which cause to close websocket and therefore no calling messages will be received to establish the call.

### Solutions

Observe lifecycle of `CallActivity` as well so we the app can correctly know if the app is visible or not (not just WireActivity as it was before)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


yamilmedina (approved):

app/src/main/kotlin/com/wire/android/ui/calling/CallActivity.kt
```diff
@@ -85,6 +89,7 @@ class CallActivity : AppCompatActivity() {
 
         setUpCallingFlags()
         setUpScreenShootPreventionFlag()
+        lifecycle.addObserver(currentScreenManager)
```
saleniuk, line 92: question: won't that create a race condition between `CallActivity` and `WireActivity` if both Activities lifecycles are observed at the same time by the `currentScreenManager`? Shouldn't we also `removeObserver` when opening and closing another Activity?

saleniuk: According to the documentation:
image
so first `onStart` from new Activity and then `onStop` from old one. 
We have `visibilityCount` there so it shouldn't have a negative impact in case of switching between activities, but still if we have multiple Activities now we should also remove these observers when Activity is destroyed. 
Also, logs from `CurrentScreenManager` will probably become hard to understand as they don't say which Activity is stopped or started, so when opening second Activity, in logs we will get `onStart` and then `onStop` so from logs it will look like app is not visible, but it's only the previous activity that has become stopped, not the whole app. 🤔

AndroidBob:
Build [4906](https://10.10.124.134/job/AR-build-pipeline/4906/) **failed**.

app/src/main/kotlin/com/wire/android/ui/calling/CallActivity.kt
ohassine: Actually there will be no race condition because visibilityCount is update when starting CallActivity `onStart` and it get updated `onStop` of WireActivity.
But yeah in logs it will not be clear which is which.

AndroidBob:
Build [4981](https://10.10.124.134/job/AR-build-pipeline/4981/) **failed**.

ohassine:
Closing it in favour of https://github.com/wireapp/wire-android/pull/3037

PR #3110, opened 2024-06-19
fix: missing remove device button [WPB-8819]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8819" title="WPB-8819" target="_blank">BugWPB-8819</a>  [Android] "Remove Device" button not showing in Beta app
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sometimes "remove device" button is missing for some old clients.

### Causes (Optional)

The `clientType` field could be null.

### Solutions

Use kalium fix to update this field properly on conflict.
Hide "remove device" button only for legal hold devices, so in case this field is null, button will be visible.
Fix loading states on `SelfDevicesScreen` and `RemoveDeviceScreen`.
Also, add handling errors on these screens when getting and fetching clients list, if some crucial data are missing then show dialog error with the option to retry.
Update `DeviceItem` UI element to match designs.

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/2821

#### Test Coverage (Optional)

#### How to Test

Open "my devices" screen or "remove device" screen (must reach the limit of registered clients).
To verify whether "remove device" button is not missing, you probably need to have the app version from before https://github.com/wireapp/kalium/pull/1544 and update it.

### Attachments (Optional)

| Before | After |
| before | after |
|   |  |
| <video width="400" src="https://github.com/wireapp/wire-android/assets/30429749/309bcdf1-a039-4ef4-9143-2bb6e9f47723"/> | <video width="400" src="https://github.com/wireapp/wire-android/assets/30429749/2b9f09dc-3a5f-4e00-bc7f-d6a8131b7485"/> |

Note: for these videos, a delay of 3s was added to show the difference in loading state.

----

----

----
##### References


vitorhugods:
> Note: for these videos, a delay of 3h was added to show the difference in loading state.

I envy your dedication 😅

app/src/test/kotlin/com/wire/android/ui/settings/devices/DeviceDetailsViewModelTest.kt
```diff
@@ -169,8 +167,8 @@ class DeviceDetailsViewModelTest {
                 arrangement.deleteClientUseCase.invoke(any()) wasNot Called
             }
             verify { arrangement.onSuccess wasNot Called }
-            assertTrue(viewModel.state.removeDeviceDialogState is RemoveDeviceDialogState.Visible)
-            assertTrue(viewModel.state.error is RemoveDeviceError.None)
+            assertEquals(true, viewModel.state.removeDeviceDialogState is RemoveDeviceDialogState.Visible)
```
vitorhugods: Better assertion errors if the tests break:
```suggestion
            assertInstanceOf(RemoveDeviceDialogState.Visible::class.java, viewModel.state.removeDeviceDialogState)
```

Maybe we could add kotlin-test to the project to get the Kotlin assertions like `assertIs<T>(value)`

app/src/test/kotlin/com/wire/android/ui/settings/devices/SelfDevicesViewModelTest.kt
```diff
@@ -44,19 +50,54 @@ import org.junit.jupiter.api.extension.ExtendWith
...
+    @Test
+    fun `given error, when retrying, then execute fetch properly`() = runTest {
+        val (arrangement, viewModel) = Arrangement()
+            .withFetchSelfClientsResult(SelfClientsResult.Failure.Generic(NetworkFailure.NoNetworkConnection(null)))
+            .arrange()
+        advanceUntilIdle()
+
+        viewModel.retryFetch()
+
+        coVerify(exactly = 2) { // first time during init, second time when retrying
```
vitorhugods: Could you please split it to make sure it is actually called the second time after the retry?

```kotlin
// ...
advanceUntilIdle()

coVerify(exactly = 1) { /** */ }

viewModel.retryFetch()

coVerify(exactly = 2) { /** */ } // Not sure if 1 or 2 here actually
```

app/src/main/kotlin/com/wire/android/ui/settings/devices/SelfDevicesViewModel.kt
```diff
@@ -53,37 +54,47 @@ class SelfDevicesViewModel @Inject constructor(
...
+                    currentDevice = result.clients
+                        .firstOrNull { it.id == currentClientId }
+                        ?.let { Device(it, e2eiCertificates[it.id.value]) },
+                    deviceList = result.clients
+                        .filter { it.id != currentClientId }
+                        .map { Device(it, e2eiCertificates[it.id.value]) }
+                )
+            }
+            is ObserveClientsByUserIdUseCase.Result.Failure -> {
+                state.copy(isLoadingClientsList = false, error = SelfDevicesState.Error.InitError)
```
MohamadJaara: If we have the clients list locally, there is no need to display the error message

app/src/main/kotlin/com/wire/android/ui/authentication/devices/remove/RemoveDeviceViewModel.kt
```diff
@@ -84,12 +84,14 @@ class RemoveDeviceViewModel @Inject constructor(
...
-            if (selfClientsResult is SelfClientsResult.Success) {
-                state = state.copy(
+            state = if (selfClientsResult is SelfClientsResult.Success) {
+                state.copy(
                     isLoadingClientsList = false,
                     deviceList = selfClientsResult.clients.filter { it.type == ClientType.Permanent }.map { Device(it) },
                     removeDeviceDialogState = RemoveDeviceDialogState.Hidden
                 )
+            } else {
+                state.copy(isLoadingClientsList = false, error = RemoveDeviceError.InitError)
```
MohamadJaara, line 94: In this case, I guess it is fine to display the error message since this is used directly after login and there is no local data, we just need to make sure the dialog is dismissible and the user can close it to take an action like logout

MohamadJaara (changes requested):

app/src/test/kotlin/com/wire/android/ui/settings/devices/DeviceDetailsViewModelTest.kt
saleniuk: Cool idea! For now I added this function in `TestUtil` so that it's accessible in `app.unitTest`

app/src/test/kotlin/com/wire/android/ui/settings/devices/SelfDevicesViewModelTest.kt
saleniuk: This test got removed because it's not needed after resolving comments from Mo 😅

app/src/main/kotlin/com/wire/android/ui/settings/devices/SelfDevicesViewModel.kt
saleniuk: I just thought it would be a nice additional safety, but maybe you're right. I removed these errors from self devices screen, just double checked and we do get full client data for self user with `Event.User.NewClient` and during slow sync (also initial slow sync) when updating self supported protocols. For very old devices, they cannot be legal hold ones as it's not yet used so we can quite safely assume that they are either permanent or temporary, both can be removed, so it's fine.

app/src/main/kotlin/com/wire/android/ui/authentication/devices/remove/RemoveDeviceViewModel.kt
saleniuk: Dialog is dismissible, there are two buttons: "retry" and "cancel" which executes the same thing as if the user pressed system back button or X on top bar - shows another dialog asking whether the user wants to cancel the action and logout.

saleniuk:
> > Note: for these videos, a delay of 3h was added to show the difference in loading state.
> 
> I envy your dedication 😅

and then I sped up the video to make the 3h last 3s 😜

MohamadJaara:
> > > Note: for these videos, a delay of 3h was added to show the difference in loading state.
> > 
> > 
> > I envy your dedication 😅
> 
> and then I sped up the video to make the 3h last 3s 😜


i-love-your-passion-for-this-business-manjit-minhas

vitorhugods (approved):

yamilmedina (approved):

MohamadJaara (approved):

PR #3127, opened 2024-06-25
fix: audio converting time [WPB-9705]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9705" title="WPB-9705" target="_blank">BugWPB-9705</a>  [Android] App freezes and delays when sending voice messages
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Converting audio file to m4a takes too long

### Causes (Optional)

Bad user experience waiting for audio to be encoded

### Solutions

- change audio encoding to wav
- introduce encoding state to be showed when user recorded long audio file

#### How to Test

1. Record long audio file (about 3mins)
2. After stop it should show Audio encoding message for few seconds
3. Send audio message

### Attachments (Optional)

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/sendmessage/SendMessageViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/sendmessage/SendMessageViewModel.kt
@@ -40,7 +40,7 @@ import com.wire.android.ui.home.messagecomposer.model.MessageBundle
 import com.wire.android.ui.home.messagecomposer.model.Ping
 import com.wire.android.ui.navArgs
 import com.wire.android.ui.sharing.SendMessagesSnackbarMessages
-import com.wire.android.util.AUDIO_MIME_TYPE
+import com.wire.android.util.SUPPORTED_AUDIO_MIME_TYPE
 import com.wire.android.util.ImageUtil
 import com.wire.android.util.dispatchers.DispatcherProvider
 import com.wire.android.util.getAudioLengthInMs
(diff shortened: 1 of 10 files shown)


saleniuk (approved):
Wav will be bigger in size than m4a, maybe it's not as big as to become a problem, but maybe we could think of compressing it not when recording or right after, but when sending - it won't be as problematic to the user as now. Also, maybe applying effect takes longer for wav than for m4a because of bigger file size, but I have no idea, just would be good to verify it. 😄 
Anyway, well done!

alexandreferris (approved):
Amazing work done here 👏🏻 

Generation of the file with effects is now super fast as compared to before 🏃🏻‍♂️ 

Tested on Android + iOS + Mac wrapper and everything is as it should be, awesome!

alexandreferris:
> Wav will be bigger in size than m4a, maybe it's not as big as to become a problem, but maybe we could think of compressing it not when recording or right after, but when sending - it won't be as problematic to the user as now. Also, maybe applying effect takes longer for wav than for m4a because of bigger file size, but I have no idea, just would be good to verify it. 😄 Anyway, well done!

The previous issue with `m4a` file was due to having to decode from `m4a` to `wav` and then from `wav` to `m4a` back again thus leaving us with a bit of delay when generating the file with effects, my personal opinion is that `wav` will still be faster than what/how we were doing before, but it would be good if we could run performance checks on both types 👌🏻

PR #3136, opened 2024-06-27
fix: location sharing without gms when not moving [WPB-9724]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9724" title="WPB-9724" target="_blank">BugWPB-9724</a>  [Android] Location sharing on graphene devices sometimes not working
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sometimes when we want to share the location on grapehene devices, the app is stuck on loading. 

### Causes (Optional)

On graphene there is no google services so the app uses the native manager to get the location, by requesting location updates. It works fine when the user is moving, but not when standing in the same place, probably because `requestLocationUpdates` emits items only when the location changes.

### Solutions

On newer Android versions use new way of getting location - `getCurrentLocation` and on old ones instead of `requestLocationUpdates`, use `requestSingleUpdate` because the app needs and uses only a single data anyway, doesn't need to listen for multiple changes.
Also, to make it even faster and more user friendly, first the last known location is checked and if it's not too old (currently if it's not older than 1m) then just use this last location because it looks like the user hasn't moved since then.
If the last location is too old, then just request the current one. There is also a timeout of 10s to get the current location, `getCurrentLocation` has some timeout anyway, but it's not configurable and it's unclear how big it is, so to make it unified a custom one is created - if the new location doesn't appear within 10s, then user gets the info that the app couldn't get the location.
Added tests for both `LocationPickerHelper` (to test getting location without google services on both new and old Android versions using robolectric) and `LocationPickerHelperFlavor` (to test getting location with google services using mockk).

#### Test Coverage (Optional)

#### How to Test

Open conversation screen and try to share location.

----

----

----
##### References


yamilmedina (approved):

yamilmedina:
I just realized these tests, were not in 4.6. But they are on release/candidate and up, so just raising as probably they will be conflicts to RC and Develop 🙈  I think if you have, just prefer these changes over whatever we have nowadays.

alexandreferris (approved):

PR #3138, opened 2024-06-27
fix: handle NoCommonProtocol error while starting conversation cherry-pick 

borichellow:
For some reason auto-cherry-pick didn't work, so i raised it by myself 
cherry-pick of https://github.com/wireapp/wire-android/pull/3120

# What's new in this PR?

### Issues

When user is trying to start 1o1 conversation with user that has no common protocol with, the error is no displayed anyhow.

### Causes (Optional)

Error handling in creating conversation was implemented only for NoKeypackages error. 

### Solutions

Add handling for all errors (specific texts for `NoKeypackages` and `NoCommonProtocol` errors and some default text for other errors.)

### Attachments (Optional)

Screenshot 2024-06-24 at 12 20 43


Garzas (approved):

yamilmedina (approved):

vitorhugods:
I see it failed with `MissingTranslation`. Maybe we should disable this check in the linter. We don't want to hardcode translations in English or something like that. It's preferable to keep it empty.

PR #3142, opened 2024-06-28
chore: Move Analytics to user-level (WPB-9584)

alexandreferris:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9584" title="WPB-9584" target="_blank">TaskWPB-9584</a>  [Android] Move Analytics Settings to User-Level
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Anonymous analytics was being initialized from a global-level (meaning there was no need of a valid session)

### Solutions

Move anonymous analytics initialization to user-level (user must be logged in and have a valid session)

#### Test Coverage (Optional)

- [X] I have added automated test to this contribution

#### How to Test

WIP


app/src/main/kotlin/com/wire/android/WireApplication.kt
```diff
@@ -71,6 +74,9 @@ class WireApplication : BaseApp() {
     @Inject
     lateinit var globalDataStore: Lazy<GlobalDataStore>
 
+    @Inject
+    lateinit var userDataStoreProvider: Lazy<UserDataStoreProvider>
```
MohamadJaara, line 78: injecting this here will not work when switching accounts, meaning the current user data store object will be injecting when starting the app, and never changed

MohamadJaara (changes requested):

app/src/main/kotlin/com/wire/android/WireApplication.kt
MohamadJaara: Alternatively you can try assisted injection to get the current user data store everytime

alexandreferris: We are not getting the `UserDataStore` directly but rather its provider to then later listen to active account and get its respective `UserDataStore`

edit: also discussed on 1:1

MohamadJaara: nvm this only the provider not the direct user data store object

app/src/main/kotlin/com/wire/android/WireApplication.kt
```diff
@@ -153,21 +159,27 @@ class WireApplication : BaseApp() {
...
+
+                AnonymousAnalyticsManagerImpl.init(
+                    context = this,
+                    analyticsSettings = analyticsSettings,
+                    isEnabledFlowProvider = userDataStore::isAnonymousUsageDataEnabled,
+                    anonymousAnalyticsRecorder = anonymousAnalyticsRecorder,
+                    dispatcher = Dispatchers.IO
+                )
+            }
+        }
```
MohamadJaara, line 184: does this need the check if if the build have analytics enabled?

MohamadJaara: before the check was inside AnonymousAnalyticsManagerImpl.init but now since we are doing some extra work with the user session maybe we can also check for it here

alexandreferris: The verification is done in gradle level, where if its enable we include whatever analytics we have, otherwise we will be using stub calls here, so `AnonymousAnalyticsManagerImpl.init(..)` will call "nothing" and return an `Unit`

Does this makes sense? as its done in gradle including libs, or shall we do another check here?

MohamadJaara: i mean to now run the whole check for current active user in case it is not enabled for the build

alexandreferris: Added the check for analytics enabled from build config

MohamadJaara (approved):

Garzas (approved):

alexandreferris (commented):
Note: Force merging this PR as the only failing action is from `smoke tests` and its due to an internal Jenkins error.

PR #3171, opened 2024-07-08
refactor: remove old date time parser APIs (WPB-9934)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9934" title="WPB-9934" target="_blank">TaskWPB-9934</a>  [Android] Replace SimpleDateFormat for modern APIs DateTimeFormatter or Instant
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, the application has an excessive use of `SimpleDateFormat` in reloaded, this is mainly used for formatting dates in messages (conversation screen), devices, read receipts, to name a few.

### Causes (Optional)

- Not Thread Safe.
- Creating instances is CPU intensive.
- Time Zones can mutate for instances.
- Slower performance according to benchmarks in devices, this translates ie, into increased battery consumption.

### Solutions

Upgrade to newer APIs, replacing:

- `SimpleDateFormat` for `DateTimeFormatter`
- `java.text` apis for `java.time` apis
- Deprecate old extensions functions formatters, but keep it to compare perf. we can remove later.
- Added instrumented tests for the old APIs vs new, to compare.

Some samples running in an emulator for `10_000_000` conversions has the following results.

Formatter | Time taken
-- | --
SimpleDateFormat | **2m 3.602497059s**
DateTimeFormat | 53.214024317s
Instant | 53.528359068s

#### Test Coverage (Optional)

----

----

----
##### References


app/src/main/kotlin/com/wire/android/util/DateAndTimeParsers.kt
```diff
@@ -0,0 +1,106 @@
...
+
+//region convenience ext functions, for retro compatibility
+fun String.deviceDateTimeFormat(): String? = DateAndTimeParsers.deviceDateTimeFormat(this)
+fun String.serverDate(): Date? = DateAndTimeParsers.serverDate(this)
+fun String.formatMediumDateTime(): String? = DateAndTimeParsers.formatMediumDateTime(this)
+fun String.formatFullDateShortTime(): String? = DateAndTimeParsers.formatFullDateShortTime(this)
+fun Instant.fileDateTime(): String = DateAndTimeParsers.fileDateTime(this)
+fun Instant.uiReadReceiptDateTime(): String = DateAndTimeParsers.uiReadReceiptDateTime(this)
+fun Date.toMediumOnlyDateTime(): String = DateAndTimeParsers.toMediumOnlyDateTime(this)
+fun String.uiMessageDateTime(): String? = DateAndTimeParsers.uiMessageDateTime(this)
```
MohamadJaara: are those functions stable? meaning for the same input they always have the same output, if so we can annotate them with Stable so they will not cause any unwanted recompression especially the one used in the UI

yamilmedina: Good point, I just migrated initially the old functions. 
We are using a couple on ui, but anyway they are idempotent, so added the `@Stable` marker if we decided to use it on ui.

app/src/main/kotlin/com/wire/android/util/DateAndTimeParsers.kt
```diff
@@ -0,0 +1,123 @@
...
+            } catch (e: Exception) {
+                null
+            }
+
+        fun formatFullDateShortTime(stringDate: String): String? =
+            try {
+                stringDate.serverDate()?.let { fullDateShortTimeFormatter.format(it.toInstant()) }
+            } catch (e: Exception) {
+                null
+            }
```
vitorhugods: I'd love to mark these as Deprecated as well.

Or with a similar warning, so this are seen as delicate APIs.

Assuming the data we receive through the API and database are immediately transformed to Instant, the app should transform _to_ String when showing in the UI, and not _from_ String at any other point.

app/src/test/kotlin/com/wire/android/util/DateTimeUtilKtTest.kt
```diff
@@ -18,124 +18,201 @@
...
+        @Test
+        fun `return a medium format date, when a valid date`() {
+            val result = "2022-03-24T18:02:30.360Z".deviceDateTimeFormat()
+            assertEquals("March 24, 2022, 6:02 PM", result)
+        }
+
+        @Test
+        fun `return a medium format, when formatMediumDateTime is called`() {
+            val result = "2022-03-24T18:02:30.360Z".formatMediumDateTime()
+            assertEquals("Mar 24, 2022, 6:02:30 PM", result)
```
vitorhugods: These tests fail if the JVM running them doesn't have `US` as the default locale. For instance, if I set my computer's language to German:

```
org.opentest4j.AssertionFailedError: expected: <March 24, 2022, 6:02 PM> but was: <24. März 2022, 18:02>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
	at com.wire.android.util.DateTimeUtilKtTest$DateAndTimeParsers.return a medium format date, when a valid date(DateTimeUtilKtTest.kt:47)
```

### Ideally

It would be nice to allow receiving the `Locale` as parameter in these functions:
Maybe leave `Locale.getDefault()` as default argument.
```kotlin
@Stable
fun String.deviceDateTimeFormat(locale: Locale = Locale.getDefault()): String? = DateAndTimeParsers.deviceDateTimeFormat(this)
```

This way we could write the tests like so:

```kotlin
        @Test
        fun `return a medium format date, when a valid date`() {
            val result = "2022-03-24T18:02:30.360Z".deviceDateTimeFormat(Locale.US)
            assertEquals("March 24, 2022, 6:02 PM", result)
        }
```

However, I see it would be annoying to start these DateTimeFormatter instances and keep them in memory.

### Shortcut

Maybe we can just set the default Locale before the test.

```kotlin
        @Test
        fun `return a medium format date, when a valid date`() {
            Locale.setDefault(Locale.US)
            val result = "2022-03-24T18:02:30.360Z".deviceDateTimeFormat()
            assertEquals("March 24, 2022, 6:02 PM", result)
        }
```

app/src/main/kotlin/com/wire/android/util/DateAndTimeParsers.kt
```diff
@@ -0,0 +1,123 @@
...
+fun String.serverDate(): Date? = DateAndTimeParsers.serverDate(this)
+
+@Stable
+fun String.deviceDateTimeFormat(): String? = DateAndTimeParsers.deviceDateTimeFormat(this)
+
+@Stable
+fun String.formatMediumDateTime(): String? = DateAndTimeParsers.formatMediumDateTime(this)
+
+@Stable
+fun String.formatFullDateShortTime(): String? = DateAndTimeParsers.formatFullDateShortTime(this)
```
vitorhugods: Same here. It would be nice to deprecate these.

vitorhugods (commented):
Nicely done.

Just a couple of small adjustments to make sure the test works in all machines, and further encourage dropping of `String` for internal date representation.

About dropping String, take the [`Device` class](https://github.com/wireapp/wire-android/blob/980213462d0e846b1174f245111422d45e6548a6/app/src/main/kotlin/com/wire/android/ui/authentication/devices/model/Device.kt#L37) for example. It has a `registrationTime`, which we transform from Instant to String.
And then we format this String _again_ in order to show it in the UI.

There are multiple places where we do this and we should drop it. Keep Instant all the way through until it's time to display in the UI.

app/src/test/kotlin/com/wire/android/util/DateTimeUtilKtTest.kt
yamilmedina: The  test you mention is not failing anymore 😅  it was solved, with the `@BeforeAll`, `@AfterAll` blocks to set the Locale, but funny enough, the old `java.text.DateFormat`, the implementation we are deprecating, is not respecting this, so even the shortcut does not works. 

The test failing is for retro compatibility 

```
return the same MediumOnlyDateTime format result, when calling date with new DateTimeFormatter format
```

yamilmedina: But I agree we should find a better way to be more reliable on the tests in all context, so I would love to have it passing the locale to the parser 👍

app/src/main/kotlin/com/wire/android/util/DateAndTimeParsers.kt
yamilmedina: Thanks for the comments, I think deprecating this was not part of the ticket, since we are focusing on removing the older APIs 🤔  and focusing in perf of SimpleDateFormat vs DateTimeFormatter (from java.time API)

yamilmedina: I think we can address this in another ticket/effort, unless you see another option.

vitorhugods (approved):

app/src/test/kotlin/com/wire/android/util/DateTimeUtilKtTest.kt
vitorhugods: Oooh indeed. I missed that in the tests. Didn't see the `@BeforeAll` and `@AfterAll` down there.

app/src/main/kotlin/com/wire/android/util/DateAndTimeParsers.kt
yamilmedina: After syncing, I Added anyway the deprecation warnings for each of one related to `strings`.

app/src/main/kotlin/com/wire/android/util/DateAndTimeParsers.kt
yamilmedina: After syncing, I Added anyway the deprecation warnings for each of one related to `strings`.

MohamadJaara (approved):
Great work as always 🔥

PR #3215, opened 2024-07-22
fix: user logged out listener [WPB-10114]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10114" title="WPB-10114" target="_blank">BugWPB-10114</a>  [Android] Alerts for forcing user to log out are not displayed anymore
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Listener responsible for observing logout state is triggered only once

### Causes (Optional)

When current user device get's removed from another client it doesn't show proper dialog and user is stuck on connecting

### Solutions

Instead of updating logout state dialog by side effect, move it to separate observer to listen only this change without taking only first event (`.first()`)

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
@@ -84,7 +84,6 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
```diff
@@ -206,6 +200,20 @@ class WireActivityViewModel @Inject constructor(
         }
     }
 
+    private fun observeLogoutState() {
+        viewModelScope.launch(dispatchers.io()) {
+            currentSessionFlow.get().invoke()
+                .distinctUntilChanged()
+                .collectLatest {
```
saleniuk: suggestion: if we want to handle all invalid sessions that are emitted then we shouldn't use `collectLatest` but regular `collect` 😉

saleniuk (approved):

ohassine (approved):

PR #3298, opened 2024-08-08
fix: Remove Reset Session for MLS conversations

borichellow:
# What's new in this PR?

### Issues

When we receive decryption errors in an MLS conversation, we show a “reset session” button.
But this button does not help to reset the session and is useless.

Therefore, we should hide it and not show it to the user.

### Causes (Optional)

Nobody think about differences in MLS conversations :) 

### Solutions

Think about it and remove "Reset Session" btn and text "Try resetting the session ..." text
+ small refactoring in Compose MessageItem to make it bit more readable

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationTopAppBar.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationTopAppBar.kt
@@ -250,7 +250,7 @@ fun PreviewConversationScreenTopAppBarLongTitle() {
             conversationName = UIText.DynamicString(
                 "This is some very very very very very very very very very very long conversation title"
             ),
-            conversationDetailsData = ConversationDetailsData.Group(QualifiedID("", "")),
+            conversationDetailsData = ConversationDetailsData.Group(null, QualifiedID("", "")),
             conversationAvatar = ConversationAvatar.OneOne(null, UserAvailabilityStatus.NONE),
         ),
         onBackButtonClick = {},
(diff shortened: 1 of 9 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItemComponents.kt
```diff
@@ -240,6 +242,9 @@ internal fun MessageDecryptionFailure(
                 text = stringResource(R.string.label_learn_more)
             )
             VerticalSpace.x4()
+
+            if (conversationProtocol !is Conversation.ProtocolInfo.Proteus) return@Column
```
borichellow, line 246: in fact this is the only changes in Compose, all others - just refactoring to make it more readable

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationTopAppBar.kt
```diff
@@ -250,7 +250,7 @@ fun PreviewConversationScreenTopAppBarLongTitle() {
             conversationName = UIText.DynamicString(
                 "This is some very very very very very very very very very very long conversation title"
             ),
-            conversationDetailsData = ConversationDetailsData.Group(QualifiedID("", "")),
+            conversationDetailsData = ConversationDetailsData.Group(null, QualifiedID("", "")),
```
alexandreferris: can we add named params? thought it could be easier to understand what we are passing as `null`

app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageItemComponents.kt
```diff
@@ -307,7 +312,12 @@ fun PreviewMessageSendFailureWarning() {
...
     WireTheme {
-        MessageDecryptionFailure(mockHeader, MessageFlowStatus.Failure.Decryption(false)) { _, _ -> }
+        MessageDecryptionFailure(
+            mockHeader,
+            MessageFlowStatus.Failure.Decryption(false),
+            { _, _ -> },
+            null
+        )
     }
 }
```
alexandreferris: can we have a preview for when the protocol is `Proteus` and the decryption failure is shown?

alexandreferris (commented):
Great catch on the issue 👌🏻 

Just added 2 comments for better readability on `ConversationTopAppBar` and if we can add one more preview with decryption error

alexandreferris (approved):

yamilmedina (approved):

PR #3421, opened 2024-09-06
fix: some files not visible in downloads folder [WPB-1854]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1854" title="WPB-1854" target="_blank">BugWPB-1854</a>  Files without mimetype are not shown in downloads when tapping "Show" after downloading
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When we receive a file without a mimetype that can be recognised by android, then it’s not shown in the downloads folder, to which we are redirected after tapping on “show”.

### Causes (Optional)

For Androids up to 10 the app uses deprecated `DownloadManager` to notify the system that there's a new file added to "downloads", but it required to pass non-null mime type, so there was a check and system was not notified about all files with null mime types. Also, even for files with proper mime types, the system could show that file after some time, it wasn't visible instantly.

### Solutions

For null mime types, just add general mime type `*/*` - it works.
For the case with files not being visible instantly, revert the order of actions on Androids up to 10 - on new ones, when `ContentResolver` is used, it's required to first notify the system by calling `insert` and then use the uri that's returned from that function to copy the file. Older approach for Androids up to 9 require the reverse order - first copy the file to the "downloads" folder and only after that, notify the system by calling `addCompletedDownload`.

#### Test Coverage (Optional)

#### How to Test

Save a downloaded asset into "downloads" and click "show" on snackbar to navigate to "downloads" folder.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/util/FileUtil.kt
+++ b/app/src/main/kotlin/com/wire/android/util/FileUtil.kt
@@ -117,7 +117,7 @@ private fun Context.saveFileDataToDownloadsFolder(assetName: String, downloadedD
                 /* title = */ availableAssetName,
                 /* description = */ availableAssetName,
                 /* isMediaScannerScannable = */ true,
-                /* mimeType = */ mimeType ?: "*/*",
+                /* mimeType = */ mimeType.orEmpty().ifEmpty { "*/*" },
                 /* path = */ destinationFile.absolutePath,
                 /* length = */ fileSize,
                 /* showNotification = */ false
--- a/app/src/main/kotlin/com/wire/android/util/FileUtil.kt
+++ b/app/src/main/kotlin/com/wire/android/util/FileUtil.kt
@@ -104,25 +104,26 @@ private fun Context.saveFileDataToDownloadsFolder(assetName: String, downloadedD
             put(MIME_TYPE, mimeType)
             put(SIZE, fileSize)
         }
-        resolver.insert(MediaStore.Downloads.EXTERNAL_CONTENT_URI, contentValues)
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/util/FileUtil.kt
```diff
@@ -104,25 +104,26 @@ private fun Context.saveFileDataToDownloadsFolder(assetName: String, downloadedD
...
-            val downloadManager = getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager
+        val downloadManager = getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager
+        FileProvider.getUriForFile(this, authority, destinationFile).also { downloadedUri ->
+            resolver.copyFile(downloadedUri, downloadedDataPath)
             downloadManager.addCompletedDownload(
                 /* title = */ availableAssetName,
                 /* description = */ availableAssetName,
                 /* isMediaScannerScannable = */ true,
-                /* mimeType = */ mimeType,
+                /* mimeType = */ mimeType ?: "*/*",
```
MohamadJaara: looking at the iOS code there is a chance that the mimeType is an empty string "" so maybe we can check for that here

saleniuk: Good idea! Added that check to the code

alexandreferris (approved):

Garzas (approved):

PR #3425, opened 2024-09-09
fix: empty dependencies section [WPB-10929]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10929" title="WPB-10929" target="_blank">BugWPB-10929</a>  [Android] App does not show dependency versions (core-crypto, avs) on Bund version
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

 App does not show dependency versions (core-crypto, avs).

### Causes (Optional)

The string with dependencies is remembered to keep it during recompositions, but without any key and the dependencies map is empty initially, so when the proper data appears, it takes the remembered empty value instead of creating new proper one with dependencies.

### Solutions

Use `dependencies` as a key when remembering string so that when dependencies change, the string is rebuilt with current data.
Also, `DependenciesItem` texts stylings and paddings are unified with all other debug screen elements.

#### How to Test

Open "Debug Settings" and check "Dependencies" section.

### Attachments (Optional)

| Before | After |
|  |  |

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
@@ -512,6 +512,6 @@ fun PreviewOtherDebugOptions() {
         handleE2EIEnrollmentResult = {},
         dismissCertificateDialog = {},
         checkCrlRevocationList = {},
-        dependenciesMap = persistentMapOf("avs" to "1.0.0", "core-crypto" to "1.0.0")
+        dependenciesMap = persistentMapOf("preview dependency" to "1.0.0", "another dependency" to "1.0.0")
     )
 }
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -512,6 +512,6 @@ fun PreviewOtherDebugOptions() {
         handleE2EIEnrollmentResult = {},
         dismissCertificateDialog = {},
         checkCrlRevocationList = {},
-        dependenciesMap = persistentMapOf()
+        dependenciesMap = persistentMapOf("avs" to "1.0.0", "core-crypto" to "1.0.0")
```
MohamadJaara: maybe we can have a default value of null just in case something is wrong it will not show 1.0.0 which is incorrect value

```suggestion
        dependenciesMap = persistentMapOf("avs" to "null", "core-crypto" to "null")
```

saleniuk: These values are only for a preview 😄 
I changed these names for some more generic ones to avoid confusion

MohamadJaara: ohhhhh my bad, I saw in the pr discretion mentioning empty state and thought it was changed here

MohamadJaara (approved):

alexandreferris (approved):

PR #3432, opened 2024-09-11
feat: Disable remote domain users for Proteus conversation [WPB-10788]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10788" title="WPB-10788" target="_blank">StoryWPB-10788</a>  [Android] Disable adding remote domain users in Proteus Conversations
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

We need to disable adding users from another domain into the Proteus conversation or if the team's default protocol is Proteus.

### Solutions

add `isOtherDomainAllowed` field into `AddMembersSearchNavArgs` and use it in `SearchUserViewModel` by passing to `FederatedSearchParser`. 
in updated kalium `FederatedSearchParser` accept new `isOtherDomainAllowed` and "do the magic".


yamilmedina (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/SearchUserViewModel.kt
```diff
@@ -82,7 +82,7 @@ class SearchUserViewModel @Inject constructor(
     }
 
     private suspend fun search(query: String) {
-        val (searchTerm, domain) = federatedSearchParser(query)
+        val (searchTerm, domain) = federatedSearchParser(query, addMembersSearchNavArgs?.isOtherDomainAllowed ?: true)
```
alexandreferris: Should really the default be true if `isOtherDomainAllowed` is null? 🤔

alexandreferris (commented):
Just a question 🤔

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/SearchUserViewModel.kt
borichellow: that question made me to think more about the implementation and I've noticed that all the `addMembersSearchNavArgs`  might be null here (in case of creating a new conversation).

So I've updated PR and moved all the `isOtherDomainAllowed` logic directly into `SearchUserViewModel`.

Answering your question: if addMembersSearchNavArgs is null we use only default team protocol to get if other domain allowed.

alexandreferris (approved):

saleniuk (approved):

PR #3461, opened 2024-09-23
fix: message item click and long click areas [WPB-9077]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9077" title="WPB-9077" target="_blank">BugWPB-9077</a>  [Android] Search result tap area within a conversation
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The area that I can click to navigate to that message in conversation is small, whole item area should be clickable and navigate to the message in conversation, but now for instance when I click on avatar, the ripple animation is visible but nothing happens.

### Solutions

- implemented a modifier to intercept clicks and long clicks on a level of `RegularMessageItem` composable, now no matter where user starts the long click it will always be intercepted and handled exactly the same for the whole message item including the ripple effect animation
- thanks to that, `onLongClick`s and `isContentClickable` flags could be removed from all message content elements
- grouped message click actions into two types to make it easier to use: 
  - `Content` - all content elements can be clicked separately and each has its own action
  - `FullItem` - the whole row is clickable and always executes a single, the same action
- moved handling message clicks to `RegularMessageItem` (system messages are not clickable anyway)
- updated `customizeMessageBackground` modifier to use single `Animatable` and make it so that it can make transitions between all possible types smoothly

#### How to Test

Search for a message in a conversation and try clicking on different areas of that message item.
The same for long clicking on any message type on conversation screen or conversation media files list.

### Attachments (Optional)

| Before | After |
| <video width="400" src="https://github.com/user-attachments/assets/c8e69c78-4f70-4533-ae9d-cdba2dafe21b"/> | <video width="400" src="https://github.com/user-attachments/assets/42b60301-368e-4809-8883-fd0467820c91"/> |

----

----

----
##### References


Garzas (approved):
👏

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/InterceptClickable.kt
```diff
@@ -0,0 +1,93 @@
...
+        awaitEachGesture {
+            val pass = PointerEventPass.Initial
+            val longPressTimeout = viewConfiguration.longPressTimeoutMillis
+
+            // wait for the first down press
+            val downEvent = awaitFirstDown(pass = pass)
+            val press = PressInteraction.Press(downEvent.position)
+
+            if (onClick != null) {
+                // if click needs to be intercepted then consume the down event
```
vitorhugods, line 49: What if it has both Click and Long Click?

Normal click interaction should only be triggered after `up`.

I think we should call `waitForUpOrCancellation` in any case, and measure the time (if it was shorter os longer than `longPressTimeout`).

saleniuk: Here it's not about triggering click, it's just about consuming the event so that it doesn't reach children and cause multiple triggers of the same event.
If it has both `onClick` and `onLongClick` then it should consume the event anyway because it means that children shouldn't get this event at all. 
Actual triggering the click is on the line 85 and it's done exactly as you explained, `waitForUpOrCancellation` is called in any case and if it has both `onClick` and `onLongClick` then it will make use of the `longPressTimeout` to decide which one should be triggered.

MohamadJaara:
Question: just thinking out loud did you test those changes with messages that have clickable test e.g. hyper links

vitorhugods (approved):

saleniuk:
> Question: just thinking out loud did you test those changes with messages that have clickable test e.g. hyper links

I tested that, but I haven't added a video to prove it, so here it is 😄 

https://github.com/user-attachments/assets/9bda2fd7-b98a-42fb-af99-e31e97b7fe68

PR #3465, opened 2024-09-24
fix: self-deleting asset message timer [WPB-1807]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1807" title="WPB-1807" target="_blank">StoryWPB-1807</a>  Self-deleting: Delete assets from internal storage once the timer runs out
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

For self-deleting message with asset, the timer doesn't start counting down until the user opens the asset.

### Causes (Optional)

We had mixed old and new logic when it comes to starting the countdown. First we implemented self-deleting messages with asset in a way that it only starts counting down when the user actually opens the asset. It was changed some time ago to unify it with other platforms, so now it should start counting down at the moment the message is shown to the user, no matter if the user actually opens the asset or not. The problem is that part of the old logic was still in the code and because of that the timer was not being shown until the user opened the asset.

### Solutions

Show the timer at the moment the message is being shown on the screen, so the logic is now the same for all types of self-deleting messages, no matter what content it has and no matter if the content has actually been opened or not.
Also, it includes new changes regarding deleting asset files when self-deleting message disappears (https://github.com/wireapp/kalium/pull/3027).

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/3027

#### How to Test

Send a self-deleting message with asset and check if the timer is clearly visible from the first moment the message appeared on the screen.

### Attachments (Optional)

| Before | After |
| <video src="https://github.com/user-attachments/assets/91603b9f-6abc-44ce-aa83-26ccbb4aa967"/> | <video src="https://github.com/user-attachments/assets/469e4e5d-dddc-48de-ac37-54073c5e9f84"/> |

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/MessageExpiration.kt
@@ -32,9 +32,6 @@ import androidx.lifecycle.repeatOnLifecycle
 import com.wire.android.R
 import com.wire.android.ui.home.conversations.model.ExpirationStatus
 import com.wire.android.ui.home.conversations.model.UIMessage
-import com.wire.android.ui.home.conversations.model.UIMessageContent
-import com.wire.kalium.logic.data.asset.AssetTransferStatus
-import com.wire.kalium.logic.data.asset.isSaved
 import com.wire.kalium.logic.data.message.Message
 import kotlinx.coroutines.delay
 import kotlinx.datetime.Clock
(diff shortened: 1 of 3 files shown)


vitorhugods (approved):

borichellow:
We apply the same delete logic to all messages. 
What if delete timer is `1m` and there is 2 minutes audio message??

borichellow (commented):
We apply the same delete logic to all messages.
What if delete timer is 1m and there is 2 minutes audio message??? User has no chance to listen the whole message ;P

saleniuk:
> We apply the same delete logic to all messages. What if delete timer is 1m and there is 2 minutes audio message??? User has no chance to listen the whole message ;P

It's the sender who decides to send a 2m long audio message knowing that he sends it as a 1m self-deleting message, it's not the app's fault 😄  
With this PR we just fully unified our logic to work the same as on iOS and web, moreover, it's been like that for a few months now with only difference that the timer visibility was broken 🙃

borichellow (approved):

PR #3518, opened 2024-10-16
fix: Crash during asset sharing  [WPB-10700]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10700" title="WPB-10700" target="_blank">BugWPB-10700</a>  [Android] file not found when sending assets
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Crash  from GooglePlay during sharing the asset.

### Causes (Optional)

Exact reason wasn't found, but idea is that `imageUtil.extractImageWidthAndHeight` may through it.

### Solutions

Wrap `imageUtil.extractImageWidthAndHeight` into try-catch and show error message to user in case of exception.


yamilmedina (commented):
LGTM 🚀 ! I think we need to port it or target `release/candidate`

ohassine:
I'd suggest to check if the file exists before sending the asset

borichellow:
> I'd suggest to check if the file exists before sending the asset

good point :)

app/src/main/kotlin/com/wire/android/ui/home/conversations/sendmessage/SendMessageViewModel.kt
```diff
@@ -290,22 +290,27 @@ class SendMessageViewModel @Inject constructor(
...
+                                    assetWidth = imgWidth,
+                                    assetHeight = imgHeight,
+                                    assetDataSize = dataSize,
+                                    assetMimeType = mimeType,
+                                    audioLengthInMs = 0L
+                                )
+                                    .handleLegalHoldFailureAfterSendingMessage(conversationId)
+                                    .handleAssetContributionEvent(assetType)
+                            } else {
+                                appLogger.e("There was an FileNotFoundException error while sending image asset")
```
yamilmedina:
```suggestion
                                appLogger.e("There was a FileNotFoundException error while sending image asset")
```

yamilmedina (approved):
Looks great! just added a small typo correction, but I leave it ✅ now

app/src/main/kotlin/com/wire/android/ui/home/conversations/sendmessage/SendMessageViewModel.kt
```diff
@@ -290,22 +290,27 @@ class SendMessageViewModel @Inject constructor(
...
+                                onSnackbarMessage(ConversationSnackbarMessages.ErrorSendingAsset)
```
damian-kaczmarek: This will display error: "There was an error trying to send that file. Please check your Internet connection and try again". Which is better then a crash but doesn't give right information to the user. It points them towards Internet issue while this is file system issue. Maybe a new message with "Couldn't read file" or something. I guess a design question.

We could also use `ErrorSendingImage` as it is a little more specific for the image. - "The picture could not be sent. Please check your internet connection and try again."
```suggestion
                                onSnackbarMessage(ConversationSnackbarMessages.ErrorSendingImage)
```

damian-kaczmarek (approved):

PR #3541, opened 2024-10-22
feat: Add accessibility strings to CreateGroup [WPB-9786]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9786" title="WPB-9786" target="_blank">TaskWPB-9786</a>  [Android] accessibility strings – Create group
  </td></table>
  <br />
 

# What's new in this PR?

Added Accessibility strings to create new group screens.


app/src/main/kotlin/com/wire/android/ui/common/textfield/WireTextFieldLayout.kt
```diff
@@ -101,8 +103,8 @@ internal fun WireTextFieldLayout(
                 .fillMaxWidth()
                 .background(color = colors.backgroundColor(state).value, shape = shape)
                 .border(width = dimensions().spacing1x, color = colors.borderColor(state, interactionSource).value, shape = shape)
-                .semantics {
-                    (labelText ?: placeholderText ?: descriptionText)?.let {
+                .semantics(mergeDescendants = true) {
+                    (semanticDescription ?: labelText ?: placeholderText ?: descriptionText)?.let {
```
MohamadJaara, line 113: tbh i am not fan of this implicant behaviour, i would expect if i sent `semanticDescription = null` to exclude an item from being readable it should not be readable

maybe add a doc comment to this function disrupting this behaviour?

borichellow: I got what you mean. 
Honestly: I can't imagine the case when we need empty description for TextField, but I added doc, describing how the fields are used and what to do if we need empty contentDescription :)

MohamadJaara (approved):

Garzas (approved):

PR #3545, opened 2024-10-23
chore: Change user search name surname highlight logic #WPB-6536

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6536" title="WPB-6536" target="_blank">BugWPB-6536</a>  [Android] Highlight missing at @ sign while searching for a user
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-6536

# What's new in this PR?

### Issues

When we're searching for users, we were not highlighting `@` 

### Solutions

We take into considerations the prefix, and now queries with `@` highlight properly. There is an edge case where our search for `@mat` would return results without `@` and we do not highlight anything, but I confirmed with Vina that it is fine in that case.

#### How to Test

Search for users with `@` and without, check if proper parts of name and surname are highlighted. In the task description there are also places where it shouldnt change after this changes

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/search/HighLightName.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/search/HighLightName.kt
@@ -40,14 +40,12 @@ fun HighlightName(
     modifier: Modifier = Modifier
 ) {
 
-    val queryWithoutSuffix = searchQuery.removeQueryPrefix()
-
     val highlightIndexes = QueryMatchExtractor.extractQueryMatchIndexes(
-        matchText = queryWithoutSuffix,
+        matchText = searchQuery,
         text = name
     )
 
-    if (queryWithoutSuffix != String.EMPTY && highlightIndexes.isNotEmpty()) {
+    if (searchQuery != String.EMPTY && highlightIndexes.isNotEmpty()) {
         Text(
             buildAnnotatedString {
                 withStyle(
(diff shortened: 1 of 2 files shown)


borichellow (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/HighLightName.kt
```diff
@@ -40,14 +40,12 @@ fun HighlightName(
     modifier: Modifier = Modifier
 ) {
 
-    val queryWithoutSuffix = searchQuery.removeQueryPrefix()
-
     val highlightIndexes = QueryMatchExtractor.extractQueryMatchIndexes(
-        matchText = queryWithoutSuffix,
+        matchText = searchQuery,
```
MohamadJaara, line 44: when the user add @ it means to do search based on user handle, adding it to the search query will result to handle search being done with @ in the beginning 
for example my is a handle is "mo_wirer" then this is what will be stored in the database and I need to do handle search then the user enter "@mo_wire" and the actual search in the database will be against "@mo_wire" which will result to no valid results

m-zagorski: Totally agree, but the logic for search itself didnt change, we're still doing it as you explained, so if `@` is present it will be done like that, as it was before, this changes only the way its highlighted, so if we're looking for `@mat` and got results like:
- mat mat - wont highlight anything
- @mateusz.123 - will highlight @mat

The same goes if the query was done without `@` so if we search for `mat` it would be:
- mat mat - both maths are going to be highlighted
-@mateusz.123 - `mat` will get highlighted

Those changes will not affect any search results in any case

saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/HighLightName.kt
MohamadJaara: ok got it

PR #3555, opened 2024-10-25
feat: filter conversation list [WPB-11638]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11638" title="WPB-11638" target="_blank">StoryWPB-11638</a>  [Android] User Story for empty conversations/filters
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- renamed `MenuBottomSheetItem` arguments to leading and trailing
- added conversation filter button and bottom sheet
- added empty conversation list for each filter
- showing favorite conversations will be handled in another PR

https://github.com/user-attachments/assets/1525de1a-0b61-4e51-825e-705446860e82


app/src/main/kotlin/com/wire/android/ui/home/conversationslist/filter/ConversationFilterSheetContent.kt
```diff
@@ -0,0 +1,69 @@
...
+                        trailing = {
+                            if (filter == currentFilter) {
+                                MenuItemIcon(
+                                    id = R.drawable.ic_check_circle,
+                                    contentDescription = stringResource(R.string.label_selected),
+                                    tint = MaterialTheme.wireColorScheme.switchEnabledChecked,
+                                )
+                            }
+                        },
+                        onItemClick = { onChangeFilter(filter) },
```
borichellow, line 56: can you please add `onItemClickDescription = stringResource(R.string.content_description_select_label)`

borichellow (commented):
Cool! Just a small wish to improve accessibility ;P

MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/all/AllConversationsScreen.kt
```diff
@@ -43,7 +44,52 @@ fun AllConversationsScreen(homeStateHolder: HomeStateHolder) {
             searchBarState = searchBarState,
             conversationsSource = ConversationsSource.MAIN,
             lazyListState = currentLazyListState,
-            emptyListContent = { AllConversationsEmptyContent() }
+            emptyListContent = { ConversationsEmptyContent(filter = ConversationFilter.ALL) }
+        )
+    }
+}
+
+@HomeNavGraph()
```
damian-kaczmarek: Unnecessary empty `()` here and some other places

app/src/main/kotlin/com/wire/android/navigation/HomeDestination.kt
```diff
@@ -24,28 +24,54 @@ import com.ramcosta.composedestinations.spec.Direction
 import com.wire.android.R
```
damian-kaczmarek, line 24: When I select the filter (e.g. fav or group) screen stays empty. I check and `emptyListContent` never gets triggered. In `ConversationsScreenContent` the lazyPagingItems contains 3 elements. All nulls (called  get on them)

image

app/src/main/kotlin/com/wire/android/navigation/HomeDestination.kt
```diff
@@ -83,7 +109,31 @@ sealed class HomeDestination(
...
+    }
+}
+
+fun HomeDestination.currentFilter(): ConversationFilter {
+    return when (this) {
+        HomeDestination.Conversations -> ConversationFilter.ALL
+        HomeDestination.Favorites -> ConversationFilter.FAVORITES
+        HomeDestination.Group -> ConversationFilter.GROUPS
+        HomeDestination.OneOnOne -> ConversationFilter.ONE_ON_ONE
+        HomeDestination.Archive,
```
damian-kaczmarek, line 124: It is not in the issue. But from consistent UX perspective, should we also have same "How to" for archive screen?

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -136,7 +141,7 @@ class ConversationListViewModelImpl @AssistedInject constructor(
 
     private val searchQueryFlow: MutableStateFlow<String> = MutableStateFlow("")
 
-    private val containsNewActivitiesSection = conversationsSource == ConversationsSource.MAIN
+    private val containsNewActivitiesSection = conversationsSource != ConversationsSource.ARCHIVE
```
damian-kaczmarek: This is true for now but can be dangerous if we add a new one. Any newly added type in enum will have this set to true. It will be hard to predict then that adding ConversationsSource can cause this. I would propose a more verbose but safer private function that will use `when` with explicit mention of each type. That way analyser will bring developer notice that they should decide if this should be true or false for a new type.

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/all/ConversationsEmptyContent.kt
```diff
@@ -0,0 +1,134 @@
...
+                style = MaterialTheme.wireTypography.body02.copy(
+                    textDecoration = TextDecoration.Underline,
+                    color = MaterialTheme.colorScheme.onBackground
+                ),
+                modifier = Modifier.clickable {
+                    CustomTabsHelper.launchUrl(context, supportUrl)
+                }
+            )
+        }
+        if (filter != ConversationFilter.FAVORITES) {
```
damian-kaczmarek: Can be else or guard pattern with `return@Column` after line 90 and remove if

damian-kaczmarek (changes requested):
Only real issue for me is that I can't see those screen when using filters

damian-kaczmarek (approved):

PR #3561, opened 2024-10-28
feat: Add accessibility string to ConversationList [WPB-9789]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9789" title="WPB-9789" target="_blank">TaskWPB-9789</a>  [Android] accessibility strings – Conversation list
  </td></table>
  <br />
 

# What's new in this PR?

Added accessibility strings for the ConversationsList (AKA Home) and ConnectionRequest screens


app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
```diff
@@ -123,7 +123,7 @@ fun ConnectionActionButton(
             leadingIcon = {
                 Icon(
                     painter = painterResource(id = R.drawable.ic_check_tick),
-                    contentDescription = stringResource(R.string.content_description_right_arrow),
+                    contentDescription = null,
```
damian-kaczmarek, line 126: Are these (and other in this PR) removal on contentDescription communicated to @CinnamonNinja. I think she mentioned that this causes trouble for testing.

damian-kaczmarek (approved):

Garzas (approved):

app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
borichellow: didn't communicated yet. Thanks for mentioning!
I'll talk (she's off RN). But i believe we could solve that problem in some other way later (some `testTag` or something :) )

PR #3564, opened 2024-10-29
fix: Location services availability crash #WPB-11304 

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11304" title="WPB-11304" target="_blank">BugWPB-11304</a>  [Android] C apps on ss4sk device crashing when sharing location
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-11304

# What's new in this PR?

### Issues

When location services were removed from device - checking for their availability is crashing the application

### Solutions

We need to try/catch the error.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/location/LocationPickerHelper.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/location/LocationPickerHelper.kt
@@ -27,6 +27,7 @@ import android.os.CancellationSignal
 import androidx.annotation.VisibleForTesting
 import androidx.core.location.LocationManagerCompat
 import com.wire.android.AppJsonStyledLogger
+import com.wire.android.appLogger
 import com.wire.android.di.ApplicationScope
 import com.wire.android.ui.home.appLock.CurrentTimestampProvider
 import com.wire.kalium.logger.KaliumLogLevel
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/location/LocationPickerHelper.kt
```diff
@@ -117,9 +118,18 @@ class LocationPickerHelper @Inject constructor(
...
-        return LocationManagerCompat.isLocationEnabled(locationManager)
+        return try {
+            val locationManager = context.getSystemService(Context.LOCATION_SERVICE) as LocationManager
+            LocationManagerCompat.isLocationEnabled(locationManager)
+        } catch (e: Exception) {
+            appLogger.i(
+                message = "Checking for location services availability failed",
+                throwable = e
+            )
+            false
```
MohamadJaara, line 131: returning false here will make it seems like location is only disabled when the whole API is missing, maybe returning an Either or an Enum to reflect the state is better for usability, what do you think @m-zagorski

m-zagorski: If we were to display different message for the end users it would make sense to have some kind of enum or something else here, but as long as we're about to rely on what we have currently (ie the message about location not available) it probably doesnt make sense.

MohamadJaara (approved):

saleniuk (approved):

PR #3566, opened 2024-10-29
fix: do not update draft initially when opening screen [WPB-11808]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11808" title="WPB-11808" target="_blank">BugWPB-11808</a>  [Android] app stalls when fetching a page of conversations from the local DB
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When opening conversation screen, draft message is being initially saved with initial empty value unnecessarily and then back with the actual draft if there is one. This notifies other queries which observe `MessageDraft` table unnecessarily.

### Solutions

Do not save draft text until user actually enters some text.

This is part of improvements for paginated conversation list loading time so it contains changes from kalium PR: https://github.com/wireapp/kalium/pull/3080

#### Test Coverage (Optional)

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionHolder.kt
```diff
@@ -111,6 +112,7 @@ class MessageCompositionHolder(
 
     suspend fun handleMessageTextUpdates() {
         snapshotFlow { messageTextState.text to messageTextState.selection }
+            .dropWhile { (messageText, _) -> messageText.isEmpty() }
```
Garzas, line 115: What will happen when you will type something and delete all chars?

saleniuk: `dropWhile`:

> Returns a flow containing all elements except first elements that satisfy the given predicate.

In this case it filters out all initial empty values until first non-empty one appears and after that it emits all changes normally.

MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionHolder.kt
vitorhugods: Not sure if @Garzas question was answered 😅

I guess his concern is that this new line prevents the draft from being erased, no? I kinda see the same here.

If the user drafts "hello", and then erases "ello", it will work fine: Draft = "h".

If the user now deletes the "h": nothing happens?

Garzas (approved):

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionHolder.kt
saleniuk: I thought I explained well enough how `dropWhile` works 😅 
You probably think that it filters out all empty items, but in reality it filters out only empty elements before any non-empty element appears and then it emits all items, even empty ones. 

The code inside `dropWhile` is this:
```
    var matched = false
    collect { value ->
        if (matched) {
            emit(value)
        } else if (!predicate(value)) {
            matched = true
            emit(value)
        }
    }
```
First item that doesn't meet the predicate changes the `matched` state and after that all items are emitted, no matter if they meet the predicate or not.

So maybe some example:
```
flowOf(1, 2, 3, 2, 1, 0)
      .dropWhile { it < 3 }
      .collect { println(it) }
```
This will emit: `3 2 1 0`

In this particular case, it's there to filter out initial empty values to not save empty draft unnecessarily, but if user enters something and removes it, it will save this cleared empty draft:
- "" (initially empty string) - `dropWhile` filters out
- "h" (user enters) - `dropWhile` notices that it's first non-empty string and passes it
- "" (user clears) - `dropWhile` already noticed and changed `matched` state so it emits this empty string

PR #3627, opened 2024-11-13
fix: blank lists or flickering on Home Destinations - wrong lazyListStates [WPB-14276]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14276" title="WPB-14276" target="_blank">BugWPB-14276</a>  [Android] Conversation screen is not visible when navigating between screens
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When we navigate around screens sometimes the conversations are not visible until we scroll the list manually.
It looks like there are three different scenarios:
-the full list disappears for a moment when navigating back
-the list blinks quickly and then again blinks quickly when scrolling
-the list doesn't blink but then it disappears when scrolling

### Causes (Optional)

Each scrollable screen from "home" graph in theory has its own `LazyListState`, but each view for its own purposes gets it through the `currentLazyListState`. Now, when app executes navigation and changes the `currentNavigationItem`, if the screen that's being closed still needs to recompose during the transition animation or quickly before that, then it again gets the `LazyListState` for itself using `currentLazyListState` but now it returns the `LazyListState` for the new screen that's being navigated to, because `currentNavigationItem` is already changed, so it recomposes with wrong `LazyListState`.

### Solutions

Make each screen in the "home" graph to always and only use its dedicated `LazyListState`.

Thanks to that, it's finally possible to make `HomeStateHolder` not being `remember`ed with key `currentNavigationItem` thus recreated each time the app navigates, but instead there is only one instance of it remembered and the same one is used in each screen from "home" graph.

#### How to Test

- Open navigation drawer
- Open settings 
- Open account details
- Get back to conversation list

### Attachments (Optional)

| Before | After |
| <video width="400" src="https://github.com/user-attachments/assets/f6dfc111-e47b-4e11-b045-61e46569f32a"/> | <video width="400" src="https://github.com/user-attachments/assets/f9ca84c6-a373-49b4-962c-08a2ac826338"/> |

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/HomeScreen.kt
@@ -305,7 +305,7 @@ fun HomeContent(
                         }
                     },
                     collapsingEnabled = !searchBarState.isSearchActive,
-                    contentLazyListState = homeStateHolder.currentLazyListState,
+                    contentLazyListState = homeStateHolder.lazyListStateFor(currentNavigationItem),
                     content = {
                         /**
                          * This "if" is a workaround, otherwise it can crash because of the SubcomposeLayout's nature.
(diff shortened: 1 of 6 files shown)


m-zagorski (approved):
works great!

yamilmedina (approved):

PR #3637, opened 2024-11-15
fix: hide LH indicators on conversations when self user is under LH [WPB-6391]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6391" title="WPB-6391" target="_blank">BugWPB-6391</a>  The red dot is visible next to the conversation name In the conversation list view for the member who is not under LH.
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When self user is under legal hold, then all conversations on their list have the legal hold indicator next to the name because of that, so it can confuse the user to think that legal hold indicator next to another user's name on 1-1 conversation means that this user is under legal hold.

### Solutions

When self user is under legal hold, don't show any legal hold indicator next to conversations, indicator on the self user avatar in top right corner is enough for the user to know about the legal hold.

This PR also includes changes for passing `PAGINATED_CONVERSATION_LIST_ENABLED` feature flag to the view model so that it's possible to test the view model, so temporarily commented tests for the pagination are already uncommented.

#### Test Coverage (Optional)

#### How to Test

Enable legal hold for the account and check the conversation list.

### Attachments (Optional)

| Before | After |
|  |  |

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -438,6 +451,19 @@ private fun ConversationsSource.toFilter(): ConversationFilter = when (this) {
     ConversationsSource.ONE_ON_ONE -> ConversationFilter.ONE_ON_ONE
 }
 
+private fun ConversationItem.hideIndicatorForSelfUserUnderLegalHold(selfUserLegalHoldStatus: LegalHoldStateForSelfUser) =
+    // if self user is under legal hold then we shouldn't show legal hold indicator next to every conversation
+    // the indication is shown in the header of the conversation list for self user in that case and it's enough
+    when (selfUserLegalHoldStatus) {
```
damian-kaczmarek, line 457: Change name of the field in ConnectionConversation. Now it is hard to understand what this value holds without reading this comment. Maybe like `OtherUserUnderLegalHold` or something like this

damian-kaczmarek (changes requested):
Don't know how to test the lega hold flow, so just code check

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
saleniuk: `selfUserLegalHoldStatus` keeps the selfuser's legal hold status, so the name is correct.
Unless you are referring to `ConversationItem.ConnectionConversation.isLegalHold` which is 2 lines below, but `isLegalHold` is the common field for every conversation type - it basically means "this conversation is under legal hold" so it's set to true if any of conversation members is under legal hold, so not only other user but also self. "Connection" is also a conversation type in our logic, that's why it also inherits this field, and actually in case of "connection" it's always false because we may not know the legal hold status until we try to connect, but it's there because each conversation type overrides `ConversationItem` which contains `abstract val isLegalHold`, so changing the name just for this one type would break inheritance and make it disunified. 
@damian-kaczmarek did I understand your comment correctly and is this explanation sufficient?

damian-kaczmarek: Yes comment is about the `isLegalHold`, sorry for wrong line. So if 
```set to true if any of conversation members is under legal hold```
then this is no longer true right? Because if I am under the legal hold this value will be false in that object. That what I mean. If I understand correctly we change here value to false to change UI behaviour, but we also disturbed state of the field as from
```set to true if any of conversation members is under legal hold```
changes to
```set to true if any of conversation members is under legal hold, unless you the one under the legal hold then it is false```
which sounds too complex for me. If you happen to use that field in some other context you will not know that as you will not look in all cases where tie field is used to find this comment.

damian-kaczmarek (changes requested):
Just so it pop up on the list when rerequested

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
saleniuk: Sorry, my bad, I confused the fields. In this particular case, `isLegalHold` is more like "should UI show the legal hold indicator for this conversation" as `ConversationItem` is used only for the UI, so it should reflect the visual representation of the conversation, we shouldn't put too much logic into it. It's used in conversation lists and only there, it's not meant to be used anywhere else. 
But now I get why it can be confusing, the name of the field is simply too generic. 🙇 I changed the name `isLegalHold` to `showLegalHoldIndicator` - is that ok?

damian-kaczmarek (approved):

m-zagorski (approved):

ohassine (approved):

PR #3687, opened 2024-11-29
feat: Add Mentions to draft messages #WPB-12062

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12062" title="WPB-12062" target="_blank">TaskWPB-12062</a>  [Android] add mentions to draft messages
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

As a user when I type a message that have a mention and i go back to the conversation later to send the draft it should be sent with the mention.

### Causes (Optional)

Mentions are saved in draft but never used.

### Solutions

Check if draft contains mentions and use it if so


app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
```diff
@@ -298,6 +298,16 @@ fun ConversationScreen(
...
+    // add mentions when selectedMentions is not empty in MessageDraft
+    LaunchedEffect(messageDraftViewModel.state.value.selectedMentions) {
+        val compositionState = messageDraftViewModel.state.value
+        if (compositionState.selectedMentions.isNotEmpty()) {
+            messageComposerStateHolder.messageCompositionHolder.value.setMentions(
+                compositionState.draftText,
+                compositionState.selectedMentions.map { it.intoMessageMention() }
+            )
+        }
+    }
```
MohamadJaara: mentions should be set with with the whole draft and not independent from the text it self, otherwise we can risk having the text and the mention out of sync causing out of bound exceptions

MohamadJaara (dismissed):

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageComposerStateHolder.kt
```diff
@@ -77,6 +68,23 @@ fun rememberMessageComposerStateHolder(
...
+
+    LaunchedEffect(draftMessageComposition.draftText) {
+        if (draftMessageComposition.draftText.isNotBlank()) {
+            messageTextFieldValue.value = messageTextFieldValue.value.copy(
+                text = draftMessageComposition.draftText,
+                selection = TextRange(draftMessageComposition.draftText.length) // Place cursor at the end of the new text
+            )
+        }
+
+        if (draftMessageComposition.selectedMentions.isNotEmpty()) {
```
saleniuk, line 80: question: shouldn't that have its own `LaunchedEffect` to be executed every time when `draftMessageComposition.selectedMentions` changes? 🤔

borichellow: that is not necessary so far, as `selectedMentions` is updated only with `draftText` (in `MessageDraftViewModel.loadMessageDraft()`)

borichellow: that is not necessary so far, as `selectedMentions` is updated **only** with `draftText` (in `MessageDraftViewModel.loadMessageDraft()`)

yamilmedina (approved):

MohamadJaara:
please only merge in 4.11

Garzas (approved):

PR #3756, opened 2024-12-18
feat: Track ended call segmentation [#WPB-14256]

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14256" title="WPB-14256" target="_blank">StoryWPB-14256</a>  [Android] Call end - Event and segmentation
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-14256

# What's new in this PR?

### Issues

Missing tracking for ended call.

### Solutions

Map `RecentlyEndedCallMetadata` into the analytics segmentation and send the event to countly

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/WireApplication.kt
+++ b/app/src/main/kotlin/com/wire/android/WireApplication.kt
@@ -32,6 +32,7 @@ import com.wire.android.datastore.UserDataStoreProvider
 import com.wire.android.debug.DatabaseProfilingManager
 import com.wire.android.di.ApplicationScope
 import com.wire.android.di.KaliumCoreLogic
+import com.wire.android.feature.analytics.AnonymousAnalyticsManager
 import com.wire.android.feature.analytics.AnonymousAnalyticsManagerImpl
 import com.wire.android.feature.analytics.AnonymousAnalyticsRecorderImpl
 import com.wire.android.feature.analytics.globalAnalyticsManager
@@ -48,10 +49,12 @@ import com.wire.kalium.logger.KaliumLogLevel
 import com.wire.kalium.logger.KaliumLogger
 import com.wire.kalium.logic.CoreLogger
 import com.wire.kalium.logic.CoreLogic
+import com.wire.kalium.logic.feature.session.CurrentSessionResult
 import dagger.Lazy
 import dagger.hilt.android.HiltAndroidApp
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
@@ -93,6 +96,9 @@ class WireApplication : BaseApp() {
     @Inject
     lateinit var databaseProfilingManager: DatabaseProfilingManager
 
+    @Inject
+    lateinit var analyticsManager: Lazy<AnonymousAnalyticsManager>
+
     override val workManagerConfiguration: Configuration
         get() = Configuration.Builder()
             .setWorkerFactory(wireWorkerFactory.get())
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
```diff
@@ -151,6 +156,16 @@ class WireActivityViewModel @Inject constructor(
         observeScreenshotCensoringConfigState()
         observeAppThemeState()
         observeLogoutState()
+        trackRecentlyEndedCall()
+    }
+
+    private fun trackRecentlyEndedCall() {
```
saleniuk: question: wouldn't be better to have it in `WireApplication` than `WireActivity`? I mean, we can have `CallActivity` started without `WireActivity`, then after ending that call it won't trigger this event, right? 🤔

m-zagorski: Good point, sure ! going to move it up, tomorrows morning

core/analytics/src/main/kotlin/com/wire/android/feature/analytics/model/AnalyticsEvent.kt
```diff
@@ -129,6 +146,45 @@ interface AnalyticsEvent {
...
+            } else {
+                "incoming"
+            }
+        }
+
+        private fun RecentlyEndedCallMetadata.toConversationType(): String {
+            return when (conversationDetails.conversationType) {
+                Conversation.Type.ONE_ON_ONE -> "one_to_one"
+                Conversation.Type.GROUP -> "group"
+                else -> "unknown"
```
caldrian: Is "unknown" specified anywhere?

m-zagorski: This `unknown` would happen only if the conversation type would be `OWN` or `CONNECTION_PENDING` to which we cant call, so it will never happen in fact, I will change it up to an error instead

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
m-zagorski: done

saleniuk (approved):

app/src/main/kotlin/com/wire/android/WireApplication.kt
```diff
@@ -121,6 +127,19 @@ class WireApplication : BaseApp() {
...
+
+            observeRecentlyEndedCall()
+        }
+    }
+
+    private suspend fun observeRecentlyEndedCall() {
+        coreLogic.get().getGlobalScope().session.currentSessionFlow().collectLatest { sessionResult ->
+            if (sessionResult is CurrentSessionResult.Success && sessionResult.accountInfo.isValid()) {
+                coreLogic.get().getSessionScope(sessionResult.accountInfo.userId).calls.observeRecentlyEndedCallMetadata()
+                    .collect { metadata ->
```
borichellow: I don't remember why exactly, but i remember we should avoid `collect{ flow.collect { } }` pattern. Much better to use `flatMap`, `flatMapLatest`...

borichellow (commented):
basically good, but have a small code-style suggestion :)

app/src/main/kotlin/com/wire/android/WireApplication.kt
m-zagorski: 🤔 could you try to get some more context on that? The logic that happens in the `collect` is non suspending, so I believe it doesnt make difference here, but there might be something that I miss :D

m-zagorski: ah I got you now, you are right! changing it up

m-zagorski: @borichellow you can check it now :)

borichellow (approved):

PR #3758, opened 2024-12-18
fix: reusing PagingData crash [WPB-15055][WPB-15079][WPB-15064]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15055" title="WPB-15055" target="_blank">BugWPB-15055</a>  [Android] Crash when opening connection request from paginated list
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There are crashes related to reusing `PagingData` flow when doing some actions on paginated conversation list screen, like accepting legal hold request, answering 1:1 call or opening pending connection request which can fetch user data.

### Causes (Optional)

Each action that modifies (or even puts the same exact data but just requires executing insert, update or delete query) a table that then affects paginated conversation list query, makes this query dirty and executes it again, which reloads the list. 
When using `combine`, it creates a new instance of flow, so if the `combine` is used after `PagingData` is created then if the data is changed and emitted by the flow that's combined, it will re-use the same `PagingData` and put it into a new flow, so `cachedIn` won't cache the flow properly and it will crash because `PagingData` cannot be used twice.

### Solutions

Move `combine` "above" the `PagingData` flow so that each time a new value is emitted, it will always create a new `PagingData` for the new flow and also `cachedIn` will be used correctly.
Simplify `showLoading` parameter to be a simple logical equation rather than a state. Make use of mocked `previewConversationFoldersFlow` with proper `LoadStates` to represent loaded data instead of using specific `initiallyLoaded` parameter.
Fix `rememberSearchbarState` so that it actually saves the `searchQueryTextState`.

#### Test Coverage (Optional)

#### How to Test

The crash was happening when opening pending connection request (most frequently), answering 1:1 call or accepting legal hold request, so these actions shouldn't crash the app. Also, after opening connection request or 1:1 conversation, if your or that user's data hasn't changed since the last fetch, then it shouldn't refresh the conversation list when navigating back.

### Attachments (Optional)

| Before | After |
| <video width="400" src="https://github.com/user-attachments/assets/aa1b5411-6da3-40f2-a6c7-f970b89cbac3"/> | <video width="400" src="https://github.com/user-attachments/assets/d5bb1370-e195-415b-88eb-a97585ff24b6"/> |
| <video width="400" src="https://github.com/user-attachments/assets/1c0fa88e-9d3b-42ec-aa3f-2fccc9816d7e"/> |   |

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
@@ -69,8 +69,8 @@ class SearchBarState(
             },
             restore = {
                 SearchBarState(
-                    isSearchActive = it[0] as Boolean,
-                    searchQueryTextState = it[1]?.let {
+                    isSearchActive = (it.getOrNull(0) as? Boolean) ?: false,
+                    searchQueryTextState = it.getOrNull(1)?.let {
                         with(TextFieldState.Saver) {
                             restore(it)
                         }
--- a/app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
@@ -80,4 +80,3 @@ class SearchBarState(
         )
     }
 }
-
--- a/app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
@@ -29,11 +29,12 @@ import androidx.compose.runtime.setValue
 
 @Composable
 fun rememberSearchbarState(
+    initialIsSearchActive: Boolean = false,
     searchQueryTextState: TextFieldState = rememberTextFieldState()
 ): SearchBarState = rememberSaveable(
-    saver = SearchBarState.saver(searchQueryTextState)
+    saver = SearchBarState.saver()
 ) {
-    SearchBarState(searchQueryTextState = searchQueryTextState)
+    SearchBarState(isSearchActive = initialIsSearchActive, searchQueryTextState = searchQueryTextState)
 }
 
 class SearchBarState(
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
```diff
@@ -174,94 +176,105 @@ class ConversationListViewModelImpl @AssistedInject constructor(
...
+    }
+
+    private fun observeSelfUserLegalHoldState() {
+        viewModelScope.launch {
+            observeLegalHoldStateForSelfUser()
+                .map { it is LegalHoldStateForSelfUser.Enabled }
+                .flowOn(dispatcher.io())
+                .collect { isSelfUserUnderLegalHoldFlow.emit(it) }
+        }
+    }
```
borichellow, line 238: Instead of having a separate `fun` with collect you can have:
```
val isSelfUserUnderLegalHoldFlow = observeLegalHoldStateForSelfUser()
    .map { it is LegalHoldStateForSelfUser.Enabled }
    .flowOn(dispatcher.io())
    .shareIn(viewModelScope, SharingStarted.WhileSubscribed(), 1)

```

borichellow (approved):
looks good and very nice catch! Just a small suggestion

app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
```diff
@@ -57,14 +58,23 @@ class SearchBarState(
...
+                        save(it.searchQueryTextState)
+                    }
+                )
             },
             restore = {
                 SearchBarState(
-                    isSearchActive = it[0],
-                    searchQueryTextState = searchQueryTextState
+                    isSearchActive = it[0] as Boolean,
+                    searchQueryTextState = it[1]?.let {
```
yamilmedina: Question: Are there chances of getting an `IndexOutOfBounds` here ? Asking just in case we can make it safer (for example using a Pair instead of a List?), otherwise it looks good to me.

yamilmedina (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationListViewModel.kt
saleniuk: No, because `observeLegalHoldStateForSelfUser()` is actually a suspend fun for some reason so it needs to be called from a coroutine or another suspend fun 🙃 
I could use `runBlocking` but that would probably slow down the loading of the conversation list

app/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchBarState.kt
saleniuk: In theory, if that is saved properly then it shouldn't make any problems when restoring and even Google does that without any checks but it's a nice suggestion to add that just to be sure in any case 😄 So I changed `it[1]` with `it.getOrNull(1)`

PR #3759, opened 2024-12-19
feat: send and receive in-call reactions [#WPB-14254]

sergeibakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14254" title="WPB-14254" target="_blank">StoryWPB-14254</a>  [Android] In-Call Reactions
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-14254
# What's new in this PR?

- Observe and display in-call reactions with animations
- Show recent reaction in participant tile
- New panel for selecting in-call reaction
- Updated style for call controls
- Camera flip button moved to participant tile

| Showing emojis | In-call reactions panel | Custom emojis | Camera flip button |
| ----------- | ------------ | ----------- | ------------ |
| reactions_1  | reactions_2 | reactions_3 | reactions_4

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
@@ -20,6 +20,7 @@ package com.wire.android.ui.calling
 
 import android.view.View
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateMapOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
 import androidx.lifecycle.ViewModel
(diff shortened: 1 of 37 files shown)


MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
```diff
@@ -279,8 +310,48 @@ class SharedCallingViewModel @AssistedInject constructor(
...
+
     @AssistedFactory
     interface Factory {
         fun create(conversationId: ConversationId): SharedCallingViewModel
     }
 }
+
+private fun List<UICallParticipant>.senderName(userId: QualifiedID) = firstOrNull { it.id.value == userId.value }?.name
+
+private fun <T> Flow<T>.withDelayAfterFirst(timeMillis: Long): Flow<T> = withIndex()
```
borichellow: can you move it to `com.wire.android.util.extension.Flow.kt` so It might be re-used somewhere else

app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
```diff
@@ -279,8 +310,48 @@ class SharedCallingViewModel @AssistedInject constructor(
         }
     }
 
+    private fun observeInCallReactions() {
+        observeInCallReactionsUseCase(conversationId).onEach { message ->
```
borichellow: why do you use `onEach` instead of `collect`?

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/fullscreen/FullScreenTile.kt
```diff
@@ -87,7 +89,7 @@ fun FullScreenTile(
                     .height(height)
                     .padding(contentPadding),
                 participantTitleState = it,
-                isSelfUser = selectedParticipant.isSelfUser,
+                isSelfClient = selectedParticipant.clientId == participants[0].clientId,
```
borichellow: why did you change it? Os there any reason to not use `SelectedParticipant.isSelfUser`?
1. in case of design changes and self user is not always in a first tile we'll have to make changes here too.
2. It's preferred to as less logic in UI as possible.

app/src/main/kotlin/com/wire/android/ui/calling/controlbuttons/HangUpOngoingButton.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.wire.android.R
+import com.wire.android.ui.common.button.WireButtonState
+import com.wire.android.ui.common.button.WirePrimaryIconButton
+import com.wire.android.ui.common.dimensions
+import com.wire.android.ui.theme.WireTheme
+import com.wire.android.ui.theme.wireDimensions
+import com.wire.android.util.ui.PreviewMultipleThemes
+
+@Composable
+fun HangUpOngoingButton(
```
borichellow, line 38: Is it used somewhere? Can't find

sergeibakhtiarov: It is used in the `OngoingCallScreen` for showing call controls. Needed to create a separate button because it's shape is now different for ongoing call screen.

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/fullscreen/FullScreenTile.kt
sergeibakhtiarov: When you join the same call from two devices then you get two participants with `isSelfUser` flag set to true. So these users have incorrectly duplicated mute and flip camera buttons and video is only showing for one of them.
Needed to use a clientId to handle them correctly.

app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
sergeibakhtiarov: Just a matter of habit. I will change to `collect()` since it seems to be more common in the code.

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/fullscreen/FullScreenTile.kt
borichellow: so maybe lets update `SelectedParticipant` by changing `isSelfUser` to `isSelfClient` or just adding `isSelfClient` and using it from there?

sergeibakhtiarov: I am trying to fetch the clientId in the viewmodel. I think to make it correct we can add 'isSelfUser' flag to the UICallParticipant and use clientId to set it when mapping to UIParticipant. This way the flag will be correct across all UI code and will not depend on the participants ordering logic.

sergeibakhtiarov: @borichellow please check updated logic for isSelfUser based on clientId.

borichellow (approved):

PR #3776, opened 2024-12-30
fix: show no connection banner faster [#WPB-11568]

sergeibakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11568" title="WPB-11568" target="_blank">SpikeWPB-11568</a>  [Android] Investigate where the exponential backoff for no network connection can be short-circuited
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-11568
----
# What's new in this PR?
Showing the No Connection banner instantly when Internet connection is lost.

### Issues
After turning on Airplane mode the 'Waiting for connection' banner may be shown with a long delay.

### Causes
Displaying the banner is based on observed sync status which is set to Failed when WebSocket throws exception on connection loss. The problem is that the WebSocket is not guaranteed to detect the connection loss immediately and sometimes it throws exception only after the failed ping attempt (in 20 sec).

### Solutions
Observe network state in addition to sync state to detect faster the 'no connection' situation and show the banner instantly.


ohassine (approved):

yamilmedina (approved):
Good idea to solve this issue ! 💯

PR #3792, opened 2025-01-09
fix: Conversations and Profile Views are reported individually (WPB-14942)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14942" title="WPB-14942" target="_blank">BugWPB-14942</a>  [Android] Conversations and Profile Views are reported individually
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Views for conversations are reported to Countly with a unique conversation ID attached to them: Conversation(eed94f9*)
For some screen we report to countly SomeOther with screen name as route: SomeOther(route=other_user_profile_screen)

### Solutions

Check if current screen is SomeOther then we send the route otherwise we send class name

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


app/src/main/kotlin/com/wire/android/util/CurrentScreenManager.kt
```diff
@@ -123,17 +123,26 @@ class CurrentScreenManager @Inject constructor(
     }
 
     override fun onDestinationChanged(controller: NavController, destination: NavDestination, arguments: Bundle?) {
-        val currentView = currentScreenState.value.toString()
-        AnonymousAnalyticsManagerImpl.stopView(currentView)
+        val currentView = currentScreenState.value
+        handleViewAction(currentView) { screenName ->
```
sbakhtiarov: Small thing, but it was not obvious what this code is doing from the first look. It seems like `handleViewAction()` fun is suppose to handle something.
Maybe we can have a function to get current screen name from state:
```
    private fun currentScreenName() = currentScreenState.value.let { currentScreen ->
        (currentScreen as? CurrentScreen.SomeOther)?.route ?: currentScreen.javaClass.simpleName
    }
```
Then it can be used like this:
```
AnonymousAnalyticsManagerImpl.stopView(currentScreenName())
```

sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/util/CurrentScreenManager.kt
ohassine: Good point

app/src/main/kotlin/com/wire/android/util/CurrentScreenManager.kt
```diff
@@ -123,17 +123,22 @@ class CurrentScreenManager @Inject constructor(
...
         )
 
-        val newView = currentScreenState.value.toString()
-        AnonymousAnalyticsManagerImpl.recordView(newView)
+        val newScreenName = currentScreenName()
+        AnonymousAnalyticsManagerImpl.recordView(newScreenName)
+    }
+
+    private fun currentScreenName() = currentScreenState.value.let { currentScreen ->
+        (currentScreen as? CurrentScreen.SomeOther)?.route ?: currentScreen.javaClass.simpleName
```
MohamadJaara: whould "currentScreen.javaClass.simpleName" return an obfuscated name in release builds?

yamilmedina: Question: I might be wrong, but this will use reflection, and we should try to avoid it in Android or at least we might fallback to another option. Maybe is not a concern, but this will run constantly when Analytics enabled.

yamilmedina (commented):
Looking good, left a question/concern and also to add, for this to work we need to remove this flag

https://github.com/wireapp/wire-android/blob/a59ff2e0421c294ba9b41d0c178a1b07d711c060/core/analytics-enabled/src/main/kotlin/com/wire/android/feature/analytics/AnonymousAnalyticsManagerImpl.kt#L46

app/src/main/kotlin/com/wire/android/util/CurrentScreenManager.kt
ohassine: Yeah totally missed that, I changed the implementation

ohassine: You are right it may affect app performance.. I changed the implementation

ohassine:
> Looking good, left a question/concern and also to add, for this to work we need to remove this flag
> 
> https://github.com/wireapp/wire-android/blob/a59ff2e0421c294ba9b41d0c178a1b07d711c060/core/analytics-enabled/src/main/kotlin/com/wire/android/feature/analytics/AnonymousAnalyticsManagerImpl.kt#L46

Yes, we still don't need it yet.

app/src/main/kotlin/com/wire/android/util/CurrentScreenManager.kt
```diff
@@ -123,17 +123,34 @@ class CurrentScreenManager @Inject constructor(
...
+
+    private fun currentScreenName() = currentScreenState.value.let { currentScreen ->
+        when (currentScreen) {
+            is CurrentScreen.Home,
+            is CurrentScreen.Conversation,
+            is CurrentScreen.OtherUserProfile,
+            is CurrentScreen.ImportMedia,
+            is CurrentScreen.DeviceManager -> return@let currentScreen.toScreenName()
+
+            else -> return@let (currentScreen as? CurrentScreen.SomeOther)?.route
```
MohamadJaara: lets never use route since it may contain sensitive informations (e.g. ids) 
we need to find different solutions like having this defined in the Destination like using the base route?

MohamadJaara (dismissed):

app/src/main/kotlin/com/wire/android/util/CurrentScreenManager.kt
```diff
@@ -123,17 +124,34 @@ class CurrentScreenManager @Inject constructor(
...
+
+    private fun currentScreenName() = currentScreenState.value.let { currentScreen ->
+        when (currentScreen) {
+            is CurrentScreen.Home,
+            is CurrentScreen.Conversation,
+            is CurrentScreen.OtherUserProfile,
+            is CurrentScreen.ImportMedia,
+            is CurrentScreen.DeviceManager -> return@let currentScreen.toScreenName()
+
+            else -> return@let (currentScreen as? CurrentScreen.SomeOther)?.route?.getBaseRoute()
```
saleniuk: question: what about `AuthRelated` type? It will be ignored in this case because `else` tries to map it to `SomeOther`

ohassine: addressed

borichellow (approved):

core/analytics-enabled/src/main/kotlin/com/wire/android/feature/analytics/AnonymousAnalyticsManagerImpl.kt
```diff
@@ -182,7 +182,7 @@ object AnonymousAnalyticsManagerImpl : AnonymousAnalyticsManager {
         coroutineScope.launch {
             mutex.withLock {
                 if (!isAnonymousUsageDataEnabled) return@withLock
-                anonymousAnalyticsRecorder?.recordView(screen)
+                anonymousAnalyticsRecorder?.recordView(screen.convertToCamelCase())
```
MohamadJaara, line 185: is it really needed to convert this to camel case for countly ?
i mean is the screen names defined there or we can define them as we need on android ?

ohassine: it's for consistency, so we have clean data on the board. Otherwise some screen names are snake_case and some others are in camelCase.

PR #3801, opened 2025-01-14
fix: refactoring self user fetch [#WPB-15190]

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15190" title="WPB-15190" target="_blank">TaskWPB-15190</a>  [Android] race condition in getKnownUser causes multible requests to update
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-15190
----
# What's new in this PR?

Related to: https://github.com/wireapp/kalium/pull/3229

Refactoring getting / observe self user. New `GetSelfUserUseCase` use case to get self user is used where appropriate instead of observing until first value is received.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
@@ -37,7 +37,7 @@ import com.wire.kalium.logic.feature.client.NeedsToRegisterClientUseCase
 import com.wire.kalium.logic.feature.legalhold.LegalHoldStateForSelfUser
 import com.wire.kalium.logic.feature.legalhold.ObserveLegalHoldStateForSelfUserUseCase
 import com.wire.kalium.logic.feature.personaltoteamaccount.CanMigrateFromPersonalToTeamUseCase
-import com.wire.kalium.logic.feature.user.GetSelfUserUseCase
+import com.wire.kalium.logic.feature.user.ObserveSelfUserUseCase
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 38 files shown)


app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
```diff
@@ -97,15 +97,15 @@ class HomeViewModel @Inject constructor(
...
+            val userId = observeSelf().first().id
             when {
                 shouldTriggerMigrationForUser(userId) ->
                     onRequirement(HomeRequirement.Migration(userId))
 
                 needsToRegisterClient() -> // check if the client has been registered and open the proper screen if not
                     onRequirement(HomeRequirement.RegisterDevice)
 
-                getSelf().first().handle.isNullOrEmpty() -> // check if the user handle has been set and open the proper screen if not
+                observeSelf().first().handle.isNullOrEmpty() -> // check if the user handle has been set and open the proper screen if not
```
MohamadJaara: here we can get the ID and handle from first call in line 100

sbakhtiarov: true, will update it. 
This line `100` here requesting a non-null user id via a flow shows the main usage pattern of this use case.
Ideally we should replace it with `GetSelfUserUseCase' and handle null value as error on the ViewModel layer.

app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
```diff
@@ -97,15 +97,15 @@ class HomeViewModel @Inject constructor(
 
     fun checkRequirements(onRequirement: (HomeRequirement) -> Unit) {
         viewModelScope.launch {
-            val userId = getSelf().first().id
+            val userId = observeSelf().first()
```
saleniuk: suggestion: it's not only a `userId` anymore, maybe change the name to `user` or `selfUser`

saleniuk (approved):
one small suggestion but I'm approving as the rest looks ok

MohamadJaara (approved):

PR #3803, opened 2025-01-15
feat: remove conversation from folder [WPB-14630]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14630" title="WPB-14630" target="_blank">BugWPB-14630</a>  [Android] Conversation folder - Remove from folder
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- remove conversation from folder in Conversation List Screen and Conversation Details Screen

### Attachments (Optional)

https://github.com/user-attachments/assets/47818b25-91ca-4335-ad6d-abb816c3e945


MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationsScreenContent.kt
```diff
@@ -108,6 +111,10 @@ fun ConversationsScreenContent(
         hiltViewModelScoped<ChangeConversationFavoriteVMImpl, ChangeConversationFavoriteVM, ChangeConversationFavoriteStateArgs>(
             ChangeConversationFavoriteStateArgs
         ),
+    removeConversationFromFolderViewModel: RemoveConversationFromFolderVM =
+        hiltViewModelScoped<RemoveConversationFromFolderVMImpl, RemoveConversationFromFolderVM, RemoveConversationFromFolderArgs>(
```
saleniuk, line 115: question: does it need to be scoped? We provide all arguments when calling `removeFromFolder` function and don't provide any args that should distinguish instances, so there could even be a single instance of this view model for all `conversationsSource`s, so using regular `hiltViewModel` should be enough, at least in theory but maybe I'm missing something 🤔

Garzas: It doesn't need to be scoped but thanks to that we don't need to pass preview view model when we want to generate preview somewhere higher in composable tree

saleniuk (approved):

PR #3817, opened 2025-01-22
feat: bring back current login flow and add a feature flag for enterprise login [WPB-15543]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15543" title="WPB-15543" target="_blank">TaskWPB-15543</a>  [Android] Add feature flag and keep current flow intact until switching to new one
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We started implementing new enterprise login flow by modifying files from current one, but to make sure we keep having working login flow we need to keep current one intact and implement new one next to it so that it will be easy to switch to new one once everything is implemented and tested. 
Classes from current login flow: `WelcomeScreen`,  `WelcomeScreenState`, `LoginScreen` are reverted to their original forms, `LoginErrorDialog` is extracted to separate file so that it can be reused more easily.
Feature flag `enterprise_login_enabled` is created in this PR to be able to switch between old and new login flow.

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/newauthentication/welcome/NewWelcomeScreen.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.wire.android.util.ui.PreviewMultipleThemes
+import com.wire.kalium.logic.configuration.server.ServerConfig
+import kotlinx.coroutines.delay
+
+@RootNavGraph(start = true)
+@WireDestination(
+    navArgsDelegate = LoginNavArgs::class
+)
+@Composable
+// this is a temporary solution because annotation argument "start" must be a compile-time constant
```
yamilmedina, line 58: Good trick ! 

Question: can we maintain this in the future ? Mainly because we will need to resolve this not by compile time, but with API version based. If the backend is v8+ then new login flow. If not, the old/current flow.

yamilmedina (approved):
Great job!! looks good to me and can make us progress without waiting for BE. 

Just raised a question, if we can make the initial destination dynamic as it is now, but based on the API version instead of compile time. Not to solve now, but problem for the near future.

app/src/main/kotlin/com/wire/android/ui/newauthentication/welcome/NewWelcomeScreen.kt
saleniuk: Oh, so in short: yes, we can do that!

In that case we would need to create a small dedicated `ViewModel` that uses `AuthServerConfigProvider` and inject it into composable to determine the login flow version to use and this way we create a screen destination that works as a login flow "chooser", but we would need to create such composable "chooser" for the "welcome" and "login" screen (sometimes we directly open "login" screen, e.g. when migration requires the user to log in).

borichellow (approved):

PR #3852, opened 2025-01-31
fix: disable import media during slow sync [WPB-15790]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15790" title="WPB-15790" target="_blank">BugWPB-15790</a>  [Android] Client created a second conversation when sharing an image through share extension
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
- There is a race condition where sharing an image via the share extension during slow sync could send message to already migrated 1:1 conversation.

### Causes (Optional)
- The temporary inconsistency in conversation resolution led to a state where multiple 1:1 conversations were displayed across different clients.
- Eventually, the duplicate conversation was removed after a delayed sync resolution, but it caused confusion and message loss.

### Solutions
- Introduced `SyncStatusViewModel` to track the sync state and prevent sending media when slow sync is in progress.
- Disabled the send button (`loading` state) until slow sync completes, ensuring that messages are only sent to a fully resolved conversation.

### Attachments (Optional)

Screenshot 2025-01-31 at 13 50 50

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/common/sync/SyncStatusViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/sync/SyncStatusViewModel.kt
@@ -1,0 +1,76 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


saleniuk:
There is already a mechanism for blocking button click action while the app is syncing or/and connecting, it's called `ClickBlockParams` and you just pass it as a button parameter. I can see that this button already has this enabled to block when syncing and when connecting
```
clickBlockParams = ClickBlockParams(blockWhenSyncing = true, blockWhenConnecting = true),
```
 so it should act like that already if nothing is broken there but I can see that it's working, so changes from this PR are likely to be redundant 🤔

PR #3856, opened 2025-02-04
fix: Crash on inserting mention #WPB-15717

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15717" title="WPB-15717" target="_blank">BugWPB-15717</a>  [Android] play store crashes when inserting a mention
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Crash in PlayStore while inserting mention. 

### Causes (Optional)

Hard to understand exact STR, but the crash happens when mentions place in text can't be found. In that case mention's position is `-1` and crash come up in `String.subSequence(0, -1)`

### Solutions

Added checking if mention position and length are good, do not add mention otherwise and added logs.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionHolder.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionHolder.kt
@@ -24,6 +24,7 @@ import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.text.TextRange
 import androidx.compose.ui.text.substring
+import com.wire.android.appLogger
 import com.wire.android.ui.home.conversations.model.UIMention
 import com.wire.android.ui.home.conversations.model.UIMessage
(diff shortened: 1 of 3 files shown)


saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionHolder.kt
```diff
@@ -200,6 +201,10 @@ class MessageCompositionHolder(
             userId = UserId(contact.id, contact.domain),
             handler = String.MENTION_SYMBOL + contact.name
         )
+        if (mention.start < 0 || mention.length < 2) {
```
yamilmedina, line 204: Suggestion: Perhaps we can add tests for this case, I see we have a few on `MessageCompositionHolderTest` :D

yamilmedina (commented):
It would be nice to add tests for this case, apart from that good catch 🦅  :D

Garzas (approved):

PR #3864, opened 2025-02-07
feat: Disable Performance tracking #WPB-15791

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15791" title="WPB-15791" target="_blank">TaskWPB-15791</a>  [Android] Disable Performance tracking
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Due to to the high consumption of data points (reports send) the performance tracking (Application Performance Monitoring; APM) should be removed. In addition, Countly will discontinue this feature in the future.

### Solutions

- App open loading time is no longer tracked
- Time in foreground is no longer tracked
- Time in background is no longer tracked
- The client does not send Countly APM reports


yamilmedina (approved):
Countly 🫠

MohamadJaara (approved):

PR #3869, opened 2025-02-07
fix: FileSharing menu - no gaps #WPB-15276

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15276" title="WPB-15276" target="_blank">BugWPB-15276</a>  [Android] File sharing menu should not have a gaps
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

- There should be no gap between the additional options menu and the text input field or on the left and right side of the menu.
- Send message btn should not be visible when the keyboard is closed and there is no text to send. (this is out of the scope of task, but i found that regression, so just fixed it)

### Attachments (Optional)

Screenshot 2025-02-07 at 17 47 57

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInput.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInput.kt
@@ -30,8 +30,8 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.foundation.text.KeyboardOptions
-import androidx.compose.foundation.text.input.TextFieldLineLimits
 import androidx.compose.foundation.text.input.KeyboardActionHandler
+import androidx.compose.foundation.text.input.TextFieldLineLimits
 import androidx.compose.foundation.text.input.TextFieldState
 import androidx.compose.material3.HorizontalDivider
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInput.kt
```diff
@@ -258,13 +258,15 @@ private fun InputContent(
...
                 }
             }
-            MessageSendActions(
-                onSendButtonClicked = onSendButtonClicked,
-                sendButtonEnabled = canSendMessage,
-                selfDeletionTimer = viewModel.state(),
-                onChangeSelfDeletionClicked = onChangeSelfDeletionClicked,
-                modifier = Modifier.padding(end = dimensions().spacing8x)
-            )
+            if (canSendMessage || showOptions) {
```
MohamadJaara: please remove this condition it was removed before to show the send button even when the message composer is collapsed

MohamadJaara (changes requested):

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInput.kt
MohamadJaara: resolved in 1:1

MohamadJaara (approved):
LGTM

yamilmedina (approved):

PR #3876, opened 2025-02-13
feat: login type selector [WPB-16058]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16058" title="WPB-16058" target="_blank">TaskWPB-16058</a>  [Android] Create login flow selector
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- implemented `LoginTypeSelector` depending on common API version for the given server config
- `LoginTypeSelector` is available for all screen destinations as a non-navigation parameter and it's used whenever the login flow needs to be opened, it can take specific custom server config to select proper login flow for it or can use build's default config if not passed any
- to make it easier and faster to use (since default config is primarily used), for default config there's a `StateFlow`, for other custom ones it needs to run a suspend fun to get that info
- simplified the way of handling new intents, used `Channel` with built-in buffering without replaying instead of trying to achieve it using `SharedFlow` and moved deeplink handling to be closer to the navigation to be able to select proper destination
- removed `LocalNavigator` as it was not used anywhere and was partly commented out
- removed `ENTERPRISE_LOGIN_ENABLED` feature config flag as it's not needed anymore - from now on select login flow based on API or enforce specific one by adding a `login-type` parameter to the custom backend deeplink

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/3287

#### Test Coverage (Optional)

#### How to Test

Open deep link with custom backend that supports API v8, for instance `dev` build to support development builds and `anta` backend which supports development v8.

### Attachments (Optional)

https://github.com/user-attachments/assets/5ee50ca5-7179-4183-8905-16e7b3e3c6eb

`anta` supports development v8 so new login flow for `anta`, fallback to old flow for others.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
@@ -1,0 +1,54 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.android.navigation
+
+import com.wire.android.config.DefaultServerConfig
+import com.wire.android.di.KaliumCoreLogic
+import com.wire.android.util.dispatchers.DispatcherProvider
+import com.wire.kalium.logic.CoreLogic
+import com.wire.kalium.logic.configuration.server.ServerConfig
+import com.wire.kalium.logic.feature.auth.LoginContext
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.runBlocking
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class LoginTypeSelector @Inject constructor(
+    dispatcherProvider: DispatcherProvider,
+    @KaliumCoreLogic private val coreLogic: CoreLogic,
+) {
+    private val scope = CoroutineScope(SupervisorJob() + dispatcherProvider.default())
+    private val loginContextForDefaultServerConfigStateFlow: StateFlow<LoginContext> = runBlocking {
+        // it needs to be initialised before navigation is setup
+        loginContextFlow(DefaultServerConfig).stateIn(scope, SharingStarted.Eagerly, loginContextFlow(DefaultServerConfig).first())
+    }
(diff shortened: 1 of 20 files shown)


yamilmedina (approved):

Garzas (approved):

app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
```diff
@@ -0,0 +1,54 @@
...
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.runBlocking
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class LoginTypeSelector @Inject constructor(
```
sbakhtiarov, line 39: To be honest, I have a problem understanding what's going on here. Maybe there is a way to make it clearer or at least add some comments to help with reading?

saleniuk: What exactly is difficult to understand? I've added comment blocks for each function and the whole class to explain when and how to use them, so hopefully no one else will have such problems anymore.

app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
```diff
@@ -0,0 +1,73 @@
...
+    @KaliumCoreLogic private val coreLogic: CoreLogic,
+) {
+    private val scope = CoroutineScope(SupervisorJob() + dispatcherProvider.default())
+
+    // StateFlow of the login context for the default server config, so that the value is kept ready to use and the use case doesn't need
+    // to be executed every time the app needs to determine if the new login flow can be used.
+    private val loginContextForDefaultServerConfigStateFlow: StateFlow<LoginContext> = runBlocking {
+        // it needs to be initialised before navigation is setup
+        loginContextFlow(DefaultServerConfig).stateIn(scope, SharingStarted.Eagerly, loginContextFlow(DefaultServerConfig).first())
+    }
```
ohassine: `runBlocking` will current thread until the coroutine inside it completes, causing an ANR.
Can't we show a loading indicator till this is fetched ?

ohassine (commented):
Maybe better to check with someone from design, but I think bottom sheets should show up from the bottom.

app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
sbakhtiarov: Hard to follow the implementation in the class: combination of runBlocking, getting the first value of the flow for another StateFlow based on the same flow and some more logic that reads from these flows.

app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
saleniuk: Sure, it just takes a value from db so it shouldn't take that long to generate ANR, but I removed `runBlocking` and move initialisation to the place where we initialise all other things in `WireActivity` so that it will keep showing placeholder until initialised.

saleniuk:
> Maybe better to check with someone from design, but I think bottom sheets should show up from the bottom.

These screens look like bottom sheets, but they are not bottom sheets 🙃 We have specification for screens that they should always animate left-to-right, I already checked and it still looks smoother than closing current screen by animating it to the bottom and showing another from the bottom, but it's just a matter of changing the style in screen's annotation, so it can be changed very quickly if needed 😄

app/src/main/kotlin/com/wire/android/navigation/LoginTypeSelector.kt
saleniuk: It's required to provide initial value to create a `StateFlow`, but it's not relevant anymore as I changed that to adjust to comments from Oussama so now the simpler version of `stateIn` can be used without `runBlocking`.

sbakhtiarov (approved):

ohassine (approved):

PR #3878, opened 2025-02-17
fix: CallControlButtons shape #WPB-16054

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16054" title="WPB-16054" target="_blank">StoryWPB-16054</a>  [Android] UI Missing Calling Button Shape
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

The calling buttons (camera, speaker, and mic) are missing their intended round shape on the preview screen in version 

### Causes (Optional)

missed it somewhere :) 

### Solutions

add a borders for buttons

### Attachments (Optional)

| Before | After |
|  photo_5303385085950684910_y | photo_5303385085950684859_y  |
| photo_5303385085950684911_y  |  photo_5303385085950684860_y |

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/calling/controlbuttons/WireCallControlButton.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/calling/controlbuttons/WireCallControlButton.kt
@@ -42,11 +42,10 @@ fun WireCallControlButton(
     height: Dp = dimensions().defaultCallingControlsHeight,
     iconSize: Dp = dimensions().defaultCallingControlsIconSize
 ) {
-    val shape = CircleShape
     WireSecondaryIconButton(
         onButtonClicked = onClick,
         iconResource = iconResId,
-        shape = shape,
+        shape = CircleShape,
         colors = with(colorsScheme()) {
             wireSecondaryButtonColors().copy(
                 selected = inverseSurface,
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/wire/android/ui/calling/controlbuttons/WireCallControlButton.kt
```diff
@@ -63,6 +68,6 @@ fun WireCallControlButton(
         minSize = DpSize(width, height),
         minClickableSize = DpSize(width, height),
         iconSize = iconSize,
-        modifier = modifier,
+        modifier = modifier.border(width = 1.dp, color = MaterialTheme.wireColorScheme.secondaryButtonDisabledOutline, shape = shape)
```
saleniuk: These buttons are more "secondary" than "primary", `WireSecondaryIconButton` already contains a border, so I think it would be better to just use proper `WireSecondaryIconButton` and just change `selected` color instead of using `WirePrimaryIconButton` and applying `wireSecondaryButtonColors` and border to it

borichellow: fair point :)

ohassine (approved):

MohamadJaara (approved):

PR #3887, opened 2025-02-21
feat: Cells UI #1 (#WPB-15745)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15745" title="WPB-15745" target="_blank">TaskWPB-15745</a>  [Android] Upload file to cell as draft
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-15736
https://wearezeta.atlassian.net/browse/WPB-15738
https://wearezeta.atlassian.net/browse/WPB-15740
https://wearezeta.atlassian.net/browse/WPB-15745

# What's new in this PR?

### Issues
- Updated UI for message attachments.
- Adding multiple attachments to a message. Show upload progress. Add / remove attachments.
- Show draft attachments when re-opening conversation.
- New UI for attachments in conversations.
- Video / PDF files previews.

https://github.com/user-attachments/assets/8c15355e-c1cb-4379-bae9-476965834a96

image
image
image


app/src/main/kotlin/com/wire/android/mapper/RegularMessageContentMapper.kt
```diff
@@ -144,10 +147,11 @@ class RegularMessageMapper @Inject constructor(
         with(assetContent) {
             return UIMessageContent.AudioAssetMessage(
                 assetName = name ?: "",
-                assetExtension = mimeType,
+                assetExtension = name?.fileExtension() ?: mimeType,
```
saleniuk, line 150: Can we safely assume that name always contains the extension? I mean if we get an asset with name like "photo from 05.03.2025" without extension then this method will return `2025` as an extension. Maybe we should keep relying on the mime type and use `getExtensionFromMimeType`. 🤔
Also, now we're passing extension or mime type and both are slightly different, maybe it was wrongly `mimeType` because this `assetExtension` field hasn't been used for the audio asset anywhere but we should probably pass proper value if we want to start using it somewhere. 😄

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/AttachmentDraftView.kt
```diff
@@ -0,0 +1,59 @@
...
+                attachment = attachment,
+            )
+
+            AttachmentFileType.VIDEO -> AttachmentVideoView(
+                attachment = attachment,
+            )
+
+            else -> AttachmentFileView(
+                attachment = attachment,
+                modifier = Modifier.width(300.dp)
```
saleniuk: Please use `dimensions()` wherever possible 🙇

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/AttachmentDraftView.kt
```diff
@@ -0,0 +1,59 @@
...
+
+@Composable
+fun AttachmentDraftView(
+    attachment: AttachmentDraftUi,
+    onClick: () -> Unit,
+    onClickDelete: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+
+    val extension = remember(attachment.fileName) { attachment.fileName.substringAfterLast('.') }
```
saleniuk: Can we reuse `fileExtension()` method so that we have the code for retrieving extension from file name in one place and it always works the same way?

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/AttachmentFileView.kt
```diff
@@ -0,0 +1,114 @@
...
+                    .fillMaxHeight()
+                    .weight(1f)
+            )
+            Text(
+                modifier = Modifier
+                    .padding(horizontal = dimensions().spacing12x)
+                    .fillMaxWidth(),
+                style = typography().body02,
+                color = colorsScheme().onSurface,
+                fontSize = 14.sp,
```
saleniuk: We shouldn't change the font size, the idea is that our `typography()` contains all fonts that are used in the app. If there is no such font then it means that either it's wrong on the designs or this new font was added to the design documentation and we haven't updated our `typography()` to have it as well.
In this case I think that it should be `15.sp` so it's exactly the size that `typography().body02` has, no need to change that to `14.sp` 😃 
image

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/FileHeaderView.kt
```diff
@@ -0,0 +1,116 @@
...
+                text = label,
+                style = typography().subline01,
+                color = labelColor ?: colorsScheme().secondaryText,
+            )
+        }
+    }
+}
+
+@Suppress("MagicNumber")
+private fun Long.formattedFileSize() = when {
```
saleniuk: We already have such method: `DeviceUtil.formatSize(sizeInBytes: Long): String`

app/src/main/kotlin/com/wire/android/ui/home/cell/CellListItem.kt
```diff
@@ -0,0 +1,115 @@
...
+                    .padding(16.dp)
+                    .background(
+                        color = colorsScheme().onPrimaryVariant,
+                        shape = RoundedCornerShape(4.dp)
+                    )
+                    .padding(4.dp),
+                text = "DRAFT",
+                color = colorsScheme().inverseOnSurface,
+                fontSize = 10.sp
+            )
```
saleniuk, line 96: I couldn't find the designs for this "draft" label, but the closest I could find was this badge on conversation list:
image
so we could reuse `MembershipQualifierLabel` here, but if not then I would change these things in the current code:
- background color should be `colorsScheme().primary` instead of `onPrimaryVariant` and color of the text should be `colorsScheme().onPrimary` instead of `inverseOnSurface` - the ones starting with "on..." are used for used for drawing content on top of the given color (like we have `primary` and if you want to put a text on that color then you use `onPrimary`
- move text "DRAFT"  to `strings.xml`
- do not set `fontSize` but use the ones from our design system, in this case `badge01` would be the most appropriate if you need `10.sp` size, or `label03` - we use that for other badges

app/src/main/kotlin/com/wire/android/ui/home/cell/CellViewModel.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class CellViewModel @Inject constructor(
+    private val observeCellFilesUseCase: ObserveCellFilesUseCase,
+) : ViewModel() {
+
+    private val _uiMessage = MutableSharedFlow<String>()
```
saleniuk, line 40: I don't see any items being emitted into this `_uiMessage` anywhere in this view model - will that be implemented in the future? 🤔

app/src/main/kotlin/com/wire/android/ui/home/cell/WireCellScreen.kt
```diff
@@ -0,0 +1,121 @@
...
+    ) {
+        LazyColumn(
+            modifier = Modifier
+                .fillMaxSize()
+                .weight(1f),
+            verticalArrangement = Arrangement.spacedBy(2.dp)
+        ) {
+            items(
+                items = state.files,
+                key = {
```
saleniuk, line 91: I would suggest to also add `contentType`, it helps with item compositions of the same type so that they can be reused more efficiently.

app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/video/VideoMessage.kt
```diff
@@ -0,0 +1,293 @@
...
+            style = typography().body02,
+            color = colorsScheme().onSurfaceVariant,
+        )
+
+        Box(
+            modifier = Modifier
+                .fillMaxWidth(widthFraction(width, height))
+                .aspectRatio(aspectRatio(width, height))
+                .background(
+                    color = colorsScheme().backdrop,
```
saleniuk: `backdrop` color in our app is called `scrim` and it's exactly black color with 55% opacity, I'm almost sure that it's been named like this in our color documentation when we were implementing it, so maybe it was changed recently to "backdrop".  You can just change the name of `scrim` color in our `WireColorScheme` to `backdrop` if you want as it's not used anywhere else currently 😄 
But please remember to not add any colors to our color scheme unless it's also added to our [design color documentation](https://www.figma.com/design/JGiz61Nl8WCobwPkHoF9yZ/Documentation?node-id=10-132) - we tried hard to keep them synced 😅

app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/video/VideoMessage.kt
```diff
@@ -0,0 +1,293 @@
...
+                    model = model,
+                    contentScale = ContentScale.Fit,
+                    contentDescription = null,
+                )
+                duration?.let {
+                    Text(
+                        modifier = Modifier
+                            .fillMaxWidth()
+                            .align(Alignment.BottomEnd)
+                            .background(color = Color.Black.copy(alpha = 0.5f))
```
saleniuk: Here, instead of applying the color manually please use color from our scheme as well - just apply `backdrop`/`scrim` again as it should actually also be this exact color (black with 55% opacity instead of 50%)
image

core/ui-common/src/main/kotlin/com/wire/android/ui/theme/WireColorPalette.kt
```diff
@@ -317,4 +317,7 @@ object WireColorPalette {
     val Pink = Color(0xFFEB239B)
     @Stable
     val Turquoise = Color(0xFF01718E)
+
+    @Stable
+    val SmokeyGray = Color(0xFF737373)
```
saleniuk: This color is not a part of our [design color palette](https://www.figma.com/design/JGiz61Nl8WCobwPkHoF9yZ/Documentation?node-id=3091-15033) so it shouldn't be added to it

core/ui-common/src/main/kotlin/com/wire/android/ui/theme/WireColorScheme.kt
```diff
@@ -83,7 +83,9 @@ class WireColorScheme(
     val wireAccentColors: WireAccentColors,
 
     val emojiBackgroundColor: Color,
+    val backdrop: Color,
```
saleniuk: Change the name of `scrim` color to `backdrop` instead of adding a new one

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/AttachmentDraftView.kt
sbakhtiarov: Yes, will do. 
Apart from consistency, could you tell what is the reason for having a code like `dimensions().spacing300x` instead of `300.dp` ?
I see the reason behind`dimensions().fabIconSize` or similar, but cannot figure out the reason for `spacing12x`

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/AttachmentDraftView.kt
sbakhtiarov: Yes, updating to use `CommonUtils.fileExtension()` in all places.

app/src/main/kotlin/com/wire/android/ui/home/cell/CellListItem.kt
sbakhtiarov: The draft label is just for testing, will be replaced with something else when we have actual design.

app/src/main/kotlin/com/wire/android/ui/home/cell/CellViewModel.kt
sbakhtiarov: It was implemented in the past :) This was the first screen connected to cells sdk that I added for testing. Then I refactored it for demo and keeping it as a placeholder for the actual files list. I will be working on it next week.

app/src/main/kotlin/com/wire/android/mapper/RegularMessageContentMapper.kt
sbakhtiarov: This change was done to support showing the file extension in the new asset message design. I will review the extension usage in other places.

app/src/main/kotlin/com/wire/android/ui/common/attachmentdraft/ui/AttachmentDraftView.kt
saleniuk: I don't fully remember but I think that we started with some meaningful names but then we realised that it may become overwhelming with having too many values. On designs there are only numerical values so we were trying to extract such reusable values ourselves and give them some specific names, but sometimes it was difficult so some of us just started using `spacing_x` and we never really found any good solution to clean it up.  😓 
Now at least we try to keep all these values in `dimensions()` so that they are in a single place and it has a logic that can allow us to easily adjust it for different screen sizes (we could simply just multiply or divide these values if needed; names wouldn't match but we wouldn't have to go through the whole code).
If you know any better idea for that how to clean it up, let me know, I will be happy to help with this. 😄

saleniuk (approved):

PR #3907, opened 2025-03-07
feat: Cells UI (Multipart message) (#WPB-15746)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15746" title="WPB-15746" target="_blank">TaskWPB-15746</a>  [Android] Send message with cell attachments
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-15746
----
# What's new in this PR?

Support for showing new Multipart message in group conversation.

image

- UI for cell support configuration in group settings
- New UI message type
- New UI for showing multipart message with assets preview


app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationScreen.kt
```diff
@@ -729,21 +744,21 @@ fun ConversationScreen(
         }
     }
 
-//    imagePreviewScreenResultRecipient.onNavResult { result ->
```
sbakhtiarov, line 732: Uncommenting old code for handling attachments to support existing attachment UX for conversations without wire cell support.

app/src/main/kotlin/com/wire/android/ui/home/conversations/edit/MessageOptionsModalSheetLayout.kt
```diff
@@ -106,14 +108,28 @@ fun MessageOptionsModalSheetLayout(
...
                         {
-                            (message.messageContent as? UIMessageContent.TextMessage)?.let {
-                                sheetState.hide {
-                                    onEditClick(
-                                        message.header.messageId,
-                                        message.messageContent.messageBody.message.asString(context.resources),
-                                        (message.messageContent.messageBody.message as? UIText.DynamicString)?.mentions ?: listOf()
-                                    )
-                                }
+                            when (message.messageContent) {
```
sbakhtiarov, line 111: Support for editing text in Multipart message.

app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/multipart/MultipartAttachmentsView.kt
```diff
@@ -0,0 +1,141 @@
...
+}
+
+@Composable
+private fun AttachmentsGrid(
+    attachments: List<MultipartAttachmentUi>,
+    onClick: (MultipartAttachmentUi) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    LazyVerticalGrid(
+        modifier = modifier.heightIn(max = 1000.dp),
```
saleniuk: I think this could be the value that I would create a dedicated field in `WireDimensions` for, like `attachmentsGridMaxHeight` as we could want to have a different value for different screen dimensions so having a dedicated field would allow us to set a different value for `landscape` or `tablet` dimensions type in the future 😃

app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/multipart/standalone/VideoAssetPreview.kt
```diff
@@ -0,0 +1,238 @@
...
+            item.metadata?.durationMs()?.let {
+                Text(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .align(Alignment.BottomEnd)
+                        .background(color = colorsScheme().scrim)
+                        .padding(dimensions().spacing4x),
+                    text = DateAndTimeParsers.videoMessageTime(it),
+                    style = typography().subline01,
+                    color = Color.White,
```
saleniuk: It needs to always be white because it's drawn on `scrim` background, right?
In that case, instead of hardcoding this color, you can add it to `WireFixedColorScheme` (it's the scheme for colors that are always the same no matter what theme is currently used) or use directly dark palette like we do for calling `ParticipantTile` (it's always drawn on dark background so it needs dark color scheme even for light theme), so in this case probably `darkColorsScheme().onSurface`

PR #3912, opened 2025-03-12
fix: crashes on audio message playback [WPB-16195]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16195" title="WPB-16195" target="_blank">TaskWPB-16195</a>  [Android]Audio not available or app crashes on playback
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The app crashes as soon as the user taps the play button on audio message.

### Causes (Optional)

Race condition when fetching audio asset file - it can be executed multiple times at once as it's also called to fetch waves mask and then asset file can be overwritten and broken (created but empty for instance) resulting in a crash when trying to play empty file. 
The "fetch waves mask" action can also be called multiple times for each message as it's called every time paging data is updated. The whole map of audio states is propagated through all composables all way down from the screen to the `AudioMessage` item and it's not stable so it can be also prone to more unneeded recompositions. 
`ConversationAudioMessagePlayer` dropping and losing some of the crucial states emitted because of `DROP_OLDEST` and for instance making the audio message be played without any indication on the screen.

### Solutions

Secured `getAssetMessage` by keeping the deferred until it's completed so that each execution during that time will reuse the same deferred and it won't be called multiple times at the same time - no issues with overwriting the file and no unnecessary requests.
Created dedicated scoped `AudioMessageViewModel` which handles the state for the given message and executes calls to play the audio associated with that message, change the audio speed or fetch waves mask for that audio so that it happens only once and is kept in the view model. Thanks to that, a map with all audio states doesn't need to be passed all the way to the `AudioMessage` but instead it uses state from injected scoped `AudioMessageViewModel`.
Replaced `DROP_OLDEST` with `SUSPEND` in `ConversationAudioMessagePlayer` to not lose any state updates.

#### Test Coverage (Optional)

#### How to Test

Receive audio messages and try to play them.

----

----

----
##### References


app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
```diff
@@ -413,14 +412,27 @@ class ConversationAudioMessagePlayer
...
-        currentAccountResult: CurrentSessionResult,
+        userId: UserId,
         conversationId: ConversationId,
-        messageId: String
-    ) = coreLogic
-        .getSessionScope((currentAccountResult as CurrentSessionResult.Success).accountInfo.userId)
-        .messages
-        .getAssetMessage(conversationId, messageId)
+        messageId: String,
+    ): MessageAssetResult = withContext(Dispatchers.IO) {
```
vitorhugods: I think we should inject dispatchers into this MessagePlayer.
Otherwise it isn't fully testable.

vitorhugods (approved):
Really nice improvement 🚀

saleniuk:
closing in favour of the one that will be merged to RC first: https://github.com/wireapp/wire-android/pull/3917

PR #3917, opened 2025-03-14
fix: crashes on audio message playback [WPB-16195]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16195" title="WPB-16195" target="_blank">TaskWPB-16195</a>  [Android]Audio not available or app crashes on playback
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The app crashes as soon as the user taps the play button on audio message.

### Causes (Optional)

Race condition when fetching audio asset file - it can be executed multiple times at once as it's also called to fetch waves mask and then asset file can be overwritten and broken (created but empty for instance) resulting in a crash when trying to play empty file. 
The "fetch waves mask" action can also be called multiple times for each message as it's called every time paging data is updated. The whole map of audio states is propagated through all composables all way down from the screen to the `AudioMessage` item and it's not stable so it can be also prone to more unneeded recompositions. 
`ConversationAudioMessagePlayer` dropping and losing some of the crucial states emitted because of `DROP_OLDEST` and for instance making the audio message be played without any indication on the screen.

### Solutions

Secured `getAssetMessage` by keeping the deferred until it's completed so that each execution during that time will reuse the same deferred and it won't be called multiple times at the same time - no issues with overwriting the file and no unnecessary requests.
Created dedicated scoped `AudioMessageViewModel` which handles the state for the given message and executes calls to play the audio associated with that message, change the audio speed or fetch waves mask for that audio so that it happens only once and is kept in the view model. Thanks to that, a map with all audio states doesn't need to be passed all the way to the `AudioMessage` but instead it uses state from injected scoped `AudioMessageViewModel`.
Replaced `DROP_OLDEST` with `SUSPEND` in `ConversationAudioMessagePlayer` to not lose any state updates.

#### Test Coverage (Optional)

#### How to Test

Receive audio messages and try to play them.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
+++ b/app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
@@ -21,10 +21,12 @@ import android.content.Context
 import android.media.MediaPlayer
 import android.media.MediaPlayer.SEEK_CLOSEST_SYNC
 import android.net.Uri
+import androidx.annotation.VisibleForTesting
 import com.wire.android.R
 import com.wire.android.di.ApplicationScope
 import com.wire.android.di.KaliumCoreLogic
 import com.wire.android.services.ServicesManager
+import com.wire.android.util.ExpiringMap
 import com.wire.android.util.dispatchers.DispatcherProvider
 import com.wire.android.util.extension.intervalFlow
 import com.wire.android.util.ui.UIText
@@ -55,8 +57,9 @@ import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.scan
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
-import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -72,9 +75,13 @@ class ConversationAudioMessagePlayer
     @KaliumCoreLogic private val coreLogic: CoreLogic,
     @ApplicationScope private val scope: CoroutineScope,
     private val dispatchers: DispatcherProvider,
+    currentTime: () -> Long = { System.currentTimeMillis() },
 ) {
-    private companion object {
-        const val UPDATE_POSITION_INTERVAL_IN_MS = 1000L
+    companion object {
+        private const val UPDATE_POSITION_INTERVAL_IN_MS = 1000L
+
+        @VisibleForTesting
+        internal const val GET_ASSET_MESSAGE_CACHE_EXPIRATION_MS = 5 * 60_000L // 5 minutes
     }
 
     init {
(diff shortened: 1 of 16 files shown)


app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
```diff
@@ -91,16 +98,14 @@ class ConversationAudioMessagePlayer
...
+        onBufferOverflow = BufferOverflow.SUSPEND
     )
 
     private val _audioSpeed = MutableStateFlow<AudioSpeed>(AudioSpeed.NORMAL)
     val audioSpeed: Flow<AudioSpeed> = _audioSpeed.onStart { emit(_audioSpeed.value) }
 
     private val seekToAudioPosition = MutableSharedFlow<Pair<MessageIdWrapper, Int>>(
-        onBufferOverflow = BufferOverflow.DROP_OLDEST,
-        extraBufferCapacity = 1
+        onBufferOverflow = BufferOverflow.SUSPEND
```
sbakhtiarov, line 109: Is "SUSPEND" mode also required here? Maybe for position seek events we are only interested in the most latest?

app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
```diff
@@ -413,14 +417,27 @@ class ConversationAudioMessagePlayer
...
+        val key = GetAssetMessageKey(userId, conversationId, messageId)
+        // keep deferred in the map to prevent multiple calls to the same asset at the same time, instead just reuse the existing deferred
+        getAssetMessageDeferredMap.getOrPut(key) {
+            coreLogic
+                .getSessionScope(userId)
+                .messages
+                .getAssetMessage(conversationId, messageId)
+                .also {
+                    it.invokeOnCompletion {
+                        getAssetMessageDeferredMap.remove(key)
```
sbakhtiarov: Could we cache the result or it must be always updated from the database?

sbakhtiarov (approved):

MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
saleniuk: It would work if this flow was used only for a single audio message but it's used for all audio messages so it could possibly drop some updates for other messages, in that case I prefer to keep `SUSPEND` here as well. 😄

app/src/main/kotlin/com/wire/android/media/audiomessage/ConversationAudioMessagePlayer.kt
saleniuk: @sbakhtiarov good idea, I wasn't sure about that but you convinced me 😄 
It's a bit more complex as it requires checking whether it failed as we would want to retry fetching when clicking on "play" again but it wasn't hard to implement - I added your `ExpiringMap` and used mutex to make actions atomic. I think 5 mins cache is fine but let me know if you think differently.

saleniuk: After discussing with Sergei, we decided that it's not necessary to make this cache expiring so I simplified it.
He also suggested to create another ticket to refactor it so that we will generate the waves mask once and store it in the database, so here it is: https://wearezeta.atlassian.net/browse/WPB-16588

sbakhtiarov (approved):

PR #3930, opened 2025-03-20
feat: Browse all files [WPB-15838]

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15838" title="WPB-15838" target="_blank">EpicWPB-15838</a>  [Android] Implement Global File Search
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-15838

# Browse all files uploaded to Wire Cell

- Loading and displaying all files from wire cells with user and conversation names.
- Download, share, delete options
- Manage public link for each file (create, load, delete, copy, share)

image
image
image


app/src/main/kotlin/com/wire/android/ui/home/cell/GlobalCellsScreen.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.wire.android.ui.home.HomeStateHolder
+
+@HomeNavGraph
+@WireDestination
+@Composable
+fun GlobalCellsScreen(
+    homeStateHolder: HomeStateHolder,
+) {
+
+    val navigator = LocalNavigator.current
```
saleniuk: `LocalNavigator` will be removed with the introduction of enterprise login as I noticed that it hasn't been used anywhere (for destination screens it's always better to just add it as a function parameter).
In this case, for home destinations, `Navigator` is already a part of `HomeStateHolder` and available from there

features/cells/src/main/java/com/wire/android/feature/cells/ui/WireCellScreen.kt
```diff
@@ -0,0 +1,236 @@
...
+import com.wire.android.ui.common.progress.WireCircularProgressIndicator
+import com.wire.android.ui.common.search.SearchBarState
+import kotlinx.coroutines.delay
+
+@Suppress("CyclomaticComplexMethod")
+@Composable
+fun WireCellScreen(
+    searchBarState: SearchBarState,
+    showPublicLinkScreen: (String, String, String?) -> Unit,
+    viewModel: CellViewModel = hiltViewModel(),
```
saleniuk, line 59: Creating previews for this screen won't work because of passing the view model.
You can extract an interface and create a "preview" version of this view model just like we do for `ConversationCallListViewModel` in `ConversationScreenContent`, so something like:
```
viewModel: CellViewModel =  when {
        LocalInspectionMode.current -> CellViewModelPreview()
        else -> hiltViewModel<CellViewModelImpl>()
```

features/cells/src/main/java/com/wire/android/feature/cells/ui/publiclink/PublicLinkScreen.kt
```diff
@@ -0,0 +1,246 @@
...
+import com.wire.android.ui.common.topappbar.NavigationIconType
+import com.wire.android.ui.common.topappbar.WireCenterAlignedTopAppBar
+import com.wire.android.ui.common.typography
+import com.wire.android.ui.theme.WireTheme
+import kotlinx.coroutines.flow.collectLatest
+
+@Destination(
+    navArgsDelegate = PublicLinkNavArgs::class,
+    style = PopUpNavigationAnimation::class,
+)
```
saleniuk: Can you also add `wrappers = [WaitUntilTransitionEndsWrapper::class],`?
This wrapper is to make sure no interaction that could break the navigation happens during the transition.
We should probably extract `annotation class WireDestination` (it has that wrapper by default) into some common module so that it could be reused in all other modules, for now you could just add it manually.

features/cells/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,49 @@
...
+    <string name="no_app_found">No application found to open the file.</string>
+    <string name="cell_files_load_failure_message">Failed to load. Please try again later.</string>
+    <string name="action_failed">Action failed.</string>
+    <string name="cancel">Cancel</string>
+    <string name="confirm_delete_title">Delete file</string>
+    <string name="confirm_delete_text">Please confirm deleting file from cell.</string>
+    <string name="reload">Reload</string>
+    <string name="error_delete_public_link">Failed to remove public link.</string>
+    <string name="error_load_public_link">Failed to load public link.</string>
+    <string name="error_create_public_link">Faile to create public link.</string>
```
ohassine, line 43:
```suggestion
    <string name="error_create_public_link">Failed to create public link.</string>
```

ohassine (approved):

PR #3934, opened 2025-03-24
fix: handle 2fa on RegisterDeviceScreen [WPB-16573]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16573" title="WPB-16573" target="_blank">BugWPB-16573</a>  [Android] Invalid information alert received although login is valid
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`RegisterDeviceScreen`, which is kind of a fallback flow when the device hasn't been registered successfully during account creation or login, has been missing a logic of handling 2FA codes when registering a client so in that case user hasn't been able to register device properly resulting in not being able to use the app properly.

### Solutions

Implemented handling 2FA code also on that screen, similarly to the way it's handled during the login - first the code that is persisted from the last login is used and if that returns `Invalid2FA` then new code is requested and the user needs to enter it to complete the registration flow.

#### Test Coverage (Optional)

#### How to Test

It's hard to reproduce manually as it can happen for instance when login is completed but device registration request that's executed right after it fails because of a crash for instance.

### Attachments (Optional)

https://github.com/user-attachments/assets/824eed0a-d975-4a7e-8597-d0647e9ebddc

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/authentication/verificationcode/VerificationCodeScreenContent.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/authentication/verificationcode/VerificationCodeScreenContent.kt
@@ -41,7 +41,6 @@ import com.wire.android.ui.theme.WireTheme
 import com.wire.android.util.ui.PreviewMultipleThemes
 import com.wire.android.util.ui.UIText
 
-
 @Composable
 fun VerificationCodeScreenContent(
     verificationCodeTextState: TextFieldState,
@@ -94,9 +93,11 @@ private fun MainContent(
         style = typography().body01,
         textAlign = TextAlign.Start
     )
-    Spacer(modifier = Modifier
+    Spacer(
+        modifier = Modifier
         .height(dimensions().spacing8x)
-        .weight(1f))
+        .weight(1f)
+    )
     VerificationCode(
         codeLength = codeState.codeLength,
         codeState = codeTextState,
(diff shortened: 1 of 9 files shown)


MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/authentication/verificationcode/VerificationCodeScreenContent.kt
```diff
@@ -0,0 +1,129 @@
...
+    BackHandler { onBackPressed() }
+    WireScaffold(
+        topBar = {
+            WireCenterAlignedTopAppBar(
+                elevation = dimensions().spacing0x,
+                title = stringResource(id = R.string.second_factor_authentication_title),
+                navigationIconType = null,
+            )
+        }
+    ) { internalPadding ->
```
ohassine, line 61: Shouldn't be the `WireScaffold` in `LoginEmailVerificationCodeScreen`, and here we have just `MainContent` ?

ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/authentication/verificationcode/VerificationCodeScreenContent.kt
saleniuk: We usually put `WireScaffold` in `...ScreenContent` composables so that we are able to create a full preview of it.
Here we would need to duplicate it in both `LoginEmailVerificationCodeScreen` and `RegisterDeviceVerificationCodeScreen`, that's why I extracted this common code into single composable.

PR #3941, opened 2025-03-27
fix: cancel login operation when going back [WPB-16670]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16670" title="WPB-16670" target="_blank">BugWPB-16670</a>  [Android] Back arrow on login screen does not cancel login operation
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

On the login screen, when a user enters their login credentials and taps the "Next" button, the login process begins. However, if the user taps the back arrow in the top left corner while the page is still loading, the UI navigates back to the previous screen, but the login process continues in the background. If the app is later restarted, the user is found to be logged in, indicating that the login was completed despite the UI navigation.

### Solutions

Keep the login operation job and cancel it when user decides to go back. After that, some clean-up is required, depending on how far the login process went, but it may be required to remove self user's email, clean the db, close the user's scope to cancel all ongoing user processes and delete the session so that it looks like the account was never logged in. To achieve that, the easiest way is to perform hard logout and then remove the session - this way we are sure it makes all required actions. After that also it's required to switch back to the previous account so the previous account's id is kept during the login process.
Some of it also applies to other existing screens like remove device - when user goes back then the app needs to logout the user, but the actual logout was never executed, just the session was being removed, so there was some data being left, so to unify these processes logout was added also to `ClearSessionViewModel` and this view model was used on `E2EIEnrollmentScreen` - previously the same logic was duplicated on `E2EIEnrollmentViewModel`, now it uses `ClearSessionViewModel` as well to make it easier to maintain.
Fixed some other small issue as well that started to be more noticeable here after creating dedicated subgraphs for login flows - screens belonging to subgraphs were not included when logging screen name properly or deciding whether connection top bar should be visible or not that started. That was because `.destinationsByRoute` and `.destinations` only return screens from particular graph not including subgraphs, so it was replaced with `.findDestination` and `.allDestinations` to have all screens covered.

#### Test Coverage (Optional)

#### How to Test

Login and navigate back during the login process. 
Close the app and open again to check if it still not logged in. 
Try to log in again using the same credentials to check that there are no unwanted data left over from the previous attempt.

### Attachments (Optional)

https://github.com/user-attachments/assets/6ef754ca-58ff-4553-b1c5-53f2c16fa8a0

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/authentication/login/email/LoginEmailViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/authentication/login/email/LoginEmailViewModel.kt
@@ -18,6 +18,7 @@
 
 package com.wire.android.ui.authentication.login.email
 
+import androidx.annotation.VisibleForTesting
 import androidx.compose.foundation.text.input.TextFieldState
 import androidx.compose.foundation.text.input.clearText
 import androidx.compose.foundation.text.input.setTextAndPlaceCursorAtEnd
@@ -43,6 +44,8 @@ import com.wire.android.util.dispatchers.DispatcherProvider
 import com.wire.kalium.logic.CoreLogic
 import com.wire.kalium.logic.data.auth.login.ProxyCredentials
 import com.wire.kalium.logic.data.auth.verification.VerifiableAction
+import com.wire.kalium.logic.data.logout.LogoutReason
+import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.feature.auth.AddAuthenticatedUserUseCase
 import com.wire.kalium.logic.feature.auth.AuthenticationResult
 import com.wire.kalium.logic.feature.auth.AuthenticationScope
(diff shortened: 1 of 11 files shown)


ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/authentication/login/email/LoginEmailViewModel.kt
```diff
@@ -134,10 +144,28 @@ class LoginEmailViewModel @Inject constructor(
...
+                } else {
+                    null
+                }
+            }
+            startLoginJob().let {
+                loginJobData.value = LoginJobData(it, previousSessionUserId)
+                it.invokeOnCompletion {
+                    loginJobData.value = null
+                }
+            }
```
kl3jvi: Hey! I hope you are doing fine. I have done some contributions to some small PR's here at Wire and I was looking into your PR. I was wondering based on your input as my context is limited, would you think canceling the previous loginJob and then starting a new one would be a good idea?
```kotlin
loginJobData.value?.job?.cancel()
startLoginJob().let { newJob ->
            loginJobData.value = LoginJobData(newJob, previousSessionUserId)
            newJob.invokeOnCompletion {
                loginJobData.value = null
            }
        }
```

saleniuk: Thank you for your input. It's generally a good idea, but in this particular case it shouldn't affect the flow, because the user simply cannot start another login job until previous one is finished (completed successfully or failed) because the button is disabled while the state is `Loading`. I think it's always good to make the code less error-prone so I added canceling the previous job and cleaning up leftovers from previous login before starting a new one, just to be sure. 😄

kl3jvi:
I can't approve PR's yet. Good job 🚀

yamilmedina (approved):

PR #3983, opened 2025-04-11
fix(channels): allow team admins to do admin tasks [WPB-17100]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17100" title="WPB-17100" target="_blank">TaskWPB-17100</a>  [Android]Team owner does not receive admin privileges when joining a channel created by a team member
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Team Admin is not able to perform admin tasks in a Channel.

### Causes

Wasn't implemented yet.

### Solutions

In general: just consider whether the user is a team admin and part of the same team as the conversation.

Code-wise, it hurts. The logic should be moved elsewhere. We're creating complicated logic in ViewModels, and even worse: within the Composables.

Ideally, we should extract most of this logic to a higher level. But for the sake of this release, I'm just fixing ASAP and targeting this as my next refactor.

### Testing

Hä? I wish. Manually tested.

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/conversation/ConversationSheetState.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/conversation/ConversationSheetState.kt
@@ -74,6 +74,7 @@ fun rememberConversationSheetState(
                             conversationId = conversationId,
                             isFromTheSameTeam = isFromTheSameTeam,
                             isPrivate = conversationItem.isPrivate,
+                            isSelfUserTeamAdmin = false // TODO: Get actual value
                         )
                     } else {
                         ConversationTypeDetail.Group.Regular(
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/conversation/ConversationSheetState.kt
```diff
@@ -74,6 +74,7 @@ fun rememberConversationSheetState(
                             conversationId = conversationId,
                             isFromTheSameTeam = isFromTheSameTeam,
                             isPrivate = conversationItem.isPrivate,
+                            isSelfUserTeamAdmin = false // TODO: Get actual value
```
vitorhugods, line 77: This is a super annoying value to get.

Most of the UI calling this function don't have a clue about user role within the team. So a compromise was made: to hardcode false.

The only implication is that a user that is a team admin won't be able to delete channels from the bottomsheet when long clicking on a Conversation, in the Conversation List.

In the future, as said before, we should refactor this.
We should have less generic composables/elements, and bring the logic to a higher layer, so instead of passing stuff like `isSelfUserTeamAdmin` or `isFromTheSameTeam`, we should pass meaningful values, like `isDeletePossible` to the presentation layer.

MohamadJaara (approved):

saleniuk (approved):

PR #3991, opened 2025-04-15
fix: fetch MLS public keys the same way as feature config [WPB-17161]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17161" title="WPB-17161" target="_blank">BugWPB-17161</a>  [Android] Excessive frequency of fetching MLS public keys
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Fetching MLS public keys should be treated the same way as feature-configs.
https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/160858877/Feature+flags+and+feature+configuration#Periodic-fetch-of-configuration

### Solutions

Add fetching MLS public keys to the `AppSyncViewModel`. Update kalium version to the one with MLS public keys changes to be able to use `MLSPublicKeysSyncWorker` (it executes fetch every 24h as well).

Update: use `ForegroundActionsUseCase` instead - solution implemented in this PR: https://github.com/wireapp/kalium/pull/3527

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/3414
- https://github.com/wireapp/kalium/pull/3527

#### Test Coverage (Optional)

----

----

----
##### References


MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/home/AppSyncViewModel.kt
```diff
@@ -83,6 +85,7 @@ class AppSyncViewModel @Inject constructor(
             listOf(
                 viewModelScope.launch(dispatcher.io()) { syncCertificateRevocationListUseCase() },
                 viewModelScope.launch(dispatcher.io()) { featureFlagsSyncWorker.execute() },
+                viewModelScope.launch(dispatcher.io()) { mlsPublicKeysSyncWorker.executeImmediately() },
```
vitorhugods: I am pretty sure this whole block of `runSyncTasks()` here was just a quick hack and now it is getting expanded further.

App knows too much about this. 

We should either:
A. Launch `mlsPublicKeysSyncWorker` when `UserSessionScope` starts, and have it wait for Sync to be LIVE before doing anything else. Once LIVE it will do its thing, be it on the background or not. And it can reschedule itself to repeat every 24h if still running.
B. Create a `ForegroundWorker` of some sorts in Kalium, so the app would call it and Kalium would take over.

saleniuk: @vitorhugods should I do it as part of this ticket or should we create a dedicated one to clear this up?

saleniuk: The documentation states that it should be called each time the user puts the app in the foreground, so running it each time the app becomes LIVE is not exactly the same.
And it actually waits until live when being executed, this logic is implemented already inside this worker, just like for feature flag sync worker and proteus sync worker.
About B idea, I don't know if I get it correctly but the only change would be that the app starts this single `ForegroundWorker` instead of starting each one separately?

saleniuk: I believe that after adjusting this PR to use new `ForegroundActionsUseCase` this can be resolved 😄

Garzas (approved):

PR #4000, opened 2025-04-23
chore: remove some non-determinism

licaon-kter:
Your app is not built reproducible in F-Droid, but we continuously test older versions on https://verification.f-droid.org/ and would like more and more apps to become repro

Looking at this report: https://verification.f-droid.org/unsigned/com.wire_100083279.apk.diffoscope.html

It seems easily fixable by updating the lib to 11.5.0 or later, as this was fixed in https://github.com/mikepenz/AboutLibraries/pull/1056

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -134,6 +134,7 @@ android {
 aboutLibraries {
     val isAboutLibrariesDisabled = System.getenv("DISABLE_ABOUT_LIBRARIES")?.equals("true", true) ?: false
     registerAndroidTasks = !isAboutLibrariesDisabled
+    excludeFields = arrayOf("generated")
 }
 
 dependencies {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -86,7 +86,7 @@ lint-compose = "1.4.2"
 openIdAppAuth = "0.11.1"
 
 # Other Tools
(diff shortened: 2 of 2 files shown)


licaon-kter:
nope @CLAassistant no can do, but anyone can pickup this change and whatever :wink:

eighthave:
FYI as this is a trivial change, copyright does not apply, therefore a CLA is unnecessary.

MohamadJaara (approved):
Thanks for the contribution

vitorhugods (approved):

vitorhugods:
It seems that the new version of AboutLibraries requires a newer version of core-ktx, which requires a newer compile version, so it isn't building 🙃

Hopefully these are the only surprises.

eighthave:
IIRC, there is a config option in _aboutlibraries_ to turn control the inclusion of the timestamp.  The timestamp used to be default on, the change was to make it default off.  So you could try keeping the same library version, then just setting it to not include the timestamp.

licaon-kter:
This one: https://github.com/mikepenz/AboutLibraries/issues/784#issuecomment-1205501876

licaon-kter:
@vitorhugods @MohamadJaara to speed it up, I've applied the other fix, and you can remove it once the lib and (core-ktx) gets upgraded

PR #4003, opened 2025-04-24
fix: move mls jobs to start from te UI [WPB-17367]

MohamadJaara:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17367" title="WPB-17367" target="_blank">BugWPB-17367</a>  [Android] unlimited MLS jobs when sync goes live
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

move MLS jobs to kick start only from the foreground

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/AppSyncViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/AppSyncViewModel.kt
@@ -19,10 +19,14 @@ package com.wire.android.ui.home
 
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import androidx.lifecycle.viewmodel.compose.viewModel
 import com.wire.android.appLogger
+import com.wire.kalium.logic.feature.client.MLSClientManager
+import com.wire.kalium.logic.feature.conversation.keyingmaterials.KeyingMaterialsManager
 import com.wire.kalium.logic.feature.e2ei.SyncCertificateRevocationListUseCase
 import com.wire.kalium.logic.feature.e2ei.usecase.ObserveCertificateRevocationForSelfClientUseCase
 import com.wire.kalium.logic.feature.featureConfig.FeatureFlagsSyncWorker
+import com.wire.kalium.logic.feature.mlsmigration.MLSMigrationManager
 import com.wire.kalium.logic.feature.server.UpdateApiVersionsUseCase
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.Job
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/wire/android/ui/home/AppSyncViewModel.kt
```diff
@@ -77,6 +84,9 @@ class AppSyncViewModel @Inject constructor(
                 viewModelScope.launch { featureFlagsSyncWorker.execute() },
                 viewModelScope.launch { observeCertificateRevocationForSelfClient.invoke() },
                 viewModelScope.launch { updateApiVersions() },
+                viewModelScope.launch { mLSClientManager() },
+                viewModelScope.launch { mLSMigrationManager() },
+                viewModelScope.launch { keyingMaterialsManager() },
```
vitorhugods, line 88: Instead of having the app explicitly knowing about MLS and how it works, can't we just mutate a `hasForegroundPriorityState` thing, and let the business logic handle itself in a single place within `Kalium`? 

Every time we add a new guy, we need to remember to do it here too. Also, other tools like TestService or whatever might be completely broken with these changes.

It would be similar to `NotifyConversationisOpenUseCase`, instead of having the ViewModels calling multiple UseCases, we centralise operations in a single point (clean up ephemeral, resolve 1:1, etc).

vitorhugods (changes requested):

app/src/main/kotlin/com/wire/android/ui/home/AppSyncViewModel.kt
MohamadJaara: that will be a nice solution for a bigger refactor, there are quite a lot of those worker that runs on kalium init that i would like to make em all cleaner, but this will take quite some time that we do not have, for now what do you think about this approach that we can revisit later?

MohamadJaara: for me the most important part is to not run all of this on every sync live update

vitorhugods: Just... other than increasing debt, it might still break TestService (not sure), so QA might not be the happiest about this "temporary change".

vitorhugods (approved):

Garzas (approved):

PR #4012, opened 2025-05-05
feat: handle federated case for user ids in QR code (WPB-10530)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10530" title="WPB-10530" target="_blank">StoryWPB-10530</a>  [Android] QR code Profile link Federation - update domain
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Federated case handling for QR code user profile. parsing, sharing, displaying.

### Causes (Optional)

Not prioritized before

### Solutions

Implement it.

----

----

----
##### References

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/userprofile/qr/SelfQRCodeScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/userprofile/qr/SelfQRCodeScreen.kt
@@ -280,7 +280,8 @@ fun PreviewSelfQRCodeContent() {
             SelfQRCodeState(
                 userId = UserId("userId", "wire.com"),
                 handle = "userid",
-                userProfileLink = "https://account.wire.com/user-profile/?id=aaaaaaa-222-3333-4444-55555555"
+                userProfileLink = "wire://user/wire.com/aaaaaaa-222-3333-4444-55555555",
+                userAccountProfileLink = "https://account.wire.com/user-profile/?id=aaaaaaa-222-3333-4444-55555555@wire.com"
             ),
             { "".toUri() },
             { }
--- a/app/src/main/kotlin/com/wire/android/util/deeplink/DeepLinkProcessor.kt
+++ b/app/src/main/kotlin/com/wire/android/util/deeplink/DeepLinkProcessor.kt
@@ -150,18 +150,22 @@ class DeepLinkProcessor @Inject constructor(
      */
     private fun getConnectingUserProfile(uri: Uri, switchedAccount: Boolean, accountInfo: AccountInfo.Valid): DeepLinkResult {
         val segments = uri.pathSegments
-        segments.takeLast(2).let { domainAndUserId ->
-            return if (domainAndUserId.size == 2) {
-                val userId = domainAndUserId[1]
-                val domain = domainAndUserId[0]
-                userId.toDefaultQualifiedId(domain).let {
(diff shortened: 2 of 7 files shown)


app/src/main/kotlin/com/wire/android/util/deeplink/DeepLinkProcessor.kt
```diff
@@ -145,22 +145,32 @@ class DeepLinkProcessor @Inject constructor(
...
+    /**
+     * Format of deeplink to parse: wire://user/domain/user-id
+     */
     private fun getConnectingUserProfile(uri: Uri, switchedAccount: Boolean, accountInfo: AccountInfo.Valid): DeepLinkResult {
-        // todo. handle with domain case, before lastPathSegment. format of deeplink wire://user/domain/user-id
-        return uri.lastPathSegment?.toDefaultQualifiedId(accountInfo.userId.domain)?.let {
-            DeepLinkResult.OpenOtherUserProfile(it, switchedAccount)
-        } ?: return DeepLinkResult.Unknown
+        val segments = uri.pathSegments
+        segments.takeLast(2).let { domainAndUserId ->
```
MohamadJaara: we should first check how many path segmements there is, and if 2 we take the last as the user ID and use self id as the domain and if 3 then take the last 2 as you did here to make sure it will work with older links

MohamadJaara (dismissed):

app/src/main/kotlin/com/wire/android/ui/userprofile/qr/SelfQRCodeScreen.kt
```diff
@@ -280,7 +280,8 @@ fun PreviewSelfQRCodeContent() {
             SelfQRCodeState(
                 userId = UserId("userId", "wire.com"),
                 handle = "userid",
-                userProfileLink = "https://account.wire.com/user-profile/?id=aaaaaaa-222-3333-4444-55555555"
+                userProfileLink = "wire://user/wire.com/aaaaaaa-222-3333-4444-55555555",
+                userAccountProfileLink = "https://account.wire.com/user-profile/?id=aaaaaaa-222-3333-4444-55555555@wire.com"
```
MohamadJaara, line 284: this does not match 
wire://user/<userDomain>/<userId> am i missing something

app/src/main/kotlin/com/wire/android/util/deeplink/DeepLinkProcessor.kt
yamilmedina: `takeLast` will take the right amount of elements (after the host, in this case `user`) and in case there are less it will just return the current items as a list, that's why it is the size 2 check inside or defaulting to the old behavior otherwise.

```kotlin
 val url0: Uri = "wire://user".toUri() --> emptyList()
 val url1: Uri = "wire://user/user-id".toUri() --> listOf(user-id)
 val url2: Uri = "wire://user/domain/user-id".toUri() --> listOf(domain, usre-id)
```

I can change it anyway and do it as you suggest.

yamilmedina: Changed ! To improve the readability anyway I think is better

Garzas (approved):

sbakhtiarov (approved):

PR #4018, opened 2025-05-12
feat: Navigate between folders (WPB-17392)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17392" title="WPB-17392" target="_blank">TaskWPB-17392</a>  [Android] Show folders in files screen
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Follow up PR of https://github.com/wireapp/wire-android/pull/4007

It aims to allow user to navigate between folders and sub folders

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


features/cells/src/main/java/com/wire/android/feature/cells/ui/AllFilesScreen.kt
```diff
@@ -60,8 +60,11 @@ fun AllFilesScreen(
         actionsFlow = viewModel.actions,
         pagingListItems = pagingListItems,
         sendIntent = { viewModel.sendIntent(it) },
+        onFolderClick = {
```
sbakhtiarov, line 63: I suggest to use existing API for communication between View and ViewModel. You can define another CellViewIntent type OnFolderClick and use existing sendIntent method to send it to ViewModel.
Or change existing intent OnFileClick to OnNodeClick and choose the handler in the ViewModel depending on the clicked node type (file / folder).

Garzas (approved):
Nice 🥇

features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
```diff
@@ -320,10 +363,17 @@ internal enum class CellError(val message: Int) {
     OTHER_ERROR(R.string.action_failed)
 }
 
-internal data class MenuOptions(
-    val file: CellFileUi,
-    val actions: List<FileAction>
-)
+sealed class MenuOptions {
+    data class FileMenuOptions(
```
sbakhtiarov: This split here is probably not necessary. Menu has the same UI for all node types, just actions are different.

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
```diff
@@ -122,12 +153,26 @@ fun ConversationFilesScreen(
...
+                actionsFlow = actions,
                 pagingListItems = pagingListItems,
-                sendIntent = { viewModel.sendIntent(it) },
-                downloadFileState = viewModel.downloadFileSheet,
-                fileMenuState = viewModel.menu,
+                sendIntent = sendIntent,
+                downloadFileState = downloadFileSheet,
+                menuState = menu,
                 isAllFiles = false,
+                onFolderClick = {
```
sbakhtiarov, line 161: I suggest to use the intent - action API and send this to a ViewModel via sendIntent, prepare the right destination and send new OpenFolder action back to the view for navigation.
This way we will have same approach for all user actions handling, move a small piece of logic (building the destination name) to the ViewModel and can add test case for it.

features/cells/src/main/java/com/wire/android/feature/cells/ui/createfolder/CreateFolderScreen.kt
```diff
@@ -85,17 +107,25 @@ fun CreateFolderScreen(
...
-                            onClick = {},
-                            state = WireButtonState.Default,
-                            clickBlockParams = ClickBlockParams(blockWhenSyncing = true, blockWhenConnecting = true),
+                            onClick = {
+                                createFolderViewModel.createFolder(
+                                    folderName = createFolderViewModel.fileNameTextFieldState.text.toString(),
+                                    onFailure = {
+                                        showErrorDialog.value = true
+                                    },
+                                    onSuccess = {
```
sbakhtiarov, line 116: Passing callbacks to the ViewModel like this dangerous and can cause crashes. It is not guaranteed that callback will be called when UI is in the right state for navigation.
ViewModel should handle success / error cases and notify UI with actions flow collected with lifecycle (see CellViewModel / CellScreenContent).

features/cells/src/main/java/com/wire/android/feature/cells/ui/dialog/FolderActionsBottomSheet.kt
```diff
@@ -0,0 +1,146 @@
...
+import com.wire.android.ui.common.bottomsheet.WireSheetValue
+import com.wire.android.ui.common.bottomsheet.rememberWireModalSheetState
+import com.wire.android.ui.common.colorsScheme
+import com.wire.android.ui.common.dimensions
+import com.wire.android.ui.common.divider.WireDivider
+import com.wire.android.ui.common.typography
+import com.wire.android.ui.theme.WireTheme
+import kotlinx.coroutines.launch
+
+@Composable
```
sbakhtiarov: Is this menu different from the one we are showing for files? Can we use one menu for both just with a different set of actions?

sbakhtiarov (changes requested):

features/cells/src/main/java/com/wire/android/feature/cells/ui/AllFilesScreen.kt
ohassine: We will not need the viewModel here as we will navigate to a different screen

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
ohassine: It's more extra work to do, pass intent from view to viewModel back to view again.
And ViewModels should be UI-agnostic, they should not know about navigation, screens..

features/cells/src/main/java/com/wire/android/feature/cells/ui/createfolder/CreateFolderScreen.kt
ohassine: Valid point, thanks

features/cells/src/main/java/com/wire/android/feature/cells/ui/dialog/FolderActionsBottomSheet.kt
ohassine: slight difference, FolderBottom sheet does not has a title and there are difference in copies.
But mainly made like that to not mix files actions with folders ones

features/cells/src/main/java/com/wire/android/feature/cells/ui/AllFilesScreen.kt
sbakhtiarov: maybe right now we do not have any extra logic, but we could have some checks in future (e.g. permissions) which would be better to do in the view model.
I guess the only real reason to handle it via the view model is to make this flow testable via view model unit tests.

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
sbakhtiarov: By passing it via viewmodel
- we will make this action testable
- user action handling logic will be in the viewmodel, not in the view

ViewModel must not know about navigation implementation details, but it should define the logic for handling user actions, like: 
user clicked item -> check if it is a folder -> send command to open folder

features/cells/src/main/java/com/wire/android/feature/cells/ui/AllFilesScreen.kt
ohassine: Let's keep it simple for now—no need to over-engineer prematurely. If we need to add permission checks or more complex logic later, we can refactor and move that into the ViewModel when it makes sense.

sbakhtiarov (approved):

PR #4026, opened 2025-05-19
feat: Move to folder (WPB-17274)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17274" title="WPB-17274" target="_blank">TaskWPB-17274</a>  [Android] Move a file/folder into folder
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Allow user to move a file/folder to a folder

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
```diff
@@ -276,20 +277,37 @@ class CellViewModel @Inject constructor(
         _menu.emit(menuOption)
     }
 
-    private fun onMenuFileAction(file: CellNodeUi.File, action: BottomSheetAction.File) {
-        when (action.action) {
+    private fun onMenuFileAction(file: CellNodeUi.File, action: BottomSheetAction) {
+        when ((action as BottomSheetAction.File).action) {
```
sbakhtiarov: Possible crash here is called with invalid action type. Can we keep the correct action parameter type here? Probably need to set the specific type in MenuFile/FolderActionSelected intents instead to avoid any type casting at all.

features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
```diff
@@ -276,20 +277,37 @@ class CellViewModel @Inject constructor(
...
+            }
         }
     }
 
-    @Suppress("UNUSED_PARAMETER")
-    private fun onMenuFolderAction(folder: CellNodeUi.Folder, action: BottomSheetAction.Folder) {
-        when (action.action) {
-            FolderAction.SHARE -> TODO()
-            FolderAction.MOVE -> TODO()
+    private fun onMenuFolderAction(folder: CellNodeUi.Folder, action: BottomSheetAction) {
```
sbakhtiarov: Same comment here as for the "onMenuFileAction" function.

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
```diff
@@ -0,0 +1,313 @@
...
+)
+@Composable
+fun MoveToFolderScreen(
+    navigator: WireNavigator,
+    moveToFolderNavArgs: MoveToFolderNavArgs,
+    modifier: Modifier = Modifier,
+    moveToFolderViewModel: MoveToFolderViewModel = hiltViewModel()
+) {
+
+    val isScreenLoading = remember {
```
sbakhtiarov: Somehow the loading state got duplicated. There is this isScreenLoading flag here and the MoveToFolderScreenState.Loading state.
I suggest keeping only the ViewModel state as a single source of truth for screen state and remove the mutable flag.

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
```diff
@@ -0,0 +1,313 @@
...
+        mutableStateOf(false)
+    }
+
+    val lifecycle = LocalLifecycleOwner.current
+    val context = LocalContext.current
+
+    LaunchedEffect(Unit) {
```
sbakhtiarov: This will request data from server every time we resume this screen. So if user selects a folder and then return, this code will request data from server again. Not sure if this is required. I would suggest moving it to the viewmodel so that we load data only once.

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
```diff
@@ -0,0 +1,313 @@
...
+                                state = WireButtonState.Default,
+                                clickBlockParams = ClickBlockParams(blockWhenSyncing = true, blockWhenConnecting = true),
+                            )
+                        }
+                    }
+                }
+            }
+        ) { innerPadding ->
+
+            val nodes = moveToFolderViewModel.nodes.collectAsState()
```
sbakhtiarov, line 156: What is the reason to have a split between nodes and uiState? We can have nodes list as a Success state property.

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
sbakhtiarov: Ok after more investigation I figured out that it not a loading state, but a flag for "move operation running" state. Please refactor this to make it clear. I suggest moving both "nodes" list and this flag (as well as renaming it to what it actually means) to properties of the Success state since then both are valid only in this state.

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class MoveToFolderViewModel @Inject constructor(
+    override val savedStateHandle: SavedStateHandle,
+    private val getNodesUseCase: GetNodesUseCase,
```
sbakhtiarov, line 46: I consider this a temporary solution before switching to "only folders" UI since this use case should not be called here. Maybe we can make it internal to avoid calling it from app in future?

sbakhtiarov (changes requested):

features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
ohassine: it should not be changed, it is back by mistake after resolving conflicts

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
ohassine: yes  you are right, moving it in one place..

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
ohassine: `uiState` is the UI state, can be (Loading, Success, Error). `Nodes` are the data to show

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderViewModel.kt
ohassine: yes it's handled in the another PR

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
sbakhtiarov: - I do not see any purpose for having nodes as "flow" here. Nodes are updated only once on initial data loading.
- Nodes list is only valid in Success state.

Sometimes there is a good reason to expose the data as a standalone flow, but I do not see it here.

sbakhtiarov (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
ohassine: we actually need it to be flow as we can create a folder when moving a file/folder

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
```diff
@@ -0,0 +1,321 @@
...
+
+@Composable
+private fun RowItem(
+    cell: CellNodeUi,
+    onFolderClick: (CellNodeUi.Folder) -> Unit = { }
+) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(56.dp)
```
saleniuk:
```suggestion
            .height(dimensions().spacing56x)
```

features/cells/src/main/java/com/wire/android/feature/cells/ui/LoadingScreen.kt
```diff
@@ -0,0 +1,67 @@
...
+@Composable
+fun FullScreenLoading(
+    modifier: Modifier = Modifier,
+) {
+    Box(
+        modifier = modifier
+            .fillMaxSize()
+            .pointerInput(Unit) {}
+            .semantics { invisibleToUser() }
+            .background(Color.Gray.copy(alpha = 0.6f)),
```
saleniuk, line 59:
```suggestion
            .background(colorsScheme().scrim),
```
please use `colorsScheme().scrim`

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderScreen.kt
```diff
@@ -0,0 +1,321 @@
...
+                    text = stringResource(R.string.no_subfolders_message),
+                    style = MaterialTheme.wireTypography.body01,
+                    color = colorsScheme().secondaryText,
+                    modifier = Modifier.padding(innerPadding)
+                )
+            }
+        } else {
+            LazyColumn(
+                modifier = Modifier.padding(innerPadding),
+                verticalArrangement = Arrangement.spacedBy(2.dp)
```
saleniuk:
```suggestion
                verticalArrangement = Arrangement.spacedBy(dimensions().spacing2x)
```

saleniuk (approved):
Just small requests to use our defined colors/dimensions

features/cells/src/main/java/com/wire/android/feature/cells/ui/LoadingScreen.kt
ohassine: this scrim value has 0.55 as aplha value

saleniuk: @ohassine can you show me the design for it with alpha 0.6?
I'm pretty sure if there is one then it's wrong because we are supposed to use only colors from design documentation and our color scheme, so if there was need for some grey color with alpha 0.6 then it should be first added to the documentation and our color scheme and it shouldn't be hardcoded `Color.Gray`

PR #4030, opened 2025-05-20
fix: align channel and group icon colors with other platforms (#WPB-17690)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17690" title="WPB-17690" target="_blank">BugWPB-17690</a>  [Android] Inconsistent channel icon color across clients (Android, iOS, Web)
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-17690

# What's new in this PR?

### Issues
Channel icon color on Android does not match color on iOS and Web clients.

### Causes (Optional)
iOS/Web use only conversation id value for calculating the color index.

### Solutions
Adjust color index calculation.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/common/AppThemeExt.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/common/AppThemeExt.kt
@@ -19,6 +19,7 @@
 package com.wire.android.ui.common
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Stable
 import com.wire.android.ui.theme.WireColorScheme
 import com.wire.kalium.logic.data.id.ConversationId
 import kotlin.math.absoluteValue
@@ -29,6 +30,7 @@ internal fun WireColorScheme.channelConversationColor(id: ConversationId) = chan
 @Composable
 internal fun WireColorScheme.groupConversationColor(id: ConversationId) = groupAvatarColors.withConversationId(id)
 
+@Stable
 private fun <T> List<T>.withConversationId(id: ConversationId): T {
     val hash = id.value.lowercase().hashCode()
     return this[hash.absoluteValue % this.size]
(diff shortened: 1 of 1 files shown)


ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/common/AppThemeExt.kt
```diff
@@ -19,20 +19,17 @@
...
+internal fun WireColorScheme.channelConversationColor(id: ConversationId) = channelAvatarColors.withConversationId(id)
 
 @Composable
-internal fun WireColorScheme.groupConversationColor(id: ConversationId): GroupAvatarColors {
-    val colors = this.groupAvatarColors
-    return colors[(id.hashCode() % colors.size).absoluteValue]
+internal fun WireColorScheme.groupConversationColor(id: ConversationId) = groupAvatarColors.withConversationId(id)
+
+private fun <T> List<T>.withConversationId(id: ConversationId): T {
+    val hash = id.value.lowercase().hashCode()
```
MohamadJaara, line 35: the lowercase is redundant since android will always have it lowercase, it is also worth anotating this as @Stable since for the same id it will always return the same color

MohamadJaara (changes requested):

app/src/main/kotlin/com/wire/android/ui/common/AppThemeExt.kt
sbakhtiarov: yes, I know about lowercase, but since it is important to have this in sync with other platforms I decided to add it to:
- ensure it is lowercase even if for some reason we get an uppercase id
- highlight that it is important to keep this lowercase to match other platforms

@MohamadJaara wdyt? can we keep it?

MohamadJaara (approved):

PR #4031, opened 2025-05-20
chore: add learn more link to the e2ei info text (#WPB-7131)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7131" title="WPB-7131" target="_blank">BugWPB-7131</a>  E2EI: "Learn more" link is missing when registering a new device 
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-7131

# What's new in this PR?

Adding "Learn more" link to the end of the E2EI info text


app/src/main/kotlin/com/wire/android/ui/e2eiEnrollment/E2EIEnrollmentScreen.kt
```diff
@@ -227,6 +217,8 @@ private fun E2EIEnrollmentScreenContent(
     }
 }
 
+private const val E2EI_INFO_URL = "https://support.wire.com/hc/en-us/articles/9211300150685-End-to-end-identity"
```
MohamadJaara: can we please have this link in the strings.xml file so it can be configured per app if there is a need to
e.g. https://github.com/wireapp/wire-android/blob/bd440050df34ec1a2f5e9704162ed2b653ce64bb/app/src/main/res/values/strings.xml#L266

MohamadJaara (changes requested):

app/src/main/kotlin/com/wire/android/ui/e2eiEnrollment/E2EIEnrollmentScreen.kt
yamilmedina: The link should be the non-localized version as well, so it is delegated to the browser locale as the rest of links from wire support.

https://support.wire.com/hc/articles/9211300150685

MohamadJaara (approved):
LGTM

yamilmedina (approved):

PR #4043, opened 2025-05-28
fix: error message not visible when trying to record audio during cal…

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-12178

# What's new in this PR?

### Issues
Recording an audio message is not allowed when there is an ongoing call. There are issues with showing the error message to the user.
- Snackbar message is not visible / partially visible
- Error is shown when trying to record the audio message. According to the design the app should show the error when user selects "Audio" from attachment options.

### Solutions
- Observe ongoing calls and add `isCallOngoing` property to the MessageComposerViewState
- Show toast message (guaranteed to be displayed on top of all views) when "Audio" is selected in attachment options and `isCallOngoing == true`


saleniuk (approved):

MohamadJaara (approved):
LGTM

PR #4044, opened 2025-05-28
fix: crash when tapping start conversation button many times (#WPB-17923)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17923" title="WPB-17923" target="_blank">BugWPB-17923</a>  [Android] App freezing and crashing when tapping start conversation multiple times
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-17923

# What's new in this PR?

### Issues
App crashing when tapping fast on the start conversation button in user profile screen.

### Causes (Optional)
Each tap starts a new coroutine trying to create a new conversation.

### Solutions
- View model should not call use case if action is already in progress. Fixed by checking the flag in the view state and ignoring the button click if action is in progress.
- [General] Fast multiple clicks on the button should be ignored. Fixed by adding a SingleClickHandler filtering multiple subsequent button clicks.
- [Additional] View must not pass lambdas to the view model for callbacks to async actions. Fixed by replacing lambdas with the actions flow handled in view with lifecycle.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/connection/ConnectionActionButton.kt
@@ -26,11 +26,15 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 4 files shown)


saleniuk (approved):

MohamadJaara (approved):
LGTM

PR #4053, opened 2025-06-04
feat(new-registration): new registration account type selector (WPB-17410)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17410" title="WPB-17410" target="_blank">StoryWPB-17410</a>  [Android] New Registrations Flows - Account creation type selector
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

New registration flows needs to be implemented, this PR includes:

- Feature flag to hide/show the new registration flow
- Account type selector according to specs at **WPB-17410** redirecting to the current team and personal flows.
- Rename and reusage of `NewAuthContainer` as it follows the new branding waves, as in new login.

### Dependencies (Optional)

None

#### Test Coverage (Optional)

#### How to Test

On internal builds, you should be able to see the account selector type, redirecting to the corresponding flows, team or personal.

### Attachments (Optional)

https://github.com/user-attachments/assets/b7e3259a-009f-438b-84da-f86fd82a044e

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/registration/selector/CreateAccountSelectorNavArgs.kt
```diff
@@ -0,0 +1,41 @@
...
+import dev.ahmedmourad.bundlizer.Bundlizer
+import kotlinx.parcelize.Parceler
+import kotlinx.parcelize.Parcelize
+import kotlinx.parcelize.TypeParceler
+
+@Parcelize
+@TypeParceler<ServerConfig.Links?, ServerConfigLinksParceler>()
+data class CreateAccountSelectorNavArgs(val customServerConfig: ServerConfig.Links? = null) : Parcelable
+
+object ServerConfigLinksParceler : Parceler<ServerConfig.Links?> {
```
saleniuk, line 32: Can we just reuse the one that's already created in `CreateAccountOverviewNavArgs.kt` as this one is a 1:1 copy of it, right? Maybe we should just move it to a dedicated class file 😄

saleniuk (approved):
Nice! 👏 Just one suggestion to reuse instead of duplicating code 😄

app/src/main/kotlin/com/wire/android/ui/registration/selector/CreateAccountSelectorNavArgs.kt
yamilmedina: That's a good point. I just duplicated it for now for 2 reasons, to maintain old code untouched since it's under a feature flag, and we want to merge sooner rather than keep in an epic, and second, it will change, since in the next PR it will have more parameters, given we need to carry over the email

sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/ui/registration/selector/CreateAccountSelectorNavArgs.kt
saleniuk: But this email won't be a part of `ServerConfig.Links`, right? 😄 Even if we would need to add some fields to `ServerConfig.Links` then both parcelers for this type would need to be updated, so it would be beneficial to have it in a single place.

If email will be added to `CreateAccountSelectorNavArgs` as a simple string then this `ServerConfigLinksParceler` won't change and these nav args will still be parcelable. 😉 

You can even just use the existing one here instead of duplicating 🤔 , just add this import:
`import com.wire.android.ui.authentication.create.overview.ServerConfigLinksParceler`

yamilmedina: Ok, I get what you mean, anyway I will remove all the code under  `...ui.authentication.create....` :)  when the new reg is finished...but for now to address this point I will then reuse it by inverting the deps, will keep the one in this file and update the imports in the old flow

yamilmedina: Except `..common..` 😅  that we need to maintain and move around

PR #4058, opened 2025-06-10
refactor: remove navigation lambdas from view models - part 3 [WPB-17842]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17842" title="WPB-17842" target="_blank">TaskWPB-17842</a>  [Android] Remove illegal usage of the android navigation from view model
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Follow-up to these PRs: 
- https://github.com/wireapp/wire-android/pull/3994
- https://github.com/wireapp/wire-android/pull/3997
- https://github.com/wireapp/wire-android/pull/4054
- https://github.com/wireapp/wire-android/pull/4055

### Issues

We are passing lambdas which execute navigation into view models. 
After recreating the Activity (for instance when rotation changes), the navigation may be executed in wrong lifecycle state resulting in a crash.

### Solutions

Replace passing lambdas with navigation commands to ViewModel which calls them after making some actions, instead use mutableState or flow with lifecycle-aware collector to ensure navigation command is executed when UI is in valid lifecycle state.

In this PR, following view models are refactored out of using navigation lambdas:
- ConversationCallListViewModel
- ConversationListCallViewModel
- IncomingCallViewModel
- RecordAudioViewModel
- SharedCallingViewModel

Also, renamed first two as their names were wrong:
- ConversationCallListViewModel -> ConversationListCallViewModel as it's used in "conversations list" composable and it's purpose is calls, currently it wrongly suggests that it's used in "conversation call list" so list of conversation calls
- ConversationListCallViewModel -> ConversationCallViewModel as it was renamed by mistake instead of the one above and it shouldn't have "list" in the name as it's used in "conversation" composable, not "conversations list" 

#### Test Coverage (Optional)

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
```diff
@@ -222,12 +228,12 @@ class SharedCallingViewModel @AssistedInject constructor(
         }.collect()
     }
 
-    fun hangUpCall(onCompleted: () -> Unit) {
+    fun hangUpCall() {
         viewModelScope.launch {
```
sbakhtiarov: Not related to the lambdas issue, but since we are changing this part of code.
I see some call finalization logic here and in resetCallConfig function. I think it would be great to move it altogether to a CallHangUpUseCase.
The call hang-up logic and clean-up seems important but it is connected to the ViewModel scope and there is a chance it will be cancelled if user closes the screen. Would be better to do it in a NonCancellable coroutine.

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
```diff
@@ -159,12 +162,15 @@ fun OngoingCallScreen(
         }
     }
 
-    val hangUpCall = remember {
-        {
-            sharedCallingViewModel.hangUpCall { activity.finishAndRemoveTask() }
+    LaunchedEffect(Unit) {
```
sbakhtiarov, line 165: We have this pattern now in most of the ViewModels:
- _actions / actions flows
- sendAction function
- collecting actions in compose code

Would be nice to move it to some common code (both for viewmodels and collecting part).

sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/OngoingCallScreen.kt
saleniuk: Sure, I have the code ready to do that and will raise another PR with this clean-up after all 3 parts are merged to apply it to all view models that use it

app/src/main/kotlin/com/wire/android/ui/calling/SharedCallingViewModel.kt
saleniuk: Great suggestion! 👍 
I did that in a separate PR here: https://github.com/wireapp/wire-android/pull/4063

Garzas (approved):

PR #4072, opened 2025-06-19
feat(cells): Show breadcrumb navigation during folder browsing (WPB-18241)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18241" title="WPB-18241" target="_blank">TaskWPB-18241</a>  [Android] Display breadcrumb path when navigating through folders
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Display the path when the user is navigating through folders

[Screen_recording_20250619_084227.webm](https://github.com/user-attachments/assets/57db276c-1723-4c3e-9341-8a29c0f2b825)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
```diff
@@ -141,18 +149,33 @@ fun ConversationFilesScreenContent(
...
+                    elevation = dimensions().spacing0x,
+                    actions = {
+                        MoreOptionIcon(
+                            contentDescription = R.string.content_description_conversation_files_more_button,
+                            onButtonClicked = { optionsBottomSheetState.show() }
+                        )
+                    }
+                )
+                breadcrumbs?.let {
+                    LazyRow(
```
saleniuk: Maybe we could reverse the breadcrumbs so that they are always scrolled to the end to the current folder and not to the root, that I think would be better for the UX 🤔 
```suggestion
                breadcrumbs?.reversedArray()?.let {
                    LazyRow(
                        reverseLayout = true,
```

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
```diff
@@ -141,18 +149,33 @@ fun ConversationFilesScreenContent(
...
+                        modifier = Modifier
+                            .height(dimensions().spacing40x)
+                            .fillMaxWidth(),
+                        contentPadding = PaddingValues(
+                            start = dimensions().spacing16x,
+                            end = dimensions().spacing16x
+                        ),
+                    ) {
+                        item { Breadcrumbs(it) }
```
saleniuk: So the `LazyRow` is used but it contains only a single element which again uses `Row` inside to compose all elements from the array - it doesn't make sense.
Either `Breadcrumbs` composable should use `LazyRow` inside instead of a simple `Row` and this one should be just removed, or just make this `Row` that's inside `Breadcrumbs` scrollable by using `modifier.horizontalScroll(...)`, or keep this `LazyRow` here but create a composable for a single `Breadcrumb` and use it here for each item of the `breadcrumbs` array.

MohamadJaara:
question: when going deep into folders should the top bar auto scroll to the current folder?

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
ohassine: I will need to check with the design team for this case if we need to show last item on the list or the root

ohassine:
> question: when going deep into folders should the top bar auto scroll to the current folder?

I will need to check with the design team for this case

saleniuk (approved):

MohamadJaara (approved):

PR #4087, opened 2025-06-27
fix: request 2fa code when removing client (#WPB-11401)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11401" title="WPB-11401" target="_blank">StoryWPB-11401</a>  [Android] Handle the case where client register got failed/interacted when 2FA is active
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-11401

# What's new in this PR?

### Issues
User logs in and need to remove client because he already has too many. No 2FA core is requested if new client registration fails with "Verification Code Missing" error. See ticket for exact scenario to reproduce.

### Causes (Optional)
2FA errors not handled when registering new client after removing client.

### Solutions
Request 2FA verification code if registering new client fails with "Verification Code Missing" error.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/authentication/devices/remove/RemoveDeviceScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/authentication/devices/remove/RemoveDeviceScreen.kt
@@ -19,7 +19,10 @@
 package com.wire.android.ui.authentication.devices.remove
 
 import androidx.activity.compose.BackHandler
+import androidx.compose.animation.AnimatedContent
+import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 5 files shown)


saleniuk (approved):

MohamadJaara (changes requested):
question: in this case do we need to send a request for a new otp?

sbakhtiarov:
> question: in this case do we need to send a request for a new otp?

This is what the `request2FACode()` method is doing, but I think it is not always results in a new OTP code generated.

MohamadJaara (approved):

MohamadJaara:
> > question: in this case do we need to send a request for a new otp?
> 
> This is what the `request2FACode()` method is doing, but I think it is not always results in a new OTP code generated.

my bad i missed this one, yes this is correct, it will generate new one iff the old one expired,, as long as we call it we should be good

PR #4093, opened 2025-07-02
fix: reply from notification (#WPB-18533)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18533" title="WPB-18533" target="_blank">BugWPB-18533</a>  [Android] Cannot reply from notification
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-18533

# What's new in this PR?

### Issues
Replying from notification is not working. Infinite progress is shown.

### Causes (Optional)
NotificationReplyReceiver is calling SendTextMessageUseCase which is waiting for the sync Live state. But the sync is not triggered from the broadcast receiver.

### Solutions
1. Request sync when sending reply
2. Add support for cancellation if sync and sending operation take too long to avoid ANR.


saleniuk (approved):

vitorhugods (approved):
👨🏻‍🍳🤌🏻

Garzas (approved):

PR #4131, opened 2025-07-18
fix: clearing content actions order and execute local deletion in background [WPB-18778]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18778" title="WPB-18778" target="_blank">BugWPB-18778</a>  [Android] Fix order of “delete locally” worker and “clear content” actions
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The order of actions when clearing content was invalid and because of that, asset files wasn't being cleared.
Removing conversation locally can take some time and but the actual deletion happens at the very end so the app still needs to wait until all is executed and block the user from making any actions to the conversation that's being deleted locally.

### Solutions

Adjust the app to fixes from https://github.com/wireapp/kalium/pull/3532 which fixes the order of actions when clearing content so that asset files are cleared properly.
Split deleting conversation locally into two steps. First mark the conversation as deleted locally using new use case which removes this conversation from the app (not appearing on lists and cannot fetch conversation details) and then execute the deletion using the same use case as before, but now it can be executed in the background.

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/3532

#### Test Coverage (Optional)

#### How to Test

Create a group, send some heavy files, leave group, remove group locally. Group should disappear from list right away and all assets sent in this group should be cleared eventually.

### Attachments (Optional)

https://github.com/user-attachments/assets/f575fb6b-4c9e-4c29-950a-844277642b4a

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/conversation/ConversationOptionsMenuViewModel.kt
```diff
@@ -106,6 +103,7 @@ interface ConversationOptionsMenuViewModel : ActionsManager<ConversationOptionsM
 @Suppress("LongParameterList", "TooManyFunctions")
 @HiltViewModel
 class ConversationOptionsMenuViewModelImpl @Inject constructor(
```
sbakhtiarov, line 107: I am a bit concerned about the way we are calling use cases in this View Model. We start potentially long running operations (deleting / leaving conversations, clearing content) in the view model scope. Which means that they will be cancelled once user leaves the screen.
I think it is important to make sure these operations are executed in a different scope (user session scope?) or with NonCancellable context.

app/src/main/kotlin/com/wire/android/workmanager/worker/DeleteConversationLocallyWorker.kt
```diff
@@ -54,29 +55,34 @@ import kotlinx.coroutines.flow.mapNotNull
...
+        val userId = qualifiedIdMapper.fromStringToQualifiedID(userIdString)
+        coreLogic.getGlobalScope().doesValidSessionExist(userId).let {
+            if (it is DoesValidSessionExistResult.Failure || (it is DoesValidSessionExistResult.Success && !it.doesValidSessionExist)) {
+                return@coroutineScope Result.failure() // If no valid session exists, fail the work
             }
-        } ?: Result.failure()
+        }
+
+        coreLogic.getSessionScope(userId).conversations.deleteConversationLocallyUseCase(conversationId)
+            .fold({ Result.failure() }, { Result.success() })
```
sbakhtiarov: Should we retry errors? What if user tries to leave / delete conversation without Internet connection? Currently we are showing error message and keep conversation. Will it work with the new background worker?

app/src/main/kotlin/com/wire/android/ui/common/bottomsheet/conversation/ConversationOptionsMenuViewModel.kt
saleniuk: Some important actions like deleting, leaving, clearing content show a dialog so in theory there's one additional step that prevents user from closing the screen, but on the other hand when the user starts some action but then clicks "cancel" then it should stop that, but stopping in the middle can potentially break something so I think it's better to also block dismissing this dialog until it's completed, so I did both: added non-cancellable context and added keeping dialog until the action is completed.

app/src/main/kotlin/com/wire/android/workmanager/worker/DeleteConversationLocallyWorker.kt
saleniuk: This worker is used only when deleting conversation locally so when the user has already left the conversation, so it's just to remove this conversation from list and clear its content. I just found out that there is one action that requires network there - sending "clear content and remove locally" message, so this worker was missing connected network constraint even before my changes, so now I added this constraint. 
I just missed the part that it sends a message and thought it solely executes some local db queries so in theory it shouldn't fail and if it failed then it would mean that there's something wrong in db and the retry won't help. But in this case maybe you're right with changing it to "retry" so that the backoff mechanism can reschedule it if sending a message fails and make sure it's completed eventually. Thanks for pointing it out.

Garzas (approved):

sbakhtiarov (approved):

PR #4147, opened 2025-07-25
fix: crash on logout [WPB-18706]

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18706" title="WPB-18706" target="_blank">BugWPB-18706</a>  [Android] Crash when account was deleted or device was removed
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-18706

# What's new in this PR?

### Issues
Application crash when current client is deleted from other device.

### Causes (Optional)
1. Current client removal triggers logout and clears data. 
2. Data clear resets the "initialSyncCompleted" flag and HomeViewModel navigates to InitialSyncScreen. 
3. This screen is not supposed to be called with no session and crashes on dependencies setup.

### Solutions
Check current session availability in HomeViewModel before navigating to InitialSyncScreen.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
@@ -31,12 +31,16 @@ import com.wire.android.model.ImageAsset.UserAvatarAsset
 import com.wire.android.model.NameBasedAvatar
 import com.wire.android.model.UserAvatarData
 import com.wire.android.navigation.SavedStateViewModel
+import com.wire.kalium.logic.data.auth.AccountInfo
 import com.wire.kalium.logic.data.user.SelfUser
 import com.wire.kalium.logic.feature.client.NeedsToRegisterClientUseCase
 import com.wire.kalium.logic.feature.legalhold.LegalHoldStateForSelfUser
 import com.wire.kalium.logic.feature.legalhold.ObserveLegalHoldStateForSelfUserUseCase
 import com.wire.kalium.logic.feature.personaltoteamaccount.CanMigrateFromPersonalToTeamUseCase
+import com.wire.kalium.logic.feature.session.CurrentSessionFlowUseCase
+import com.wire.kalium.logic.feature.session.CurrentSessionResult
 import com.wire.kalium.logic.feature.user.ObserveSelfUserUseCase
+import dagger.Lazy
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.collectLatest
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/ui/home/HomeViewModel.kt
```diff
@@ -124,7 +129,7 @@ class HomeViewModel @Inject constructor(
                 needsToRegisterClient() -> // check if the client needs to be registered
                     onRequirement(HomeRequirement.RegisterDevice)
 
-                !dataStore.initialSyncCompleted.first() -> // check if the initial sync needs to be completed
+                !dataStore.initialSyncCompleted.first() && isLoggedIn() -> // check if the initial sync needs to be completed
```
saleniuk: Nice! But I have one thought about that: 
in theory we check if `selfUser` is null before that in line number 124 so it shouldn't even get to that point because after logging out, `selfUser` should be null, right? There's definitely something fishy with that and now we'll have two different checks for that, one clearly not working as expected.
I think the problem with `selfUserFlow` is that it uses `ObserveSelfUserUseCase` which doesn't return null but we create nullable flow out of it. 😅 This `selfUserFlow` is used only here, I think it was created exactly for the scenario that you are fixing now, so maybe we could just completely remove this `selfUserFlow` and replace it with `isLoggedIn()` - what do you think?

sbakhtiarov: HomeViewModel is caching the selfUser:
`private val selfUserFlow = MutableSharedFlow<SelfUser?>(replay = 1)`
I think after clearing data the original flow from the database will not return any new values so we will keep using a cached user.
Maybe the better solution will be to clean this up and clear this user on logout?

sbakhtiarov: The flow does not emit null or error, it just completes once database is cleared. But this happens later, only when user closes the logout dialog. So this way we cannot determine if user is logged-out.

sbakhtiarov: Maybe it is better to check both?
```
            val selfUser = selfUserFlow.firstOrNull() ?: return@launch
            if (!isLoggedIn()) return@launch

```

saleniuk: Yep, it's caching, but if `ObserveSelfUserUseCase` emitted nulls then `selfUserFlow` would get this null and update itself and then we wouldn't have this crash. 😄 
 
> Maybe the better solution will be to clean this up and clear this user on logout?

As I said, this `selfUserFlow` is only used in `checkRequirements`, that's why I suggested that we could remove `selfUserFlow` completely from this view model and replace
`val selfUser = selfUserFlow.firstOrNull() ?: return@launch`
with
`if (!loggedIn()) return@launch`
in the beginning of the `checkRequirements` method.

saleniuk: Oh, okay, just now I noticed that we do also have `selfUser.handle.isNullOrEmpty() ->` at the end, so we actually use this `selfUser` later as well. Sorry, my bad. 😓 
So yeah, this shared flow could be cleared on logout, so did you have something like that in mind, or some other solution?
```
private fun observeLogout() {
    viewModelScope.launch {
        currentSessionFlow.value.invoke()
            .filter { !(it is CurrentSessionResult.Success && it.accountInfo is AccountInfo.Valid) }
            .collect { selfUserFlow.emit(null) }
    }
}
```

saleniuk: Maybe I just overcomplicated it, so going back to the beginning, I think just replacing 
`val selfUser = selfUserFlow.firstOrNull() ?: return@launch` with `if (!loggedIn()) return@launch` or adding it after but at the same time keeping the `selfUserFlow` as it is now should be enough, because if it gets logged out then `checkRequirements` is returned before going into `when`, and if not then it just executes it with current self user data.

sbakhtiarov: updated

saleniuk (approved):

yamilmedina (approved):

PR #4159, opened 2025-08-04
fix: ending 1:1 call before established breaks next calls from that person [WPB-18304]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18304" title="WPB-18304" target="_blank">BugWPB-18304</a>  [Android] If a user ends a call during setup phase, repeated call attempts from same user fail until reset.
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

If a user ends a call during setup phase, repeated call attempts from same user fail until reset.

### Causes (Optional)

The `callServiceEvents` in `ServiceManager` is in the wrong state because it responds to changes on established and ongoing calls, so non-established call doesn't trigger any action and makes it so that it becomes out of sync with the `CallService`.

### Solutions

- extract logic of handling actions into dedicated manager class that’s used by `CallService`
- include observing incoming call for non-current session as well in case the call service needs to answer a call for some other session (currently it was using only current session scope so the service could get stopped improperly in that scenario)
- observe the `CallService` state to clear `callServiceEvents` when the service is stopped (it means there are no active calls) so that the `ServiceManager` is synced with the service and can start it again properly
- remove stopping the `CallService` from `WireNotificationManager`, the service itself is responsible for stopping when needed so this manager should only handle starting the service if there are some established or outgoing calls

#### Test Coverage (Optional)

#### How to Test

Receive an incoming 1:1 call, answer it and close while it's still connecting. After that, receive another 1:1 incoming call from the same person, answer it - it shouldn't be stuck in "connecting" state.

### Attachments (Optional)

| Before | After |
| <video width="400" src="https://github.com/user-attachments/assets/f21791b4-6795-497e-a4b6-03ca740f7c8d"/> | <video width="400" src="https://github.com/user-attachments/assets/b06faee7-f88d-4621-9ffc-0b66819e1bdd"/> |

----

----

----
##### References


vitorhugods (approved):
👨🏻‍🍳🤌🏻

sbakhtiarov (approved):

PR #4163, opened 2025-08-06
chore: debug feature flags screen (WPB-19327)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19327" title="WPB-19327" target="_blank">TaskWPB-19327</a>  [Android] Feature flags debug screen
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19327

# What's new in this PR?

Debug screen for showing current feature configuration on the server.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/debug/featureflags/DebugFeatureFlagsScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/debug/featureflags/DebugFeatureFlagsScreen.kt
@@ -1,0 +1,84 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.android.ui.debug.featureflags
+
+import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/com/wire/android/ui/debug/featureflags/DebugFeatureFlagsScreen.kt
```diff
@@ -0,0 +1,81 @@
...
+        },
+        content = { paddingValues ->
+
+            val state by viewModel.state.collectAsState()
+
+            Column(
+                modifier = Modifier
+                    .fillMaxSize()
+                    .padding(paddingValues)
+                    .verticalScroll(rememberScrollState())
```
saleniuk: If you extract this scroll state above the `WireScaffold`:
`val scrollState = rememberScrollState()`
and use it in the `WireCenterAlignedTopAppBar` to calculate elevation:
```
WireCenterAlignedTopAppBar(
    elevation = scrollState.rememberTopBarElevationState().value,
    ...
```
and here then just do
`.verticalScroll(scrollState)`
then the whole screen will have a nice animation when scrolling (we use that on multiple screens) 😊

app/src/main/kotlin/com/wire/android/ui/debug/featureflags/FeatureListItem.kt
```diff
@@ -0,0 +1,118 @@
...
+    Column(
+        modifier = modifier
+            .fillMaxWidth()
+            .alpha(if (feature.status == FeatureStatus.ENABLED) 1f else 0.5f),
+    ) {
+        RowItem(
+            modifier = Modifier
+                .fillMaxWidth()
+                .defaultMinSize(minHeight = 48.dp)
+                .padding(16.dp),
```
saleniuk: This `FeatureListItem` will be used only on debuggable builds, right? Still, it would be nice to make it a habit of using `dimensions()` instead of hardcoding values 😄

saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/debug/featureflags/FeatureListItem.kt
sbakhtiarov: damn, always forget about dimensions, typography or colorscheme

yamilmedina (approved):

PR #4167, opened 2025-08-07
fix: Flickering when opening conversation details (WPB-19187)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19187" title="WPB-19187" target="_blank">BugWPB-19187</a>  [Android] Cells - Flickering when we open conversation details
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Screen is flickering when opening conversation details

### Causes (Optional)

Fetching data from data source, which lead to show the UI the default value.  and once the data is retrieved, UI gets updated again 

### Solutions

Add a skeleton UI to be shown while we fetch data

| Before | After |
| <video width="400" src="https://github.com/user-attachments/assets/5d137a06-391b-4af8-982d-a3c3b5735400"/> | <video width="400" src="https://github.com/user-attachments/assets/1c9d63a5-b361-443b-b7b5-a7b8e7868c8e"/> |

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


core/ui-common/src/main/kotlin/com/wire/android/ui/common/button/LoadingWireSecondaryButton.kt
```diff
@@ -0,0 +1,144 @@
...
+                        .width(dimensions().spacing16x)
+                        .height(dimensions().spacing16x)
+                        .background(
+                            color = colorsScheme().secondaryButtonEnabledOutline,
+                            shape = RoundedCornerShape(size = dimensions().corner100x)
+                        )
+                        .shimmerPlaceholder(
+                            visible = true,
+                            color = colorsScheme().surfaceContainerHighest,
+                            shape = RoundedCornerShape(size = dimensions().corner100x)
```
saleniuk: we should probably use the same size, so `dimensions().buttonCornerSize`

core/ui-common/src/main/kotlin/com/wire/android/ui/common/button/LoadingWireSecondaryButton.kt
```diff
@@ -0,0 +1,144 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import com.wire.android.ui.common.colorsScheme
+import com.wire.android.ui.common.dimensions
+import com.wire.android.ui.common.preview.MultipleThemePreviews
+import com.wire.android.ui.common.shimmerPlaceholder
+import com.wire.android.ui.theme.WireTheme
+
+@Composable
+fun LoadingWireSecondaryButton(
```
saleniuk, line 40: I think again it would be better to add a parameter `shimmerPlaceholder: Boolean` to the `WireSecondaryButton` instead of creating a direct copy of this class just to add one modifier, or even we could just add this modifier to the whole `WireSecondaryButton`:
```
WireSecondaryButton(
    modifier = Modifier.shimmerPlaceholder(
        enabled = true, 
        color = wireSecondaryButtonColors().enabled, 
        shape = RoundedCornerShape(size = dimensions().buttonCornerSize)
    )
```
It would also look good as the whole button would shimmer instead of a text inside it 🤔

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsScreen.kt
```diff
@@ -369,24 +380,51 @@ private fun GroupConversationDetailsContent(
...
+            AnimatedVisibility(
+                visible = !isScreenLoading.collectAsState().value,
+                enter = fadeIn(
+                    animationSpec = tween(
+                        durationMillis = 500,
+                        delayMillis = 100
+                    )
+                ) + slideInVertically(initialOffsetY = { it / 2 }),
+                exit = fadeOut()
+            ) {
```
saleniuk: Instead of having two `AnimatedVisibility` you can use `AnimatedContent` to animate changing different composables, like:
```
AnimatedContent(
    targetState = isScreenLoading.collectAsState().value,
    transitionSpec = {
        val enter = slideInVertically(initialOffsetY = { it })
        val exit = slideOutVertically(targetOffsetY = { it })
        enter.togetherWith(exit)
    },
) { isScreenLoading ->
     if (isScreenLoading) ...
}
```
You can even distinct animations for each state if it's needed:
```
transitionSpec = { 
        val isScreenLoading = this.targetState
        val enter = if (isScreenLoading) {
            slideInVertically(initialOffsetY = { it })
        } else {
            fadeIn(tween(durationMillis = 500, delayMillis = 100)) + slideInVertically(initialOffsetY = { it / 2 }),
        }
        val exit = fadeOut()
        enter.togetherWith(exit)
    },
```

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsTopBarCollapsing.kt
```diff
@@ -131,6 +138,83 @@ fun GroupConversationDetailsTopBarCollapsing(
     }
 }
 
+@Composable
+fun LoadingGroupConversationDetailsTopBarCollapsing(
```
saleniuk, line 142: I would prefer to re-use `GroupConversationDetailsTopBarCollapsing` with some mock data and add a parameter `shimmerPlaceholder: Boolean` to it which adds `.shimmerPlaceholder` to the children than to create second class just for the loading state, otherwise we will need to maintain both and that's actually a common mistake that people tend to forget about classes like this one.

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsViewModel.kt
```diff
@@ -86,11 +86,16 @@ class GroupConversationDetailsViewModel @Inject constructor(
...
     ActionsManager<GroupConversationDetailsViewAction> by ActionsManagerImpl() {
+
     private val groupConversationDetailsNavArgs: GroupConversationDetailsNavArgs = savedStateHandle.navArgs()
     val conversationId: QualifiedID = groupConversationDetailsNavArgs.conversationId
+
     private val _groupOptionsState = MutableStateFlow(GroupConversationOptionsState(conversationId))
     val groupOptionsState: StateFlow<GroupConversationOptionsState> = _groupOptionsState
 
+    private val _isFetchingInitialData: MutableStateFlow<Boolean> = MutableStateFlow(true)
+    val isFetchingInitialData: MutableStateFlow<Boolean> = _isFetchingInitialData
```
saleniuk, line 95: `GroupConversationOptionsState` contains default values for each field apart `conversationId` which we already know, so we could just add this state as a field `isFetchingInitialData` to `GroupConversationOptionsState` instead of having two separate state flows, like:
```
private val _groupOptionsState = MutableStateFlow(
    GroupConversationOptionsState(
        isFetchingInitialData = true
    )
)
```
and then just set `isFetchingInitialData = false` when updating state after fetching

core/ui-common/src/main/kotlin/com/wire/android/ui/common/button/LoadingWireSecondaryButton.kt
ohassine: I made the shimmer logic in a separate button to avoid adding conditional visuals to WireSecondaryButton, which is meant to stay simple and functional. 
Also, The button’s primary responsibility is to be an interactive UI element, which the shimmer version is not. 
So Keeping them separate will maintain a clear separation of concerns and keeps the base button focused on interaction not on loading.
..
In the design, only the content inside the button has the shimmer effect, not the entire button.

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsTopBarCollapsing.kt
ohassine: I'd prefer keeping the shimmer version as a separate layout — it keeps things cleaner and more focused. 
Also, handling the animation when shimmer fades out and real data appears is easier and smoother this way, compared to having it as a parameter inside the composable.

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsViewModel.kt
ohassine: I placed it separately intentionally because `isFetchingInitialData` relates specifically to the UI state(Loading or not), not the core GroupConversationOptions data we retrieve from data source.

sbakhtiarov (approved):

MohamadJaara:
the question here why does it take some time to fetch data? is it waiting on an API call ?

ohassine:
> the question here why does it take some time to fetch data? is it waiting on an API call ?

because we have a complex query with 11 joins, which takes at least 300ms

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsViewModel.kt
saleniuk: But `isFetchingInitialData` is also one of the possible states of this `GroupConversationDetailsScreen`. 😄  I think the best would be to make `GroupConversationOptionsState` a sealed class with two possibilities "Loading" and "Loaded" or something, because `isFetchingInitialData` is directly related - either one or another is true so keeping them as separate states may lead to some confusion and we would eliminate the situation which we have right now when `GroupConversationOptionsState` is filled with some initial empty values instead of proper data, but I don't insist as it doesn't affect the way the app works.

core/ui-common/src/main/kotlin/com/wire/android/ui/common/button/LoadingWireSecondaryButton.kt
saleniuk: About adding conditional visuals and being interactive, we already have "loading" parameter and multiple states which change the look of buttons and we already have "disabled" state which makes them not interactive so it's not true that we keep it separate. 😄 We rather prefer usability than purity in this regard and it would actually fit into our logic to add `WireButtonState.Init` or some other name. 
I'm just worried that we add this many additional lines of code which we need to maintain from now on just to add one quick animation effect to the button, but I also get your point of view, if it's just a single button then I guess it's fine but if we were to have shimmer for all buttons, we would need to find a better way to avoid having twice as many elements to maintain.

MohamadJaara:
> > the question here why does it take some time to fetch data? is it waiting on an API call ?
> 
> because we have a complex query with 11 joins, which takes at least 300ms

for getting one conversation details this is not great we can do better, lets also work on the query

saleniuk (commented):
I get your point of view about the loading shimmer composables, there's still one unresolved comment about `AnimatedContent` 🙃

saleniuk (approved):

PR #4183, opened 2025-08-13
feat: enable interaction of apps/services based on team feature flag configuration (WPB-18033)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18033" title="WPB-18033" target="_blank">StoryWPB-18033</a>  [Android] Connect to App flow from New conversation page
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Apps needs the capability to be searchable from the new button.

### Causes (Optional)

We had it hidden

### Solutions

- Implement new changes according to design, update copies
- Allow use apps based on team feature flag (disabled by default), rather than protocol (MLS/proteus) as it was.

### Depends on

- https://github.com/wireapp/kalium/pull/3586

### Attachments (Optional)

[Screen_recording_20250813_175554.webm](https://github.com/user-attachments/assets/f104f4c4-61c2-44c0-90a0-3b7d011ed6f6)

----

----

----
##### References


saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/SearchUsersAndServicesScreen.kt
```diff
@@ -134,14 +140,14 @@ fun SearchUsersAndServicesScreen(
...
-                searchBarHint = stringResource(R.string.label_search_people),
+                searchBarHint = searchBarTitle,
                 searchBarDescription = stringResource(R.string.content_description_add_participants_search_field),
                 searchQueryTextState = searchBarState.searchQueryTextState,
                 onActiveChanged = searchBarState::searchActiveChanged,
             )
         },
         topBarFooter = {
-            if (screenType == SearchPeopleScreenType.CONVERSATION_DETAILS && isServicesAllowed) {
+            if (isAppDiscoveryAllowed) {
```
ohassine, line 150: we don't need to check for `SearchPeopleScreenType.CONVERSATION_DETAILS `?

app/src/main/kotlin/com/wire/android/ui/home/newconversation/NewConversationViewModel.kt
```diff
@@ -78,26 +80,31 @@ class NewConversationViewModel @Inject constructor(
...
         getWireCellFeatureState()
     }
 
+    private fun observeAllowanceOfAppsUsage() {
+        viewModelScope.launch {
+            observeIsAppsAllowedForUsage()
+                .collectLatest { appsAllowed ->
+                    groupOptionsState = groupOptionsState.copy(
+                        isAppsUsagePossible = appsAllowed,
+                        isAllowAppsEnabled = appsAllowed
```
ohassine: Since I don't have full context on this, it's unclear what the distinction is between isAppsUsagePossible and isAllowAppsEnabled, especially they're both set to appsAllowed. Could we consider renaming one or clarifying their roles?

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/SearchUsersAndServicesScreen.kt
yamilmedina: We have new designs :D Before this component was used with the tab headers hidden. Now we need to display it, so now it is safe/needed to remove this check.

app/src/main/kotlin/com/wire/android/ui/home/newconversation/NewConversationViewModel.kt
yamilmedina: Renamed `isAppsUsagePossible` to `isTeamAllowedToUseApps` this state was there from the past, but now it depends on this feature flag at the team level, rather than the conversation protocol. 

Also renamed the function, since we are both setting this as the initial value, just that, the rest of the mutation depends on the user option selected for the conversation (if possible)

ohassine (approved):

PR #4189, opened 2025-08-14
fix: crash when downloading file (WPB-19465)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19465" title="WPB-19465" target="_blank">BugWPB-19465</a>  [Android] Crash when downloading excel file
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19465

# What's new in this PR?

### Issues
Crash when downloading file.

### Causes
Caused by uncaught I/O exception when deleting file.


features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
```diff
@@ -221,6 +219,12 @@ class CellViewModel @Inject constructor(
         }
     }
 
+    private fun deleteIfExists(filePath: Path) = runCatching {
+        if (kaliumFileSystem.exists(filePath)) {
+            kaliumFileSystem.delete(filePath)
+        }
+    }
```
ohassine, line 226: I'd say catching the exception only prevents the app from crashing, it doesn't resolve the issue.

I believe the crash started after we switched to downloading files into external storage, which is subject to scoped storage restrictions.
I've raised [PR #4172](https://github.com/wireapp/wire-android/pull/4172) to revert to using internal storage as Olga suggested. But, if we decide to continue using external storage, I think we need to use MediaStore API or Storage Access Framework (SAF) for the deletion.

Or we could download the file with a renamed name, and skip deletion ?

sbakhtiarov: I think we could keep this one to avoid crash, but we definitely need a proper solution for files management.

ohassine (approved):

saleniuk (approved):

PR #4223, opened 2025-08-27
feat: check if at least one cell available (WPB-19767)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19767" title="WPB-19767" target="_blank">TaskWPB-19767</a>  [Android] Do not request files if no cell available
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19767

# What's new in this PR?

### Issues
"All files" screen is available for the user if the team has Wire Cell feature enabled. But Cells API returns Not Authenticated error when there are no actual cells created.
So we need to check if there is at least one conversation with cell enabled before requesting all files list.

This is applicable only for All Files list since it is the only place accessible without opening a conversation with cells feature enabled.


features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
```diff
@@ -101,11 +109,29 @@ class CellViewModel @Inject constructor(
...
+    private val checkIfCellAvailable = flow {
+        if (isAllFiles()) {
+            isCellAvailable()
+                .onSuccess { emit(it) }
+                .onFailure { emit(false) }
+        } else {
+            emit(true)
+        }
+    }
+
```
ohassine: I don't see a need to make `checkIfCellAvailable` flow. It only emits a single boolean value, it does not represent a stream of data.

```
private suspend fun checkIfCellAvailable(): Boolean {
    return if (isAllFiles()) {
        isCellAvailable().getOrElse { false }
    } else {
        true
    }
}
```

features/cells/src/main/java/com/wire/android/feature/cells/ui/CellViewModel.kt
```diff
@@ -101,11 +109,29 @@ class CellViewModel @Inject constructor(
...
     }
 
-    internal val nodesFlow = combine(
+    internal val nodesFlow = checkIfCellAvailable.flatMapMerge { cellAvailable ->
+        if (cellAvailable) {
+            buildNodesFlow()
+        } else {
+            flowOf(emptyData)
+        }
+    }
```
ohassine: and we can use flow build here

```
internal val nodesFlow: Flow<DataType> = flow {
    val cellAvailable = checkIfCellAvailable()
    emitAll(if (cellAvailable) buildNodesFlow() else flowOf(emptyData))
}
```

ohassine (approved):

Garzas (approved):

PR #4237, opened 2025-09-03
chore: rename file name if it already exists upon download (WPB-19883)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19883" title="WPB-19883" target="_blank">BugWPB-19883</a>  [Android] File download error
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The issue in WPB-19883 happens because we are trying to download a file that already exists in local store

### Solution

After checking with design, we need to ensure that the file is downloaded every time the user clicks 'Download'. If a file with the same name already exists in local storage, it should be renamed automatically to avoid conflicts.

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


yamilmedina (commented):
Looks good so far! The only missing thing is to add some tests in the view model for this case so we catch this in the future.

ohassine:
> Looks good so far! The only missing thing is to add some tests in the view model for this case so we catch this in the future.

getUniqueFile() is a private function, so we can’t test it directly. Testing it indirectly would require an instrumented test for download(), since it relies on Android framework calls — and we don’t currently have such a test in place.

yamilmedina:
> > Looks good so far! The only missing thing is to add some tests in the view model for this case so we catch this in the future.
> 
> getUniqueFile() is a private function, so we can’t test it directly. Testing it indirectly would require an instrumented test for download(), since it relies on Android framework calls — and we don’t currently have such a test in place.

I would try to test it anyway, this index +1 or -1 manipulation brought some crashes in the past in the mentions feature for example ;/ 

The function might be private, but what if, we:
- Move `getUniqueFile` to `FileHelper` as public and test it there. If we need to run instrumented, you can use Robolectric (See `FileHelperTest` from the main app)

I also notice we have duplication in this class from cells `FileHelper`, for collective or for later indeed not this PR, the logic can be merged together in a `core/commons` or `core/io` for instance, tested there and reused everywhere.

sbakhtiarov (approved):

yamilmedina (approved):
Thaanks! for the coverage ! amazing work 🚀

PR #4257, opened 2025-09-10
fix: profile screen loading (WPB-20140)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20140" title="WPB-20140" target="_blank">BugWPB-20140</a>  [Android] Can only open Conversation details every second time
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20140

# What's new in this PR?

### Issues
Sometimes profile screen stays in loading screen after opening.

### Causes (Optional)
Incorrect usage of suspend function in data class copy function.

### Solutions
Remove suspend function from copy function.


saleniuk (approved):

yamilmedina (approved):
Wow! we had this bug looong time ago, good catch

PR #4260, opened 2025-09-11
fix: bottom sheet opened on top of keyboard (WPB-19885)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19885" title="WPB-19885" target="_blank">BugWPB-19885</a>  [Android] Confusing animation for file context menu when keyboard is open
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19885

# What's new in this PR?

### Issues
Bottom sheet opened on top of opened keyboard jumping down as the keyboard close.

### Solutions
Delay bottom sheet open animation if keyboard is open.

[Screen_recording_20250911_110642.webm](https://github.com/user-attachments/assets/9583eb79-08af-40c6-adea-6d2a929070bb)


MohamadJaara:
@saleniuk can you please take a look at this one i remember some time ago we had issues with keyboards and bottom sheets (something about preventing keyboards from being open in some edge cases)

features/cells/src/main/java/com/wire/android/feature/cells/ui/dialog/NodeActionsBottomSheet.kt
```diff
@@ -42,19 +47,23 @@ import com.wire.android.ui.common.bottomsheet.WireModalSheetLayout
...
 @Composable
 internal fun NodeActionsBottomSheet(
     menuOptions: MenuOptions,
     onAction: (NodeBottomSheetAction) -> Unit,
     onDismiss: () -> Unit,
-    sheetState: WireModalSheetState<Unit> = rememberWireModalSheetState<Unit>(WireSheetValue.Expanded(Unit))
+    sheetState: WireModalSheetState<Unit> = rememberWireModalSheetState<Unit>(WireSheetValue.Hidden)
 ) {
 
+    sheetState.showWithImeHidden()
```
saleniuk: We do have exact logic already implemented inside `WireModalSheetState<T>.show()` function, so I think simply calling `sheetState.show(hideKeyboard = true)` here should have the same effect and fix it as well. 😄 

We could also improve the whole thing by making the `sheetState` to have `MenuOptions` as a value, so like:
`WireModalSheetState<MenuOptions>`.
This way we wouldn't need to pass `MenuOptions` as a separate parameter and we could call `sheetState.show(value = menuOptions, hideKeyboard = true)` directly in `CallScreenContent` where `menuState` is being collected, so we could also remove 
`var menu by remember { mutableStateOf<MenuOptions?>(null) }` from there as well. 😄

saleniuk (approved):

vitorhugods (approved):

PR #4263, opened 2025-09-12
fix: open bottom sheet properly with disabled animations [WPB-19522]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19522" title="WPB-19522" target="_blank">BugWPB-19522</a>  [Android]Bottom sheet freezes when Animator duration scale is set to Off
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Bottom sheet sometimes freezes on the loading state and doesn't fully open when the data is fully fetched.

### Causes (Optional)

`AnchoredDraggable`, which is used by the Material3 `ModalBottomSheet`, kind of relies on animations - it has a `dragMutex` and it's using `tryMutate` so it doesn't suspend but skips some measure updates when they are close together, but when there's an animation ongoing then it will eventually happen in the next frame.

### Solutions

Re-trigger calculating and setting the offset by calling `SheetState.show()` when the height of the content changes - it  calls `animateTo`/`dragTo` again, so it eventually executes what has just been skipped because `dragMutex` was blocked at that moment.

#### How to Test

Disable animations and try to open conversation options or message options (these two have a "loading" state and their own ViewModels to fetch data).

### Attachments (Optional)

| Before | After |
| <video width="400" src="https://github.com/user-attachments/assets/65cd8c6a-cc1f-4c25-a9d6-76bfb6d0e08f"/> | <video width="400" src="https://github.com/user-attachments/assets/898ad4bb-1154-4775-b75e-63a30aebeb34"/> |

----

----

----
##### References

### DIFF
--- a/core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireModalSheetLayout.kt
+++ b/core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireModalSheetLayout.kt
@@ -30,9 +30,15 @@ import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.ModalBottomSheet
 import androidx.compose.material3.ModalBottomSheetProperties
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.unit.Dp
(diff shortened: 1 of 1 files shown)


core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireModalSheetLayout.kt
```diff
@@ -60,7 +66,18 @@ fun <T : Any> WireModalSheetLayout(
                 BackHandler(!shouldDismissOnBackPress) {
                     onBackPress()
                 }
-                sheetContent(expandedValue.value)
+                var contentHeight: Int by remember { mutableStateOf(0) }
```
sbakhtiarov, line 69: I think we should add a comment describing what kind of issue we are solving here. Otherwise it is not clear why do we need all these manipulations with content height.

sbakhtiarov (approved):

MohamadJaara (approved):

PR #4266, opened 2025-09-15
chore: files list skeleton loading (WPB-20175)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20175" title="WPB-20175" target="_blank">TaskWPB-20175</a>  [Android]Add skeleton loading for Recycle Bin transition
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20175

# What's new in this PR?

Showing skeleton loading for file lists

Screenshot_20250915_130159

### Refectoring:
Moved RowItem \ RowItemTemplate compose functions to common UI module.


features/cells/src/main/java/com/wire/android/feature/cells/ui/common/LoadingScreen.kt
```diff
@@ -17,27 +17,79 @@
...
+import androidx.compose.ui.draw.clip
+import com.wire.android.model.Clickable
 import com.wire.android.ui.common.colorsScheme
 import com.wire.android.ui.common.dimensions
-import com.wire.android.ui.common.progress.WireCircularProgressIndicator
+import com.wire.android.ui.common.rowitem.RowItemTemplate
+import com.wire.android.ui.common.shimmerPlaceholder
 
 @Composable
 fun LoadingScreen(
```
saleniuk, line 42: Maybe we could move this `LoadingScreen` to the common module as well - it uses `RowItemTemplate` inside anyway, not any specific item composable, so maybe we could be able to reuse this `LoadingScreen` in some other places where `RowItemTemplate` is used 🤔

saleniuk (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/common/LoadingScreen.kt
sbakhtiarov: I was thinking about it but thought that design may want some custom view for showing files loading placeholders, not the same as for conversations.

ohassine (approved):

PR #4271, opened 2025-09-16
fix: message bubble colors improvements [WPB-20065] [WPB-19597] [WPB-20066] [WPB-19604] [WPB-19599]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20065" title="WPB-20065" target="_blank">BugWPB-20065</a>  [Android] If user sends multiple messages or images, the avatar is shown at first message
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- added accent color to quoted message author
- updated color for quoted message background for message bubble
- added opacity for message bubble bottom labels
- fixed issue to show self deleting message label when message is not the last in time window
- added strings for weeks, days, hours, minutes without "left" text for message bubble
- aligned self message reactions to right
- added long click action on reactions to show message details
- fixed to show avatar of author to last not first message in time window


app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/MessageAuthorRow.kt
```diff
@@ -105,10 +106,16 @@ fun MessageSmallLabel(
         MessageStyle.NORMAL -> colorsScheme().secondaryText
     }
 
+    val alpha = when (messageStyle) {
+        MessageStyle.BUBBLE_SELF -> 0.5F
+        MessageStyle.BUBBLE_OTHER -> 1F
+        MessageStyle.NORMAL -> 1F
+    }
```
saleniuk: This alpha value `0.5F` for `BUBBLE_SELF` is used twice already, is it always true that for `BUBBLE_SELF` it should be 0.5f and for others 1.0f? If so then we could move it to the `MessageStyle` class or make a common extension property out of it for `MessageStyle` to keep that in a single place instead of hardcoding this value in other places 🤔

saleniuk (approved):
Just one small suggestion

sbakhtiarov (approved):

PR #4274, opened 2025-09-17
feat: enterprise managed restrictions support (WPB-20290)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20290" title="WPB-20290" target="_blank">StoryWPB-20290</a>  [Android] Base setup for RestrictionManager & Managed Configurations
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Provide building blocks for managed restrictions for EMM support.

### Solutions

Implement according to the specs. https://developer.android.com/work/managed-configurations#overview the possibility to:

- Added a Makefile to help executing tasks for development, building the apk, and installing it under the work profile, so can be managed by EMM (TestDCP for dev)
- Define a test schema for testing under `app_restrictions.xml` (https://developer.android.com/work/managed-configurations#define-configuration)
- Read managed restrictions (https://developer.android.com/work/managed-configurations#read-configurations) by implementing `ManagedConfigurationsRepository`
- Listen dynamically to changes, by using the runtime Broadcast Receiver (https://developer.android.com/work/managed-configurations#listen-configuration)

#### Test Coverage (Optional)

#### How to Test

To test, use TestDCP tool https://github.com/googlesamples/android-testdpc and change the value of the managed configuration, reading the value for now.

### Notes (Optional)

It is not possible to test the broadcast receiver, as we need a remote configuration tool to achieve it. 
Why? The tool is an app, and the receiver only works in runtime, while the app is in active state.

----

----

----
##### References


app/src/main/kotlin/com/wire/android/emm/ManagedConfigurationsReceiver.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.content.Context
+import android.content.Intent
+import com.wire.android.appLogger
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ManagedConfigurationsReceiver @Inject constructor(
+    @ApplicationContext val context: Context,
```
ohassine: This context is not used in this broadcast receiver

app/src/main/kotlin/com/wire/android/WireApplication.kt
```diff
@@ -166,19 +172,37 @@ class WireApplication : BaseApp() {
             override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {}
             override fun onActivityStarted(activity: Activity) {
                 globalAnalyticsManager.onStart(activity)
+                registerRuntimeReceivers()
```
ohassine: The `onActivityStarted()` callback will be triggered every time a new activity is created — this includes when navigating between activities. 
This means it will be called every time we receive a call, start a call or applock is requested, as we have the following activities: `AppLockActivity`, `OngoingCallActivity`, `StartingCallActivity`.

We should move registering the BroadcastReceiver to `Application.onCreate()` to avoid registering multiple times.

ohassine (changes requested):

saleniuk (approved):

app/src/main/kotlin/com/wire/android/emm/ManagedConfigurationsReceiver.kt
yamilmedina: Good catch ! remainder code I was trying for the next task.

app/src/main/kotlin/com/wire/android/WireApplication.kt
yamilmedina: Interesting, true it is registering multiple times, but also unregistering 🤔 . since onStop is called. But anyway made me think we can improve it to avoid this dance.

I will move it to the `WireActivity` instead and use onStart/onStop, why?  While `onCreate` would work, it is not the best practice for managed configurations.

https://developer.android.com/work/managed-configurations#listen-configuration

> Note: Ordinarily, your app does not need to be notified about configuration changes when it is paused. Instead, you should unregister your broadcast receiver when the app is paused. When the app resumes, you first check for the current managed configurations (as discussed in [Check Managed Configurations](https://developer.android.com/work/managed-configurations#check-configuration)), then register your broadcast receiver to make sure you're notified about configuration changes that happen while the app is active.

Also from the docs/guideline sabout dynamic broacast receivers.
https://developer.android.com/develop/background-work/background-tasks/broadcasts#register_receivers_in_the_smallest_scope

app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/DynamicReceiversManager.kt
```diff
@@ -0,0 +1,51 @@
...
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Manages dynamic registration and unregistration of broadcast receivers.
+ * This are receivers that are active while the app is in foreground only.
+ */
+@Singleton
+class DynamicReceiversManager @Inject constructor(
```
yamilmedina, line 34: Added this manager/helper class to reduce the code we are adding in `WireActivity` it is already ~700 lines and can be reused later for other broadcast receivers we might use.

ohassine (approved):

PR #4275, opened 2025-09-17
fix: message swipe accent color [WPB-19601]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19601" title="WPB-19601" target="_blank">TaskWPB-19601</a>  [Android] Chat bubbles: Message swipe gestures
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- added selfUserAccent view model to provide accent color for swipe gestures


app/src/main/kotlin/com/wire/android/ui/userprofile/self/SelfUserAccentViewModel.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.shareIn
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class SelfUserAccentViewModel @Inject constructor(
```
saleniuk, line 39: This one feels like it could be used as a scoped viewmodel, maybe we could extract an interface annotated with `@ViewModelScopedPreview` for it so that it could be used in previews 😄

saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/userprofile/self/SelfUserAccentViewModel.kt
sbakhtiarov: I also wonder if we really need a new ViewModel for observing user accent color. Maybe just use ConversationMessagesViewModel for that?

sbakhtiarov (approved):

PR #4280, opened 2025-09-18
feat: update tags when opening filter screen (WPB-20375)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20375" title="WPB-20375" target="_blank">BugWPB-20375</a>  [Android] List of tags not automatically updated
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20375

# What's new in this PR?

### Issues
New tags created by other users are not visible without application restart.

### Causes (Optional)
Tags are fetched when the files list screen is opened for the first time. No additional updates done.

### Solutions
Refresh list of available tags when user opens the tags filter. Add 30 seconds throttling to avoid too frequent requests.

### DIFF
--- a/features/cells/src/main/java/com/wire/android/feature/cells/ui/filter/FilterBottomSheet.kt
+++ b/features/cells/src/main/java/com/wire/android/feature/cells/ui/filter/FilterBottomSheet.kt
@@ -48,6 +48,7 @@ import com.wire.android.feature.cells.R
 import com.wire.android.ui.common.bottomsheet.WireModalSheetLayout
 import com.wire.android.ui.common.bottomsheet.WireModalSheetState
 import com.wire.android.ui.common.bottomsheet.WireSheetValue
+import com.wire.android.ui.common.bottomsheet.onShow
 import com.wire.android.ui.common.bottomsheet.rememberWireModalSheetState
 import com.wire.android.ui.common.button.WireButtonState
 import com.wire.android.ui.common.button.WirePrimaryButton
@@ -66,6 +67,7 @@ fun FilterBottomSheet(
     onApply: (Set<String>) -> Unit,
     onClearAll: () -> Unit,
     onDismiss: () -> Unit,
+    onShow: () -> Unit = {},
     sheetState: WireModalSheetState<Unit> = rememberWireModalSheetState<Unit>(WireSheetValue.Expanded(Unit))
 ) {
     WireModalSheetLayout(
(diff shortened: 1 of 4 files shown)


features/cells/src/main/java/com/wire/android/feature/cells/ui/filter/FilterBottomSheet.kt
```diff
@@ -66,6 +67,7 @@ fun FilterBottomSheet(
     onApply: (Set<String>) -> Unit,
     onClearAll: () -> Unit,
     onDismiss: () -> Unit,
+    onShow: () -> Unit = {},
```
saleniuk: `onShow` suggests (at least to me) that it triggers each time the bottom sheet is opened 😄 Maybe something like `onRefreshContent` would be better?

sbakhtiarov: I see the point, it does trigger when bottom sheet is shown. Just not every time it is shown.
Maybe I should remove the "throttling" delay from the base "onShow" function, make callback a suspend function and leave it to the caller to apply extra delays?
This way we can keep the onShow name.

saleniuk: That's actually quite nice, but I'm afraid it will stack up when the user opens this bottom sheet multiple times in a short time, right? For instance, if the user opens it three times within a few seconds, then it will execute after 30s, and then again after 60s and 90s. 
You could add `.buffer(capacity = 1, onBufferOverflow = BufferOverflow.DROP_LATEST)` to the `onShow` flow to limit emitted items that are waiting to be collected, but it would again make the flow to not emit exactly every time when it's shown. 😅  Another idea would be to create another flow in `CellViewModel` that emits items in `loadTags` and throttles/debounces `getAllTagsUseCase` executions. 😄

saleniuk: As we discussed on 1:1, it would stack up for a regular flow, but when using `snapshotFlow` it does not - it's not emitting new value until old one is not finished collecting, probably because "State change events are conflated" (it's from the `snapshotFlow` docs), so current solution should already work fine.

saleniuk (approved):

MohamadJaara (approved):

PR #4285, opened 2025-09-19
feat: disable profile QR code feature flag (WPB-20230)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20230" title="WPB-20230" target="_blank">TaskWPB-20230</a>  [Android] Respond to QR code backend feature flag and turn QR code feature on / off according to flag setting
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20230

# What's new in this PR?

New feature flag to hide QR code on user profile page.
- Hide QR code on profile page
- Disable opening profile page from deep link

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/userprofile/common/UserProfileInfo.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/userprofile/common/UserProfileInfo.kt
@@ -79,7 +79,6 @@ import com.wire.android.ui.theme.WireTheme
 import com.wire.android.ui.theme.wireColorScheme
 import com.wire.android.ui.theme.wireDimensions
 import com.wire.android.ui.theme.wireTypography
-import com.wire.android.util.debug.FeatureVisibilityFlags.QRCodeEnabled
 import com.wire.android.util.debug.LocalFeatureVisibilityFlags
 import com.wire.android.util.ifNotEmpty
 import com.wire.android.util.ui.PreviewMultipleThemes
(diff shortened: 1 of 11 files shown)


MohamadJaara:
this is only to show the QR code, i would assume the app should not react to the deeplink in this case, otherwise if the QR is obtained another way, the app will still react to it

sbakhtiarov:
> this is only to show the QR code, i would assume the app should not react to the deeplink in this case, otherwise if the QR is obtained another way, the app will still react to it

I will check it

saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/userprofile/common/UserProfileInfo.kt
```diff
@@ -258,7 +258,7 @@ fun UserProfileInfo(
                         bottom.linkTo(displayName.bottom)
                     }
             ) {
-                if (QRCodeEnabled && isLoading.not()) {
+                if (QRCodeEnabled && showQrCode && isLoading.not()) {
```
yamilmedina:
```suggestion
                if (showQrCode && isLoading.not()) {
```

I think is a good opportunity to remove this FeatureFlag from `FeatureVisibilityFlags`, I remember we were using this while developing but not anymore.

yamilmedina (commented):
LGMT! 🚀 
I think we can remove a leftover feature flag used while developing 1 year ago, since this is now controlled by BE :D

yamilmedina (approved):

PR #4303, opened 2025-09-24
chore: update nav icons and animations (WPB-20111)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20111" title="WPB-20111" target="_blank">TaskWPB-20111</a>  [Android] Improve skeleton replacement animation in Conversation Detail
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20111
https://wearezeta.atlassian.net/browse/WPB-20174

# What's new in this PR?

- Update group options placeholder animation
- Change group options nav icon to back arrow
- Change cell files / recycle bin nav icons to back arrow

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsScreen.kt
@@ -25,8 +25,6 @@ import androidx.compose.animation.Crossfade
 import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
-import androidx.compose.animation.slideInVertically
-import androidx.compose.animation.slideOutVertically
 import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.LocalOverscrollConfiguration
@@ -384,11 +382,7 @@ private fun GroupConversationDetailsContent(
             AnimatedContent(
                 targetState = isScreenLoading.collectAsState().value,
                 transitionSpec = {
-                    val enter = if (this.targetState) {
-                        slideInVertically(initialOffsetY = { it })
-                    } else {
-                        fadeIn(tween(durationMillis = 500, delayMillis = 100)) + slideInVertically(initialOffsetY = { it / 2 })
-                    }
+                    val enter = fadeIn(tween(durationMillis = 500, delayMillis = 100))
                     val exit = fadeOut()
                     enter.togetherWith(exit)
                 },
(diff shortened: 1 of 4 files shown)


ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsScreen.kt
```diff
@@ -359,7 +357,7 @@ private fun GroupConversationDetailsContent(
                     )
                     VerificationInfo(groupConversationOptionsState)
                 },
-                navigationIconType = NavigationIconType.Close(R.string.content_description_conversation_details_close_btn),
+                navigationIconType = NavigationIconType.Back(R.string.content_description_conversation_details_close_btn),
```
saleniuk: I believe that the conversation details screen was always supposed to have an X button instead of back arrow, just like the other user profile screen, and I don't see anything changing recently. 😄 
https://www.figma.com/design/UgKehvwIbiIiQ2CCEdsVuf/Android?node-id=1396-126588
All screens that open from the conversation details have arrow, it's like conversation details becomes a new "main" view and options like files or changing name are related to that view so user goes back. It's also related to the transition animation - screens with X slide from bottom and screens with back arrow slide from right to left.

sbakhtiarov: That was the request from design. I checked with Olga and she agreed to keep the conversation screen icon.
This is very strange. I do not remember the concept of "host screen" in Android and for user it looks confusing. During one flow navigation icon is changing from X to <- but the action is same.

saleniuk: But now transition animation does not match designs and we introduced another confusion - for 1:1s we have X button but for groups we have back arrow. 🫠

sbakhtiarov: I have returned X button for group details
Which animation? The window still slides in from bottom, but placeholder \ content animation is crossfade which is what design wanted. Also looks better than double-slide animation (window and the content after a small delay).

saleniuk: I agree that the placeholder->content animation inside the screen looks better now 👍 
I meant the transition animation when opening conversation details screen - for screens with X button, it should slide from bottom, for screens with back arrow it should slide from right to left, so the button type is associated with specific transition animation 😄 
https://www.figma.com/design/JGiz61Nl8WCobwPkHoF9yZ/Documentation?node-id=824-4326

saleniuk (approved):

PR #4316, opened 2025-09-26
feat(emm): get default Server URLs configuration via EMM (WPB-20305)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20305" title="WPB-20305" target="_blank">StoryWPB-20305</a>  [Android] Default Server URLs configuration via MDM
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We need to load custom server configuration from a managed device if they are available, if this is the case, then this configurations should be preferred instead of the compile time values from `default.json.`

### Solutions

- Follow the guidelines to load the fresher configs on runtime https://developer.android.com/work/managed-configurations#check-configuration The guidelines suggest `onResume`, but I believe is better to do it on demand by using injection and **non scoped** component from Dagger delaying/offloading it from doing it each time the app resumes.
- Add the parser from a Json value to a domain object.
- Test coverage of the Restrictions Manager, using Robolectric.
- Improve the Makefile to detect the workprofile instead of hardcoded id.

#### Test Coverage (Optional)

### Attachments

https://github.com/user-attachments/assets/b5ef2265-9a40-43fa-a7f7-027af37d1b04

----

----

----
##### References


app/src/main/kotlin/com/wire/android/config/ServerConfigProvider.kt
```diff
@@ -18,18 +18,58 @@
...
+            )
+        }
+    }
+}
+
+// Keep the function for backward compatibility and Hilt module usage
+fun getDefaultServerConfig(managedServerConfig: ManagedServerConfig? = null): ServerConfig.Links {
+    val provider = ServerConfigProvider()
+    return provider.getDefaultServerConfig(managedServerConfig)
+}
```
MohamadJaara, line 64: this makes me question how the domain registration should work?
should it use the baked in wire domain or the one from MDM ?

yamilmedina: The path won't go through wire domain, since this is the implementation in place for enterprise login, it will go to whatever it has on the deeplink configs (managed too), therefore if no domain registration endpoint is available in the BE, it will behave as any other on premises case.

app/src/main/kotlin/com/wire/android/di/ServerConfigModule.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun provideServerConfigProvider(): ServerConfigProvider = ServerConfigProvider()
+
+    @Provides
+    fun provideCurrentServerConfig(
+        serverConfigProvider: ServerConfigProvider,
+        managedConfigurationsRepository: ManagedConfigurationsRepository
+    ): ServerConfig.Links {
+        val managedServerConfig = runBlocking(Dispatchers.IO) {
+            managedConfigurationsRepository.getServerConfig()
+        }
```
saleniuk: This `runBlocking` will be executed each time the app needs this config, could we mark this as `@Singleton` so that it could reuse the same instance instead of providing a new one and executing it each time? 🤔

app/src/main/kotlin/com/wire/android/config/ServerConfigProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+// Keep backward compatibility
+val DefaultServerConfig get() = getDefaultServerConfig()
+
+// Deprecated: Use Hilt-injected ServerConfig.Links instead
+@Deprecated(
+    "Use Hilt-injected ServerConfig.Links instead of calling orDefault()",
+    ReplaceWith("this ?: getDefaultServerConfig(managedServerConfig)")
+)
+fun ServerConfig.Links?.orDefault(managedServerConfig: ManagedServerConfig? = null) =
+    this ?: getDefaultServerConfig(managedServerConfig)
```
saleniuk: I believe that you replaced all `.orDefault()` usages in the app into a hilt-injected `ServerConfig.Links`, is there any specific reason why we still need to keep this method and compatibility?

yamilmedina: True that ! I was trying to keep it as a just in case, I will remove it.

app/src/main/kotlin/com/wire/android/di/ServerConfigModule.kt
yamilmedina: This is tricky, for managed configurations we need to (according to the [guidelines](https://developer.android.com/work/managed-configurations#check-configuration)) to get the latest/fresher values each time the app goes `onResume` I figure it out that it is less expensive to do this **only** when we actually need it. LoginScreen, CreateAccount, WelcomeScreen, so this in practice is less expensive than doing it `onResume` imho, given these actions are executed a handful of times.

Happy to discuss another approach, but the singleton approach would not get the latest changes from the managed configurations unfortunately

saleniuk: Would it be possible to at least throttle it? Like create a provider which would throttle/cache it for 5 minutes or even more if that would be possible, or at least have a `suspend fun getManagedServerConfig()` and move the "suspend" part into view models so that we don't have this `runBlocking` part inside the "provide" method so when the stuff is being initialized - our launch time is already not that great. 🙈 But maybe I'm just overthinking and this getting restrictions is not that harmful, so It's up to you. 😄

yamilmedina: Thanks for the review, I gave it a go, and now I have changed the approach to have:
- Load the config once when the app starts with a singleton and having a variable to hold the current resolved value
- Calling the refresh of these configs, on appStart and with the broadcast receiver, this now aligned with the guidelines.

This should cover all the grounds plus not be a blocking task anymore.

Let me know what do you think :D

saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -152,6 +153,9 @@ class WireActivity : AppCompatActivity() {
     @Inject
     lateinit var dynamicReceiversManager: DynamicReceiversManager
 
+    @Inject
+    lateinit var managedConfigurationsRepository: ManagedConfigurationsRepository
+
```
ohassine, line 158: We should avoid using repositories directly in Activities, as it breaks separation of concerns.
Instead, we can create a use case `ServerConfigRefresherUseCase` that encapsulates the logic.
And this use case can be injected into the ViewModel and called from the UI layer.

ohassine (dismissed):

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
yamilmedina: In this case is just naming following the specs from the ticket, in reality this is more a Facade over `RestrictionsManager` than anything else, I will rename it to `ManagedConfigurationsManager` as it is more accurate for this case

I will address this rename in the following PR, to avoid conflicts with the ongoing changes there

ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
yamilmedina: Thanks for the review buddy @ohassine 🙌  I addressed the changes here as agreed. 
https://github.com/wireapp/wire-android/pull/4322/commits/8e9f2e195ec8e680b26e733305ade9a9fab8bce4

PR #4322, opened 2025-10-01
feat(emm): get default SSO code configuration via EMM (WPB-20308)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20308" title="WPB-20308" target="_blank">StoryWPB-20308</a>  [Android] Implement default SSO code via MDM
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We need to load default SSO code configuration from a managed device if they are available, if this is the case, then this configurations should be used in the login screen.

### Solutions

- Follow the guidelines to load the fresher configs on runtime https://developer.android.com/work/managed-configurations#check-configuration The guidelines suggest `onResume`, but I believe is better to do it on demand by using injection and **non scoped** component from Dagger delaying/offloading it from doing it each time the app resumes.
- Add the parser from a Json value to a domain object.
- Test coverage, using Robolectric.
- Fix comment from prev PR https://github.com/wireapp/wire-android/pull/4316#discussion_r2394836412

#### Test Coverage (Optional)

### Attachments

https://github.com/user-attachments/assets/106e9069-a098-4565-bd17-32474f07432a

----

----

----
##### References


saleniuk (approved):

app/src/main/kotlin/com/wire/android/emm/ManagedSSOCodeConfig.kt
```diff
@@ -0,0 +1,37 @@
...
+    @SerialName("sso_code")
+    val ssoCode: String
+) {
+    @Transient
+    val isValid: Boolean = try {
+        UUID.fromString(ssoCode)
+        true
+    } catch (exception: IllegalArgumentException) {
+        false
+    }
```
MohamadJaara, line 36: question: Since it is only a uuid string, do we need to wrap it in JSON?

yamilmedina: 🤔  I wanted to make it standard with the same structure we have on BE for this login case, but I guess we can simplify it by making it just a regular string key/val

MohamadJaara: ok no need to change it now just wanted to know if there is a reason for this

MohamadJaara (approved):

PR #4390, opened 2025-10-30
feat: reply to multipart messages, part I (WPB-16897)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16897" title="WPB-16897" target="_blank">TaskWPB-16897</a>  [Android] Support reply for multipart message
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-16897

# What's new in this PR?

Part 1 of the changes to support replying to multipart messages: replying and rendering reply preview.
Main changes:
- Enable reply feature for multipart message type
- New UI for multipart message reply preview (see Figma)
- Refactoring draft view model to observe draft and quoted message. This is required to correctly update the state of the multipart asset if it becomes unavailable.
- Moved `saveDraft` method to MessageDraftViewModel

Figma: https://www.figma.com/design/grnVU2vAihHXwYgHryu2xE/Cells---Files?node-id=9487-193420&p=f&t=hnZxJrjkkZKEAwoD-0


app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/QuotedMultipartMessage.kt
```diff
@@ -0,0 +1,226 @@
...
+                contentDescription = null,
+            )
+
+            if (AttachmentFileType.fromMimeType(attachment.mimeType) == VIDEO) {
+                Image(
+                    modifier = Modifier
+                        .size(dimensions().spacing24x)
+                        .align(Alignment.Center),
+                    painter = painterResource(id = R.drawable.ic_play_circle_filled),
+                    contentDescription = null,
```
saleniuk, line 146: I think it's better to always just fill in the `contentDescription` for the accessibility reasons

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/QuotedMultipartMessage.kt
```diff
@@ -0,0 +1,226 @@
...
+            text = if (file.assetAvailable) name else stringResource(R.string.asset_message_failed_download_text),
+            color = colorsScheme().secondaryText
+        )
+    }
+}
+
+@Composable
+private fun UIMultipartQuotedContent.imageModel(): ImageRequest {
+
+    val builder = ImageRequest.Builder(LocalContext.current)
```
saleniuk, line 204: Is `UIMultipartQuotedContent` used only for cell-related assets and preview urls for cells are always public? If so then that's why we can just fetch the preview directly from the url, in contrast to regular asset messages where we need to fetch the data using dedicated request and decode it. 
Either way I would love to have the asset fetching logic unified and in one place, as we used to only use `WireSessionImageLoader` so we had everything we needed in there: fetching, caching, handling errors and retrying, even handling gifs and compose previews, but now for some newer cases we started creating separate `ImageRequest`s outside of it.

sbakhtiarov: Yes, the preview URLs are always public.
- Cell files are not encrypted, so we do not need to decrypt.
- My assumption is that if we have a local file it is always better to get preview from local file rather than making a network call.
- Coil will cache the image for us but preview URLs also expire and are updated.

I think unified loader would be great but it is a bigger task to mix it with anything else.

saleniuk (approved):

Garzas (approved):

PR #4406, opened 2025-11-11
fix: text colors for cell attachment file names and reply preview (WPB-21712)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21712" title="WPB-21712" target="_blank">BugWPB-21712</a>  [Android] Incorrect text colors for text in file name in message view and reply preview
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-21712

# What's new in this PR?

### Issues
Fixing invalid text colors in multipart replies and file names.

Screenshot_20251111_164209
Screenshot_20251111_164146
Screenshot_20251111_164032

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/QuotedMessage.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/QuotedMessage.kt
@@ -43,6 +43,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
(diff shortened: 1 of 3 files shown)


MohamadJaara (approved):

saleniuk (commented):
Looks like borders of the replied messages are also wrong for bubbles:

| Current | Designs |
|---------|---------|
|  |  |

Can we also remove these borders for bubbles as well while we're at it?

ohassine (approved):
I think we are not using the right color for border in light mode

saleniuk:
> Looks like borders of the replied messages are also wrong for bubbles:
> 
> Current	Designs
> 	
> Can we also remove these borders for bubbles as well while we're at it?

I believe this PR already fixes that: https://github.com/wireapp/wire-android/pull/4408
So it's all good, no need to do that here

saleniuk (approved):

PR #4410, opened 2025-11-12
feat: update and refactoring for move to folder feature (WPB-21155)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21155" title="WPB-21155" target="_blank">TaskWPB-21155</a>  [Android]Update “Move to Folder” empty and none-selected states (disable Move button, adjust CTA hierarchy)
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-21155

# What's new in this PR?

Updating logic and refactoring the Move To Folder feature screen.

- Disable "Move Here" button when no folder selected to avoid moving file to the same folder
- Show "Create Folder" button as primary button if no folder selected
- New text for empty screen if no (root) folders available for selection

Refactoring:
- Moving some logic from composable files to view model
- Move screen parameters to view state
- Extract composable function for content into another file to reduce file length

https://www.figma.com/design/grnVU2vAihHXwYgHryu2xE/Cells---Files?node-id=9021-421573&t=OtIBYIZDGP1jX71t-0


saleniuk (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/movetofolder/MoveToFolderViewModel.kt
```diff
@@ -43,56 +43,107 @@ class MoveToFolderViewModel @Inject constructor(
...
+    val folders: List<CellNodeUi.Folder> = emptyList(),
+    val breadcrumbs: List<String> = emptyList(),
+    val isAllowedToMoveToCurrentPath: Boolean = false,
+    val isInRootFolder: Boolean = false,
+)
+
 sealed interface MoveToFolderViewAction {
     data object Success : MoveToFolderViewAction
     data object Failure : MoveToFolderViewAction
+    data class NavigateToBreadcrumb(val steps: Int) : MoveToFolderViewAction
```
ohassine, line 141: The name `NavigateToBreadcrumb` might be a bit confusing, how about `NavigateThroughBreadcrumb` ?

ohassine (approved):
tbh, I’m not a fan of sending navigation actions through the ViewModel. I feel like we’re overcomplicating things, when it would be much easier and cleaner for reading to just call the navigate command directly from the UI instead of going through the ViewModel and sending the action back.

I also like to keep ViewModels focused on business logic rather than UI stuff like navigation. Otherwise, it can start to feel like we’re mixing responsibilities and adding extra boilerplate.

PR #4450, opened 2025-11-27
feat(cells): Fetch and display node versions (WPB-21661)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21661" title="WPB-21661" target="_blank">TaskWPB-21661</a>  [Android] Implement versioning
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Fetch and display node versions 

[Screen_recording_20251127_170718.webm](https://github.com/user-attachments/assets/50ca82b2-5dce-451e-aaa0-20d6608c6845)

### Causes (Optional)

_Briefly describe the causes behind the issues. This could be helpful to understand the adopted solutions behind some nasty bugs or complex issues._

### Solutions

_Briefly describe the solutions you have implemented for the issues explained above._

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


features/cells/src/main/java/com/wire/android/feature/cells/ui/CellListItem.kt
```diff
@@ -120,18 +125,22 @@ internal fun CellListItem(
                     }
                 }
             }
-
-            Box(
+            Icon(
```
saleniuk, line 128: Removing the outer `Box` with specified size makes this `Icon` very small and way harder to click as it now has the clickable area equal to the size of the image vector, so only 16x16 dp. 
Is there any specific reason why it was changed and not be wrapped in the `Box` anymore just like we have in all other places?

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+            visible = isFetchingContent,
+            enter = fadeIn(),
+            exit = fadeOut()
+        ) { LoadingScreen() }
+
+        AnimatedVisibility(
+            visible = !isFetchingContent,
+            enter = fadeIn(),
+            exit = fadeOut()
+        ) {
```
saleniuk, line 92: Instead of using two `AnimatedVisibility`, you can use just one `AnimatedContent` or `Crossfade` to animate between multiple states

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
```diff
@@ -0,0 +1,130 @@
...
+            .map { (date, items) ->
+                val dateLabel = when (date) {
+                    today -> "${resourceProvider.getString(R.string.today_label)}, " +
+                            "${date.format(DateTimeFormatter.ofPattern(DATE_PATTERN))}"
+
+                    yesterday -> "${resourceProvider.getString(R.string.yesterday_label)}, " +
+                            "${date.format(DateTimeFormatter.ofPattern(DATE_PATTERN))}"
+
+                    else -> date.format(DateTimeFormatter.ofPattern(DATE_PATTERN))
+                }
```
saleniuk: It would be better to pass `UIText` instead, just like in all other ViewModels that need to pass some string resources with arguments. `UIText` is already moved to the common module, so you can simply use it here without the need of introducing this `ResourceProvider` and getting resources in the ViewModel as it should be done in the UI part of code, not VM.
I would suggest to also move this string formatting into string resources, just like we have already for messages and meetings, in case some languages do that differently.
So creating string resources with arguments:
```
<string name="today_label">Today, %s</string>
<string name="yesterday_label">Yesterday, %s</string>
<string name="other_day_label">%s</string>
```
and passing it as `UIText`:
```
val dateLabel = UIText.StringResource(
    when (date) {
        today -> R.string.today_label
        yesterday -> R.string.yesterday_label
        else -> R.string.other_day_label
    },
    date.format(DateTimeFormatter.ofPattern(DATE_PATTERN))
)
```

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
```diff
@@ -0,0 +1,130 @@
...
+        )
+        var size = bytes.toDouble()
+        var index = 0
+
+        while (size >= 1024 && index < units.size - 1) {
+            size /= 1024
+            index++
+        }
+        return String.format("%.2f %s", size, units[index])
+    }
```
saleniuk: Can we move/create that in the common module, like `core/ui-common`, to be able to reuse that everywhere in the app? As we already use similar logic in a few places (like `DeviceUtil.formatSize`) so it would be nice to have that in a single place.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryNavArgs.kt
```diff
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see http://www.gnu.org/licenses/.
  */
-package com.wire.android.feature.cells.ui.versionhistory
+package com.wire.android.feature.cells.ui.versioning
 
 class VersionHistoryNavArgs(
     val uuid: String? = null,
```
sbakhtiarov, line 21: What would be the scenario for calling this screen with null node UUID? I think we should make this not nullable.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    companion object {
+        const val DATE_PATTERN = "d MMM yyyy"
+        const val TIME_PATTERN = "HH:mm"
```
sbakhtiarov, line 121: I assume that ideally we need to respect the 24h / 12h (am/pm) time formats. I was also struggling with this a little bit for the public link expiration.

sbakhtiarov (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryNavArgs.kt
ohassine: we will display an empty list

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
ohassine: Right, using "HH:mm" forces a 24-hour format.
Switching to use the system’s time-format preference

features/cells/src/main/java/com/wire/android/feature/cells/ui/CellListItem.kt
ohassine: The clickable area is 24x24 dp and changed it to have circular ripple effect instead of the boxed one we currently have

saleniuk (approved):

PR #4462, opened 2025-12-04
feat: Restore node version (WPB-21661) 

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21661" title="WPB-21661" target="_blank">TaskWPB-21661</a>  [Android] Implement versioning
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Follow up PR of https://github.com/wireapp/wire-android/pull/4450

[Screen_recording_20251204_235356.webm](https://github.com/user-attachments/assets/bf77947d-d0cc-4172-8623-d6064ab9b33b)

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


core/ui-common/src/main/kotlin/com/wire/android/ui/common/WireDialog.kt
```diff
@@ -103,7 +103,7 @@ fun WireDialog(
         contentPadding = contentPadding,
         title = title,
         titleLoading = titleLoading,
-        text = buildAnnotatedString {
+        text = if (text != null) buildAnnotatedString {
```
sbakhtiarov, line 106: text?.let { 
  ...
}

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionActionBottomSheet.kt
```diff
@@ -0,0 +1,120 @@
...
+                        onClicked = { onDownloadVersionClicked(state.versionId) },
+                    )
+                }
+            }
+        )
+    }
+}
+
+@Composable
+private fun RestoreVersionModalItem(
```
sbakhtiarov, line 70: it looks like Restore / Download items is the same function. I think we only need one just call with different title / onClick params.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
```diff
@@ -0,0 +1,221 @@
...
+                        )
+                    }
+                },
+                navigationIconType = NavigationIconType.Close(),
+                elevation = dimensions().spacing0x,
+            )
+        },
+    ) { innerPadding ->
+
+        AnimatedVisibility(
```
sbakhtiarov, line 141: I think Michal mentioned that in a different PR. For switching between different content views (e.g. Loading / Content / Error) we should use AnimatedContent function.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
```diff
@@ -0,0 +1,221 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        VersionActionBottomSheet(
+            sheetState = optionsBottomSheetState,
+            onDismiss = { optionsBottomSheetState.hide() },
+            onRestoreVersionClicked = showRestoreConfirmationDialog,
```
sbakhtiarov, line 203: Do we actually hide the bottom sheet here when user selects the action? I do not see the '.hide()' call. 
My advise would be to use the `fun hide(onComplete: suspend () -> Unit = {})` function and execute the action in the callback. This way we get a nice transition with the bottom sheet hide animation and then dialog appearing on the screen.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    init {
+        fetchNodeVersionsGroupedByDate()
+    }
+
+    fun fetchNodeVersionsGroupedByDate() =
+        viewModelScope.launch {
+            isFetchingContent.value = true
+            navArgs.uuid?.let {
+                getNodeVersionsUseCase(navArgs.uuid).onSuccess {
```
sbakhtiarov: Should we handle `onError` somehow? Seems like we just end up with the empty screen if there is no connection or server error.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
```diff
@@ -0,0 +1,199 @@
...
+                restoreVersionState = RestoreVersionState.Restoring
+            )
+
+            viewModelScope.launch {
+                val progressJob = simulateRestoreProgress()
+
+                restoreNodeVersionUseCase(navArgs.uuid ?: "", value.versionId)
+                    .onSuccess {
+                        delay(DELAY_500_MS) // delay since server takes some time to restore the version
+                        val fetchJob = fetchNodeVersionsGroupedByDate()
```
sbakhtiarov: I think it may be better to make this `fetchNodeVersionsGroupedByDate()` function suspend function. So you do not need to do these job start/join calls. And in the init{} block just call it in viewModelScope.launch{} block.

sbakhtiarov (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionActionBottomSheet.kt
ohassine: It will be confusing, I think It's cleaner to have them separated

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
ohassine: yes we do, in the top level function before calling `showRestoreConfirmationDialog()`

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
ohassine: yes, I will add a TODO for next PR

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryViewModel.kt
ohassine: yes it was like that, changed it so I can invoke it from UI without the need to lunch a new coroutine

sbakhtiarov: I think it is easier to read / understand:
```
  coroutineScope { fetchNodeVersions() }
```
then start() / join() for the fetchJob. Took me some time to figure out what is going on here especially with the progressJob handling in the same fun.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
```diff
@@ -52,65 +62,133 @@ fun VersionHistoryScreen(
...
-                            )
-                            WireDivider(
-                                modifier = Modifier.fillMaxWidth(),
-                                color = colorsScheme().outline
-                            )
+            if (isFetching) {
+                LoadingScreen()
+            } else {
+                PullToRefreshBox(
+                    isRefreshing = isFetchingContent,
```
saleniuk: I think you should use separate fields for initial loading state and "pull to refresh" `isRefreshing`, like `isFetchingContent` and `isRefreshingContent` or create an enum with three values `{ INITIAL_FETCHING, REFRESHING, LOADED }`.
Right now you use the same field for both `PullToRefreshBox` and `LoadingScreen`, so it will actually never use it here inside `PullToRefreshBox` because it will always  show `LoadingScreen` instead of `PullToRefreshBox` because it's the `targetState` of this `AnimatedContent`.

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
```diff
@@ -52,65 +62,133 @@ fun VersionHistoryScreen(
...
+                    onRefresh = onRefresh,
+                ) {
+                    LazyColumn(Modifier.padding(innerPadding)) {
+                        versionsGroupedByTime.forEach { group ->
+                            item {
```
saleniuk: Can you assign an id as a key to each item? That helps with maintaining proper scroll state and list modifications.
To make it more efficient knowing that all items are exactly the same, then you could also hardcode some `contentType` like `contentType = "versionItem" so that the LazyColumn knows that for compositions it can just reuse items freely. 😃

features/cells/src/main/java/com/wire/android/feature/cells/ui/versioning/VersionHistoryScreen.kt
ohassine: valid point, thanks

saleniuk (approved):

PR #4464, opened 2025-12-05
chore(l10n): update localization strings via Crowdin (WPB-9776)

AndroidBob:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9776" title="WPB-9776" target="_blank">TaskWPB-9776</a>  [GitHub Action] Update translation
  </td></table>
  <br />
 

This PR pulls in the latest localization translations from Crowdin.

### DIFF
--- a/core/ui-common/src/main/res/values-de/strings.xml
+++ b/core/ui-common/src/main/res/values-de/strings.xml
@@ -48,4 +48,9 @@
     <string name="label_ok">OK</string>
     <string name="label_cancel">Abbrechen</string>
     <string name="label_no_selected_date">Kein Datum ausgewählt</string>
+    <string name="size_unit_bytes">B</string>
+    <string name="size_unit_kilobytes">KB</string>
+    <string name="size_unit_megabytes">MB</string>
+    <string name="size_unit_gigabytes">GB</string>
+    <string name="size_unit_terabytes">TB</string>
 </resources>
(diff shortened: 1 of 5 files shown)


yamilmedina (approved):

core/ui-common/src/main/res/values-de/strings.xml
```diff
@@ -48,4 +48,9 @@
     <string name="label_ok">OK</string>
     <string name="label_cancel">Abbrechen</string>
     <string name="label_no_selected_date">Kein Datum ausgewählt</string>
+    <string name="size_unit_bytes">B</string>
+    <string name="size_unit_kilobytes">KB</string>
+    <string name="size_unit_megabytes">MB</string>
+    <string name="size_unit_gigabytes">GB</string>
+    <string name="size_unit_terabytes">TB</string>
```
saleniuk, line 55: I believe B is a standard international symbol for byte, and K- M- G- T- are just international unit prefixes so they should be the same in each language, right?
I think that we should keep them only in main `values/strings.xml` file and mark them as `translatable="false"`, WDYT?

yamilmedina: Makes sense to me ! Are you able to make the change 🙏 ?

saleniuk: I thought that since standard unit prefixes are used to describe file size then the “byte” symbol is also a world-wide standard just like all the SI base units, but apparently some languages, like french, have a different opinion on that, so I will just resolve this comment and approve this PR 😅 
image

saleniuk (approved):

PR #4471, opened 2025-12-09
chore: replace material icons

sbakhtiarov:
# What's new in this PR?

Material Icons library is not recommended for usage and is removed from the recent Material library version.
https://developer.android.com/jetpack/androidx/releases/compose-material3#compose_material3_version_14_2

- Removing material-icons-extended dependency
- Replacing material icons with the local vector icons from app resources


kalium
saleniuk, line 1: does it require changing the kalium version as well or was it unintentional?

saleniuk (approved):

ohassine (approved):
There are still 2 icons need change in `FileSelectionScreen`

PR #4474, opened 2025-12-09
refactor: build waveform after fetching audio asset and update waveform colors [WPB-22118]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22118" title="WPB-22118" target="_blank">BugWPB-22118</a>  [Android] Some audio files to not generate an audio wave when shared on android
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Waveforms generation can be executed each time the audio message is shown on the screen.
Waveform colors are outdated and don't match current bubbles designs.

### Causes (Optional)

Generation is executed from the ViewModel so it can be called multiple times. It can cause some problems and unnecessary use of resources when user sends an audio message - then first waveform build attempt is done while sending, and then another attempt can be called right after when the message is shown.
After implementing lat changes to bubbles designs, waveforms were omitted so there are colors that don't match designs and current state of bubbles.

### Solutions

Change the logic so that the waveform is generated only right after downloading the audio asset, so generating it is removed from `AudioMessageViewModel`, `ConversationAudioMessagePlayer` and `RecordAudioMessagePlayer` in favor of this new logic implemented in kalium. This way it's executed only once and never retries if it failed (no need - if it failed the first time then it will fail each next time for that same exact audio file). Generation is executed inside a newly created worker which is scheduled as an one-time job right when the successful download is finished, so that it can be executed in the background (it's not crucial for the app to have it instantly). 
In the `AudioMessageViewModel` only fetching of audio asset is initialized and then the waveform is being observed by collecting changes in the message data, which contains waveform, so it only reads current values.
Extracted `waveform` from `AudioState` as this state should only contain fields used for the audio player and waveform is only used for displaying it on the conversation screen so it's better to keep both separated.
Added two new color chooser extension functions to properly apply colors to the waveform and "current time" label and updated color for the dark secondary button border according to the designs.
Also, removed some dead code.

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/3764

#### Test Coverage (Optional)

#### How to Test

Send audio message from app that doesn't yet support sending waveform along with the audio message, so some old Android app like 4.17.0 or iOS. After opening that conversation and receiving that message on new Android app, it should fetch the audio file and then generate a waveform for it.

----

----

----
##### References


app/src/main/kotlin/com/wire/android/media/audiomessage/AudioMessageViewModel.kt
```diff
@@ -86,7 +92,23 @@ class AudioMessageViewModelImpl @Inject constructor(
...
+            observeMessageById(args.conversationId, args.messageId)
+                .map {
+                    (it as? ObserveMessageByIdUseCase.Result.Success)?.message?.content?.let {
+                        (it as? MessageContent.Asset)?.value?.metadata?.let {
+                            (it as? AssetMetadata.Audio)?.normalizedLoudness?.toWavesMask()
+                        }
+                    }
+                }
+                .distinctUntilChanged()
+                .collectLatest { state = state.copy(wavesMask = it) }
```
sbakhtiarov: Is this mask supposed to be updated more than once? Maybe we can wait for the first non-empty value and stop observing after getting it?

sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/media/audiomessage/AudioMessageViewModel.kt
saleniuk: Right, it should be generated only once so it shouldn't be updated more than once, nice suggestion, added to the code

MohamadJaara (approved):

PR #4476, opened 2025-12-11
fix: image normalized size crash coercein [WPB-22318]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22318" title="WPB-22318" target="_blank">BugWPB-22318</a>  [Android] Message bubble sometimes crashes while normalizing image dimensions
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
    - [ ] is free of optional paragraphs, and you have filled the relevant parts to the best of your ability
----
# What's new in this PR?

### Issues
- App crashed during image rendering due to invalid min/max bounds passed to `coerceIn` in `normalizedSize`.

### Causes (Optional)
- `minW` and `maxW` (or height equivalents) could become reversed, especially when using `ScreenFraction`, causing `coerceIn` to throw an exception.

### Solutions
- Added safe clamping to ensure valid min/max ordering
- Adjusted logic to prevent minimum values from exceeding computed max bounds.
- Added targeted tests covering this crash scenario to prevent regression.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/image/VisualMediaParams.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/image/VisualMediaParams.kt
@@ -26,7 +26,6 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.min
 import com.wire.android.ui.common.dimensions
 import kotlinx.serialization.Serializable
-import kotlin.let
 
 /**
  * Parameters describing visual media (image or video) used to calculate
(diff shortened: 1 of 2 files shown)


yamilmedina (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/model/messagetypes/image/VisualMediaParams.kt
```diff
@@ -69,15 +68,25 @@ data class VisualMediaParams(
             }
         }
 
+        // Guard against pathological max bounds (e.g. fraction == 0f)
+        if (effMaxW <= 0.dp || effMaxH <= 0.dp) {
+            return NormalizedSize(minW, minH, isPortrait = realMediaHeight > realMediaWidth)
+        }
```
ohassine: It would be better to avoid abbreviations, it's not clear what `eff` mean without digging into logic.

sbakhtiarov (approved):

PR #4518, opened 2026-01-12
chore: add crowdin entries automagically [WPB-8645]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8645" title="WPB-8645" target="_blank">EpicWPB-8645</a>  [Android] Infrastructure code and developer experience
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some Crowdin entries are missing from configuration.

### Causes

We need to remember to add it manually when we create new modules.

### Solutions

Add new entries automatically when building each module.

1. Create a new Gradle tasks that will check if `_moduleName_/src/main/res/values/string.xml` exists.
2. If it exists, it will check the content of `/crowdin.yml`.
3. If the entry is not in the Crowdin file: add it.

The task is added to all `package*Resources` tasks. So if the module is built at any point, it will trigger the task.

> [!NOTE]
> This is **NOT** 100% failproof, but it's super good enough:
> Someone could create a new module, and still raise a PR and merge it without building the app, or running the `addEntryToCrowdinFile` Gradle task.
However, anyone else building the app later would notice changes to crowdin configuration, and they could be committed in the future without issues.

### Testing

Ran locally, missing entries are added by the new task, not by me.

#### How to Test

0. Remove entries from `crowdin.yml`.
1. Run `./gradlew addEntryToCrowdinFile`; OR build the app / build one library module.
2. Check `crowdin.yml`.

----

----
##### References

### DIFF
--- a/build-logic/plugins/build.gradle.kts
+++ b/build-logic/plugins/build.gradle.kts
@@ -20,6 +20,7 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     `kotlin-dsl`
+    alias(libs.plugins.kotlin.serialization)
 }
 
 // Configure the build-logic plugins to target JDK 17
@@ -41,6 +42,7 @@ dependencies {
     testImplementation(libs.junit4)
     testImplementation(kotlin("test"))
     implementation(libs.ksp.symbol.processing.plugin)
+    implementation(libs.ktx.serialization)
 }
 
 gradlePlugin {
--- a/build-logic/plugins/src/main/kotlin/AndroidLibraryConventionPlugin.kt
+++ b/build-logic/plugins/src/main/kotlin/AndroidLibraryConventionPlugin.kt
@@ -19,6 +19,7 @@ import com.android.build.gradle.LibraryExtension
 import com.wire.android.gradle.configureAndroidKotlinTests
 import com.wire.android.gradle.configureCompose
 import com.wire.android.gradle.configureKotlinAndroid
+import com.wire.android.gradle.crowdin.AddEntryToCrowdinTask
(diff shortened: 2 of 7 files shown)


yamilmedina (commented):
Looks good to me; codewise, it's clear and nice usage of conventions plugin 🙌 

My only concern is, is it perhaps a solution for a problem we don't have?  since there is no guaranteed or 100% failproof and configuring a module is more or less tasks we don't do everyday.

vitorhugods:
It's guaranteed to catch at some point. When the app is built, for example.

There is currently no automation to detect misconfiguration / missing entries in `crowdin.yml`.

Being pragmatic, I think the alternatives are:
1. Keep the way it is, and rely solely on people remembering to do one extra step when creating modules
2. Have this as a reminder / assistant that adds the entry.
3. Work on some other test/verification to warn when a new `values*/strings.xml` is added

I think this one doesn't hurt and can be easily customisable or disabled if needed in the future, without breaking anything.

I don't like the idea of keeping project-specific extra steps if they can be avoided. Adding Gradle modules should be easy and hassle free.
With this, Crowdin would be almost invisible:

1. Create your `*ModuleName*/build.gradle.kts`
2. Test your implementation locally, build it, etc.
3. Done. Crowdin was set automatically in step 2.

It could be more aggressive as well, and do it at configuration phase. Not sure if better or worse.
It also could be way more overkill and wait until `:app` is configured, to check all module dependencies and check if they have `src/res/` string files and _then_ build the `crowdin.yml` file.

sbakhtiarov (approved):

MohamadJaara (approved):

PR #4548, opened 2026-01-27
fix: update last notified correctly, notification delay (WPB-22701)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-22701

https://wearezeta.atlassian.net/browse/WPB-22701
----

# What's new in this PR?

### Issues
Application was not showing all notification when receiving a lot of messages.

### Causes (Optional)
1. Messages for notifications are queried from the `Messages` table based on the message timestamp and conversation `last_notified_date`. Last notified date was not updated correctly and some messages were filtered out by query.

2. Notification updates are based on the `activeNotifications` property of `NotificationManager`. New notifications for same conversation are combined with already shown notifications. When multiple notifications are shown very fast `activeNotifications` property does not always have up-to-date list of notifications. This is causing incorrect updates based on out-of-date list.

### Solutions
1. Get the timestamp of the last notified message and use it as last_notified_date for conversation.
2. Add small delay after adding/updating notification to let the NotificationManager update `activeNotifications`.


saleniuk (approved):

app/src/main/kotlin/com/wire/android/notification/MessageNotificationManager.kt
```diff
@@ -52,14 +53,16 @@ class MessageNotificationManager
...
+    suspend fun handleNotification(newNotifications: List<LocalNotification>, userId: QualifiedID, userName: String) {
         if (newNotifications.isEmpty()) return
 
         addNotifications(newNotifications, userId, userName)
         updateNotifications(newNotifications, userId)
         removeSeenNotifications(newNotifications, userId)
 
+        // This delay is required to let notification manager update activeNotifications list
+        delay(NOTIFICATION_UPDATE_DELAY)
+
```
MohamadJaara, line 65: is this delay needed now since we update the ID with the correct one from the last notified ?

sbakhtiarov: I kept it here for two reasons:
- There will be less questions from QA. Without this delay you can still reproduce the issue by sending messages fast.
- It will not hide potential message delivery issues. If we see that there is a notification missing for some message we know it is not the NotificationManager issue. Something else is wrong.

MohamadJaara (approved):

PR #4569, opened 2026-02-05
feat: app lock forgot logout navigation [WPB-23188]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-23188

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- The “Forgot App Lock” flow needs to switch from “Remove device” to a logout flow that preserves history unless the user opts to wipe data.
- After logout, the app should always return to the correct login flow without leaving Home on the back stack.

### Causes (Optional)

- The previous flow and copy were built around device removal.
- Login navigation was triggered within AppLockActivity, leaving WireActivity (Home) behind.

### Solutions

- Reworked “Forgot App Lock” to use a logout-based flow (soft or hard based on the checkbox) and updated all related copy.
- After logout, AppLockActivity now launches WireActivity with CLEAR_TASK/NEW_TASK and finishes itself, so the login flow is the only thing on the stack.


app/src/main/kotlin/com/wire/android/ui/home/appLock/forgot/ForgotLockCodeScreen.kt
```diff
@@ -180,37 +176,45 @@ fun ForgotLockCodeScreenContent(
...
     modifier: Modifier = Modifier.fillMaxWidth(),
 ) {
     val interactionSource = remember { MutableInteractionSource() }
     Column(modifier = modifier) {
         WirePrimaryButton(
-            text = stringResource(R.string.settings_forgot_lock_screen_reset_device),
-            onClick = onContinue,
+            text = stringResource(R.string.user_profile_logout),
+            onClick = onLogout,
             state = if (enabled) WireButtonState.Default else WireButtonState.Disabled,
```
saleniuk: Maybe we could add also `loading` state to this button? Logging out can take some time so it would be better to show loading spinner on the disabled button while user is waiting 🤔 
Just add `loading = !enabled` or even change the name of `enabled` parameter to `loggingOut` or something 😄

saleniuk (approved):

sbakhtiarov (approved):

PR #4574, opened 2026-02-06
chore: update kotliun and migrate navigation 

MohamadJaara:

- The **PR Title**
    - [ ] contains a reference JIRA issue number like `SQPIT-764` (replace placeholder below)

- The **PR Description**

----

# What's new in this PR?

### Issues

- Kotlin/build tooling and navigation dependencies were outdated for the current Compose Destinations setup.
- Navigation graph wiring relied on manual graph composition and copied shared annotations across modules.
- Kalium needed alignment with the updated toolchain.

### Causes (Optional)

- The previous navigation setup depended on older Compose Destinations patterns and a custom annotation-copying flow.
- Toolchain upgrades (Kotlin/AGP/Gradle/KSP) required API and build-logic adjustments.

### Solutions

- Upgraded core toolchain and deps:
  - Kotlin `2.1.0 -> 2.3.0 -> 2.3.10`
  - AGP `8.10.1 -> 9.0.0`
  - Gradle wrapper `8.11.1 -> 9.1.0`
  - Compose BOM `2025.09.00 -> 2026.01.00`
  - Compose Navigation `2.7.7 -> 2.9.5`
  - Compose Destinations `1.11.9 -> 2.3.0` (`animations-core` -> `core`)
  - KSP `2.1.0-1.0.28 -> 2.3.4`
  - Hilt `2.55 -> 2.59.1`
- Migrated navigation to new graph structure:
  - Replaced `WireMainNavGraph` with `@NavHostGraph` root and external feature graphs (`CellsGraph`, `SketchGraph`).
  - Introduced module-local destination annotations for app/cells/sketch.
  - Removed custom copied-annotation KSP flow from build logic.

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test
----

----

- [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated (no new config variables introduced).

----
##### References


app/src/main/kotlin/com/wire/android/navigation/annotation/app/AppDestinations.kt
```diff
@@ -0,0 +1,124 @@
...
+import com.wire.android.ui.authentication.login.NewLoginNavGraph
+import com.wire.android.ui.home.HomeNavGraph
+import com.wire.android.ui.home.newconversation.common.NewConversationNavGraph
+import com.wire.android.ui.userprofile.teammigration.PersonalToTeamMigrationNavGraph
+import kotlin.reflect.KClass
+
+@Destination<WireRootNavGraph>(
+    wrappers = [WaitUntilTransitionEndsWrapper::class, TabletDialogWrapper::class],
+)
+internal annotation class WireRootDestination(
```
saleniuk, line 39: Previous `WireDestination` was a single annotation that could be used everywhere, now we have separate annotations for each sub graph, it's fine but more boilerplate and someone may forget about it.
Previous KSP action was there to copy this `WireDestination` from `core/navigation` to each module that uses navigation, because using annotations from other modules don't work, so thanks to that KSP action each module automatically had this destination copied inside and could use it without the need to manually declare it.

We could still use the same logic - have one generic `WireDestination<>` created in `core/navigation` like so:
``` @Repeatable
@Destination<Nothing>( // in this case, because "WireDestination" is opening the graph choice, this graph here does not matter
    wrappers = [WaitUntilTransitionEndsWrapper::class, TabletDialogWrapper::class],
)
annotation class WireDestination<T: Annotation>(
    val route: String = COMPOSABLE_NAME,
    val start: Boolean = false,
    val navArgs: KClass<*> = Nothing::class,
    val style: KClass<out DestinationStyle> = DestinationStyle.Default::class,
)
```
 and use the same KSP action to copy it into each module and then use it everywhere like this: `@WireDestination<WireRootNavGraph>`

What do you think?

app/src/main/kotlin/com/wire/android/ui/home/conversations/details/editguestaccess/EditGuestAccessScreen.kt
```diff
@@ -62,9 +62,9 @@ import com.wire.android.util.copyLinkToClipboard
 import com.wire.android.util.shareViaIntent
 
 @Suppress("ComplexMethod")
-@WireDestination(
-    navArgsDelegate = EditGuestAccessNavArgs::class,
-    style = DestinationStyle.Runtime::class, // default should be SlideNavigationAnimation
+@WireRootDestination(
+    navArgs = EditGuestAccessNavArgs::class,
+    style = SlideNavigationAnimation::class, // default should be SlideNavigationAnimation
```
saleniuk, line 67: Why the style is changed back to `SlideNavigationAnimation`?
Because of that it won't be shown on tablets properly but as a full screen just like on phones

app/src/main/kotlin/com/wire/android/navigation/TabletStyleHelper.kt
```diff
@@ -18,48 +18,8 @@
...
-    EditConversationNameScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    EditGuestAccessScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    UpdateAppsAccessScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    ChannelAccessOnUpdateScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    EditSelfDeletingMessagesScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    ConversationFoldersScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    NewConversationFolderScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-    ChangeUserColorScreenDestination.style = if (isTablet) DialogNavigation else SlideNavigationAnimation
-}
+fun AdjustDestinationStylesForTablets() = Unit
```
saleniuk, line 27: Is this intended? Because of that and removing `DestinationStyle.Runtime` from all these screens, tablet styled screens are not supported anymore

MohamadJaara: yes i tried to keep it in the same logic it didn't work i will spend some time looking if we can do the same in another way

AndroidBob:
# New ADR(s) in this PR 📚:
# 10. Tablet dialog navigation parity after Compose Destinations upgrade

Date: 2026-02-06

## Status

Accepted

## Context

Before the Kotlin/navigation migration (commit `18a563754`), the app used `DestinationStyle.Runtime` and `AdjustDestinationStylesForTablets()` to switch 17 destinations at runtime:

- tablet -> `DialogNavigation`
- phone -> `SlideNavigationAnimation` or `PopUpNavigationAnimation`

After the migration, generated destination styles became immutable (`val style`), so mutating `Destination.style` at runtime was no longer possible. This caused tablet screens to be shown as phone-style transitions, and not as true dialogs.

We also need to preserve:

- true dialog presentation semantics on tablet (not only animation changes)
- existing phone behavior for the same destinations
- one source of truth for the target destinations

## Decision

We introduced an app-level nav host engine wrapper and centralized route policy:

1. `WireNavHostEngine` (app module) mirrors default Compose Destinations engine behavior.
2. For each destination registration, style is resolved by `resolveTabletDialogParityStyle(...)`:
   - if tablet + route is in parity list -> force `DialogNavigation`
   - otherwise -> keep existing behavior (`manualAnimation ?: destination.style`)
3. The parity list of 17 routes is defined in `TabletDialogRoutePolicy`.
4. `TabletDialogWrapper` receives the same route matcher and applies rounded clipping for tablet parity routes as well, so visual corners match dialog expectations.

Implementation files:

- `app/src/main/kotlin/com/wire/android/navigation/WireNavHostEngine.kt`
- `app/src/main/kotlin/com/wire/android/navigation/TabletDialogRoutePolicy.kt`
- `core/navigation/src/main/kotlin/com/wire/android/navigation/wrapper/TabletDialogWrapper.kt`

Both app nav hosts use `rememberWireNavHostEngine(...)`:

- `MainNavHost`
- nested host in `HomeScreen`

## Consequences

### Positive

- Restores pre-migration tablet behavior for the 17 destinations as true dialogs.
- Preserves phone behavior without changing navigation call sites.
- Keeps route ownership centralized, making future additions/removals explicit.
- Keeps compatibility with Compose Destinations `2.3.x` immutable destination style generation.

### Trade-offs

- We maintain a small custom engine implementation that mirrors upstream behavior and should be kept in sync when upgrading Compose Destinations.
- Wrapper clipping now depends on an injected route matcher, which is another integration point to keep aligned with route policy.

### Maintenance rules

When adding/removing tablet dialog parity screens:

1. Update `TabletDialogRoutePolicy.destinationBaseRoutes`.
2. Add/update tests in `TabletDialogRoutePolicyTest`.
3. Verify both tablet and phone behavior manually.

When upgrading Compose Destinations:

1. Diff upstream default engine behavior against `WireNavHostEngine`.
2. Re-validate style resolution and manual animation precedence.

app/src/main/kotlin/com/wire/android/navigation/annotation/app/AppDestinations.kt
MohamadJaara: imho it is too much for not much reward, adding copying step will complicate it without huge benefit.however i will open a seperate PR to explore this idea ithing it is worth exporing at least

MohamadJaara:
this PR doubled in size since it was opened will be closed and open smaller pnes

PR #4611, opened 2026-02-23
fix: suspend function in copy method (WPB-20871)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-20871

https://wearezeta.atlassian.net/browse/WPB-20871
----

# What's new in this PR?

### Issues
Inconsistent behavior of the Read Receipt toggle in Conversation Options.

### Causes (Optional)
Calling suspend function in data class copy method.

### Solutions
- Moved suspend function call out of the copy method.
- Fixed same issue in 5 other view model classes.


MohamadJaara (approved):
good catch

Garzas (approved):

PR #4612, opened 2026-02-23
feat: Wire drive filter (WPB-23843)

ohassine:
https://wearezeta.atlassian.net/browse/WPB-23843

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Description

Implement filter feature in shared drive

https://github.com/user-attachments/assets/9a696198-3a58-48f8-8128-efb21f3f61b4

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


core/ui-common/src/main/kotlin/com/wire/android/ui/common/chip/WireFilterChip.kt
```diff
@@ -41,39 +46,74 @@ fun WireFilterChip(
...
 ) {
 
-    val rotationAngle by animateFloatAsState(
-        targetValue = if (isSelected) 0f else 45f,
-    )
-
     FilterChip(
         modifier = modifier.wrapContentSize(),
-        onClick = { onSelectChip(label) },
+        onClick = { onClick(label) },
```
sbakhtiarov, line 57: Minor comment: "label" must be always passed by the WireFilterChip caller, no need to return it back in onClick(label). Could be just onClick().

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/SearchScreenViewModel.kt
```diff
@@ -0,0 +1,276 @@
...
+                 query = params.query,
+                 fileFilters = FileFilters(
+                     tags = params.tagIds,
+                     owners = params.ownerIds,
+                     mimeTypes = params.mimeTypes,
+                     hasPublicLink = params.filesWithPublicLink
+                 ),
+             ).map { pagingData: PagingData<Node> ->
+                 pagingData.map { node: Node ->
+                     if (uiState.value.availableOwners.isEmpty()) {
```
sbakhtiarov: I do not understand how this is supposed to work. Why do we launch a new coroutine here (and only if not empty:
- It does not handle duplicates. If there are 10 files from same user it will launch 10 coroutines to get user info
- It stops adding new users if availableUsers already have some values (?). There will be no filters for users from the second page of files.

I think it's worth to re-think approach to this filter logic.

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/owner/FilterByOwnerBottomSheet.kt
```diff
@@ -0,0 +1,269 @@
...
+    val state = rememberOwnersFilterSheetState(items)
+    LaunchedEffect(sheetState) { sheetState.show() }
+
+    val searchState = remember { TextFieldState() }
+    LaunchedEffect(searchState) {
+        snapshotFlow { searchState.text.toString() }
+            .collect(state::onQueryChange)
+    }
+
+    LaunchedEffect(sheetState) { sheetState.show() }
```
sbakhtiarov: There is already same line above `LaunchedEffect(sheetState) { sheetState.show() }`

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/SearchUiState.kt
```diff
@@ -0,0 +1,44 @@
...
+data class SearchUiState(
+    val availableTags: List<FilterTagUi> = emptyList(),
+    val availableOwners: List<FilterOwnerUi> = emptyList(),
+    val availableTypes: List<FilterTypeUi> = TypeFilter.typeItems,
+
+    val showFilterByType: Boolean = false,
+    val showFilterByTags: Boolean = false,
+    val showFilterByOwner: Boolean = false,
+
+    val filesWithPublicLink: Boolean = false,
```
sbakhtiarov, line 30: Not sure if it is a bug or a feature. My assumption about this flag is that it must support three values:
- null - all files
- true - only files with public link
- false - only files without public link

sbakhtiarov (changes requested):

core/ui-common/src/main/kotlin/com/wire/android/ui/common/chip/WireFilterChip.kt
ohassine: yes I will remove it after removing `FilterBottomSheet`

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/SearchScreenViewModel.kt
ohassine: Yeah changed this by calling a separate usecase that fetches owners

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/SearchUiState.kt
ohassine: it's because API limitation, we only have two states for now: files with public link or all

sbakhtiarov: Maybe we can still handle it properly on our side? If they fix this on server side the app will not work properly.

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
```diff
@@ -220,57 +222,53 @@ fun ConversationFilesScreenContent(
...
+                        modifier = Modifier
+                            .sharedElement(
+                                sharedContentState = rememberSharedContentState(key = SHARED_ELEMENT_SEARCH_INPUT_KEY),
+                                animatedVisibilityScope = animatedVisibilityScope
+                            ),
+                        isSearchActive = conversationSearchBarState.isSearchActive,
+                        searchBarHint = stringResource(R.string.search_shared_drive_text_input_hint),
+                        searchQueryTextState = conversationSearchBarState.searchQueryTextState,
+                        onActiveChanged = conversationSearchBarState::searchActiveChanged,
+                        onTap = {
```
saleniuk, line 258: Do we really need a separate `onTap`? Can't it be done using `onActiveChanged`?
With `onTap` it will only work when user clicks on the search input, but won't work when user switches focus in a different way, and it would work when using `onActiveChanged`.

Also, I think that the UX is wrong when user navigates from `SearchConversationMessagesScreen` and clicks "Search Files" - then `ConversationFilesScreen` opens with search bar focused, so user has two "back" buttons on right next to each other, and actually needs to tap the search bar again, even though it's already focused, to actually navigate to `SearchScreen` and be able to use all these filters.
Take a look at the video:

https://github.com/user-attachments/assets/bbb0f3dc-e4ff-4682-9d44-c16f2c344113

core/ui-common/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchTopBar.kt
```diff
@@ -65,52 +65,50 @@ fun SearchTopBar(
...
     modifier: Modifier = Modifier,
     isLoading: Boolean = false,
     backIconContentDescription: String? = null,
     searchBarDescription: String? = null,
     onCloseSearchClicked: (() -> Unit)? = null,
     onActiveChanged: (isActive: Boolean) -> Unit = {},
-    bottomContent: @Composable ColumnScope.() -> Unit = {}
+    bottomContent: @Composable ColumnScope.() -> Unit = {},
+    onTap: (() -> Unit)? = null,
+    focusManager: FocusManager = LocalFocusManager.current,
```
saleniuk: You don't need to make it a parameter that needs to be passed from the parent and actually you shouldn't as that's the whole point of `LocalFocusManager` so that you can use it without the need of passing it down.

core/ui-common/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchTopBar.kt
```diff
@@ -65,52 +65,50 @@ fun SearchTopBar(
     isSearchActive: Boolean,
     searchBarHint: String,
     searchQueryTextState: TextFieldState,
+    focusRequester: FocusRequester,
```
saleniuk: Is this `focusRequester` needed to be a parameter? I don't see it being used anywhere outside of this `SearchTopBar` so it could stay just inside this composable, just as it was before.

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/owner/FilterByOwnerBottomSheet.kt
```diff
@@ -0,0 +1,231 @@
...
+        snapshotFlow { searchState.text.toString() }
+            .collect(state::onQueryChange)
+    }
+
+    fun dismiss() {
+        scope.launch { sheetState.hide() }
+            .invokeOnCompletion { onDismiss() }
+    }
+
+    ModalBottomSheet(
```
saleniuk: Why not `WireModalSheetLayout` to keep all our bottom sheets unified?

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/owner/FilterByOwnerBottomSheet.kt
```diff
@@ -0,0 +1,231 @@
...
+import com.wire.android.ui.common.preview.MultipleThemePreviews
+import com.wire.android.ui.common.typography
+import com.wire.android.ui.theme.WireTheme
+import com.wire.android.ui.theme.wireColorScheme
+import kotlinx.coroutines.launch
+import com.wire.android.ui.common.R as CommonR
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun FilterByOwnerBottomSheet(
```
saleniuk, line 66: All bottom sheets have this weird jumping when user flings and it reaches the end: 

https://github.com/user-attachments/assets/339f8060-7502-4bd6-b74d-f7d368cb1031

It looks like some general `ModalBottomSheet` issue:https://issuetracker.google.com/issues/285847707
It's happening for all our bottom sheets, it's just more prominent in these ones as they have nested `LazyColumn`.
Google fixed that in material3 `v1.5.0-alpha10`, just FYI, save yourself from trying to fix it. 😄

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/owner/FilterByOwnerBottomSheet.kt
```diff
@@ -0,0 +1,231 @@
...
+) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .clickable { onToggle() }
+            .padding(
+                start = dimensions().spacing12x,
+                end = dimensions().spacing12x,
+                top = dimensions().spacing8x,
+                bottom = dimensions().spacing8x
```
saleniuk, line 169: can be: `horizontal = dimensions().spacing12x, vertical = dimensions().spacing8x`

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
ohassine: That's the expected behavior, when user clicks on the search input.
and yeah we have a ticket for `SearchConversationMessagesScreen` UX issue.

sbakhtiarov (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/ConversationFilesScreen.kt
saleniuk: > That's the expected behavior, when user clicks on the search input. 

In that case `onActiveChanged` will also work fine when user clicks on the search input, but also will work when user changes focus using some other ways, like using external keyboard. 
I believe user shouldn't type and search on the `ConversationFilesScreen`, that's what the `SearchScreen` is for, right? Currently, when using `onTap` it's possible and it creates another bad and confusing UX - when searching, filter options are sometimes available to the user and sometimes not, also search query is lost when clicking on the input:

https://github.com/user-attachments/assets/8ee78da3-3cba-4341-9b83-9b9a76d8f351

This would not happen if `SearchScreen` were opened when search input is active, not only when user taps on it

ohassine: Yes — the expected behavior is to navigate to another screen when the user clicks on the search input.

I set the search input to read-only, so it can no longer receive focus.

saleniuk: Ok, that's fair enough.
It's just that we add new parameters when we could just keep it simple and reuse the ones that we already have.

saleniuk (approved):

PR #4624, opened 2026-03-05
feat: filter, search and sort in Drive screen (WPB-21658)

ohassine:
https://wearezeta.atlassian.net/browse/WPB-21658

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Filter, search and sort in Drive screen

https://github.com/user-attachments/assets/f3b9561e-0894-48d9-b1e9-205775a3ebc9

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/FilterByConversationBottomSheet.kt
```diff
@@ -0,0 +1,222 @@
...
+
+    WireModalSheetLayout(
+        onDismissRequest = ::dismiss,
+        sheetState = sheetState,
+        modifier = modifier,
+    ) {
+        Column(
+            modifier = Modifier
+                .fillMaxWidth()
+                .fillMaxHeight(0.8f)
```
saleniuk, line 100: Maybe we should choose some specific height in dp for these bottom sheets, we probably don't want them to take 80% of screen height on tablets as it would be too much 🤔

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/FilterByConversationBottomSheet.kt
```diff
@@ -0,0 +1,222 @@
...
+                ) { item ->
+                    ConversationRow(
+                        item = item,
+                        onClick = { state.selectConversation(item.id.toString()) }
+                    )
+                    HorizontalDivider()
+                }
+            }
+
+            FooterButtons(
```
saleniuk, line 145: These bottom buttons should be visually separated from the list, we usually put them into surface and elevate when list is scrolled:
```
Surface(
    shadowElevation = scrollState.rememberBottomBarElevationState().value,
    color = MaterialTheme.wireColorScheme.surface,
) {
    FooterButtons(...)
}
```

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/ConversationFilterSheetState.kt
```diff
@@ -0,0 +1,81 @@
...
+    var conversations by mutableStateOf(initialItems)
+        private set
+
+    var query by mutableStateOf("")
+        private set
+
+    val hasChanges: Boolean
+        get() = conversations.any { o -> initialById[o.id]?.selected != o.selected }
+
+    val filteredConversations: List<FilterConversationUi>
```
saleniuk: This is not a state, so the UI may not recompose  when this list changes

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/ConversationFilterSheetState.kt
```diff
@@ -0,0 +1,81 @@
...
+        get() {
+            val q = query.trim()
+            return if (q.isBlank()) {
+                conversations
+            } else {
+                conversations.filter { it.name.contains(q, ignoreCase = true) }
```
saleniuk: This list of conversations can become massive so we shouldn't pass the whole list and filter it in the UI state, but rather pass the search query to the DB query and receive already filtered list, just like we do in all other places

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/FilterByConversationBottomSheet.kt
ohassine: `0.8f` is the max height not the height

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/ConversationFilterSheetState.kt
ohassine: I fixed this in this follow up PR https://github.com/wireapp/wire-android/pull/4630

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/ConversationFilterSheetState.kt
ohassine: This is done on purpose on the app side to avoid debounce and give instant feedback as the user types.

The list only shows GROUP conversations with Cells ENABLED. I don't think we'll ever deal with thousands of items here. Even with 800 conversations, the filter runs in ~0.5ms since `FilterConversationUi` is a lightweight data class (just a ConversationId, 2 strings, and 3 booleans)

features/cells/src/main/java/com/wire/android/feature/cells/ui/AllFilesScreen.kt
```diff
@@ -62,68 +69,61 @@ fun AllFilesScreen(
...
+
+    WireScaffold(
+        modifier = modifier,
+        topBar = {
+            Column {
+                SearchTopBar(
+                    modifier = Modifier,
+                    isSearchActive = false,
+                    searchBarHint = stringResource(R.string.search_label),
+                    searchQueryTextState = TextFieldState(),
```
sbakhtiarov: I think there is a rememberTextFieldState() function which we can use to avoid creating a new instance on each recomposition.

sbakhtiarov (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/AllFilesScreen.kt
ohassine: `searchQueryTextState` is not used as search has been moved to SearchScreen, but still better to use rememberTextFieldState

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/FilterByConversationBottomSheet.kt
saleniuk: Yep, so "max height" still means that it can take 80% of the screen on tablets, if there are many items on the list

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/ConversationFilterSheetState.kt
saleniuk: Debounce is something we add, it's not by default by the database query, so it's up to us whether we debounce search or not, we can also make it so that query is executed each time user enters new character.
Databases are usually more efficient than us doing it manually, and right, it will never be that many elements but still, we pass the whole list unnecessarily where we could filter it right away and work and pass less data.
It's also just about the separation of concerns - composable state should be responsible for presenting the data, not processing it as it should be done in the data layer, not presentation layer.

ohassine: The list is already in memory, we're not fetching new data, just filtering what's already loaded from the start
and for small in-memory lists, list.filter {} is faster than a database round-trip

saleniuk: The list is already in memory because we fetched that specifically for that purpose. 😄  So as you can see this is also not a very efficient use of resources as we need to fetch and keep the whole list in memory even if user doesn't want to use that list and filter by conversations (how often will that happen?). 😅 
Why not do it properly from the start? Maybe already with pagination? We have already had several examples where we created something without thinking about how it would work on a larger scale and had to make costly refactoring. 🙈 
I would also still argue that clear separation of concerns is something that we should care about to keep our code clean.

ohassine: I intentionally fetched all conversations so Bottomsheet opens instantly with data instead of seeing seeing some loading bar, it's more smoother and responsive experience
Using DAO-based search would also add some extra complexity (debounce, additional flows/queries, etc.) without bringing a clear benefit at this scale for such a small feature

I don’t think this is a violation of separation of concerns. The DAO is responsible for data access, but once the data is loaded, applying a lightweight filter in the UI state is just a presentation concern. with ~500 items, in‑memory filtering is more efficient and straightforward for such a small modal

MohamadJaara (approved):

features/cells/src/main/java/com/wire/android/feature/cells/ui/search/filter/bottomsheet/conversation/ConversationFilterSheetState.kt
saleniuk: I just don't understand why it's considered a completely different thing than fetching conversations in other places and why assuming that list of conversations with Wire Drive will always be small - once it becomes available and gains popularity then the ratio of conversations that have it enabled to all conversations can change rapidly.
Then why we may still think that here, fetching conversations is a small thing and we can fetch all at once, but on conversation list doing similar thing quickly became so complex that it required implementing dedicated queries with search and filters and adding pagination and we had to make many refactors just to optimize it.
We already have such logic so it would be just a matter or reusing it or making a similar, simpler copy.
So yes, for now it's quicker, but will we be able to catch the moment when the list grows so that it takes too long to be shown on the bottom sheet?
I just don't get it why not take a safer approach from the start.

PR #4627, opened 2026-03-06
feat(nomad profiles): nomad profiles URL parsing and signature verification (WPB-23837)

marcoconti83:
https://wearezeta.atlassian.net/browse/WPB-23837

# What's new in this PR?

When starting an autologin intent, the app also parses an optional nomad profiles service host. The paramater is expected to be signed, i.e. there should be a signature passed to the intent.

The app will check that the parameter signatures one of the public keys hardcoded in the application. If it doesn't, it rejects it.

WARNING: This PR targets the `marco/nomad-profiles` branch because it should be merged AFTER that one.


app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/LogoutReceiver.kt
```diff
@@ -0,0 +1,81 @@
...
+            is CurrentSessionResult.Success -> {
+                val userId = session.accountInfo.userId
+                appLogger.i("$TAG Logging out user: ${userId.toLogString()}")
+                coreLogic.getSessionScope(userId).logout(LogoutReason.SELF_HARD_LOGOUT, waitUntilCompletes = true)
+                deleteSession(userId)
+                accountSwitch(SwitchAccountParam.TryToSwitchToNextAccount)
+                val wireActivityIntent = Intent(context, WireActivity::class.java).apply {
+                    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK)
+                }
+                context.startActivity(wireActivityIntent)
```
sbakhtiarov: This could throw SecurityException if application is in background.

app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/NomadLogoutReceiver.kt
sbakhtiarov: This could throw SecurityException if application is in background.
What is the purpose of starting this activity here?

app/src/main/kotlin/com/wire/android/ui/initialsync/InitialSyncScreen.kt
sbakhtiarov: The code in InitialSyncScreen is important and seems very fragile.
May be called twice (one per mutableState parameter change) and depends on the order in which we set value for these states:
If we change the ViewModel later and it will set isSyncCompleted first and then shouldMoveToBackground there is a possibility that LaunchedEffect may be called twice and first time it will not have the correct value for shouldMoveToBackground.

sbakhtiarov, line 53: Maybe we could use withTimeout here to avoid waiting forever if HomeScreen does not appear in back stack?

app/src/main/kotlin/com/wire/android/ui/initialsync/InitialSyncViewModel.kt
sbakhtiarov: Suggest combining these two in one state class.

app/src/main/kotlin/com/wire/android/ui/WireActivityActionsHandler.kt
sbakhtiarov: This logic and loginTypeSelector.canUseNewLogin(serverLinks) should (must?) be done in the ViewModel.
Another good reason for that is calling navigator.navigate from the coroutine which is not connected to the lifecycle. Navigation may fail.

app/src/main/kotlin/com/wire/android/util/lifecycle/AutomatedLoginManager.kt
sbakhtiarov: The "pendingMoveToBackgroundAfterSync" is read / written from coroutines and is not Volatile/Atomic.

app/src/main/AndroidManifest.xml
sbakhtiarov, line 323: Is it safe to expose this intent to all applications on the device?

MohamadJaara: yes since the data would be signed by us plus the whole flow needs to be triggered from other apps

app/src/main/kotlin/com/wire/android/util/lifecycle/IntentsProcessor.kt
```diff
@@ -61,4 +106,37 @@ class IntentsProcessor @Inject constructor() {
         val uri = runCatching { URI(url) }.getOrNull()
         return uri?.scheme?.lowercase() == "https" && !uri.host.isNullOrEmpty()
     }
+
+    private fun isValidSignature(parameter: String?, signature: String?): Boolean {
```
ohassine: It wold be cleaner to have this in a separate class like `SignatureValidator`

ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/initialsync/InitialSyncScreen.kt
MohamadJaara: no need imo the idea is to move the app to the background as soon as it is done with inital sync

sbakhtiarov (approved):

PR #4659, opened 2026-03-17
fix: failure message for adding user with no mls client (WPB-17105)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-17105

https://wearezeta.atlassian.net/browse/WPB-17105
----

# What's new in this PR?

### Issues
When adding user with no MLS client application is showing Federation error and support page in conversation.

### Causes (Optional)
MissingKeyPackages error is always mapped to Federation error.

### Solutions
Using new MissingKeyPackages error to show correct system message.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -276,6 +276,7 @@
     <string name="url_learn_about_search" translatable="false">https://support.wire.com/hc/articles/203121850</string>
     <string name="url_legal_hold_learn_more" translatable="false">https://support.wire.com/hc/articles/360002018278</string>
     <string name="url_message_details_offline_backends_learn_more" translatable="false">https://support.wire.com/hc/articles/9357718008093</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -276,6 +276,7 @@
     <string name="url_learn_about_search" translatable="false">https://support.wire.com/hc/articles/203121850</string>
     <string name="url_legal_hold_learn_more" translatable="false">https://support.wire.com/hc/articles/360002018278</string>
     <string name="url_message_details_offline_backends_learn_more" translatable="false">https://support.wire.com/hc/articles/9357718008093</string>
+    <string name="url_mls_learn_more" translatable="false">https://support.wire.com/hc/en-us/articles/12434725011485-Messaging-Layer-Security-MLS</string>
```
saleniuk: I would suggest to remove the locale part, we usually use these urls without it, so that the article will open using the current device's locale. You can also shorten this url, just like all others, to have only id of the article, like so: `https://support.wire.com/hc/articles/12434725011485`

yamilmedina (approved):

saleniuk (approved):

PR #4682, opened 2026-03-27
feat(nomad devices): Block nomad intent for Wire production backends and restrict logout intent to nomad sessions [WPB-24366]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-24366

- The **PR Title**

- The **PR Description**

---

# What's new in this PR?

### Issues

- Nomad intents should not be processed when the resolved backend API points to Wire's production infrastructure, as nomad is intended for on-premises deployments only.
- The nomad logout broadcast receiver logs out any active session regardless of whether it was created via the nomad flow.

### Solutions

- Added a production API check in `WireActivityViewModel.handleIntentsThatAreNotDeepLinks()` that rejects the nomad intent if the resolved `ServerConfig.Links.api` host matches Wire's production backend. This uses the same logic as kalium's internal `isProductionApi()` (with a TODO to consolidate once the kalium function is made public).
- Added a nomad account check in `NomadLogoutReceiver` using the existing `DoesValidNomadAccountExistUseCase`. The logout broadcast is now ignored if the current session is not a nomad session.

#### Test Coverage

- [x] I have added automated tests to this contribution

#### How to Test

1. **Production backend block:** Verify that a nomad intent whose `backendConfig` resolves to `prod-nginz-https.wire.com` is silently rejected (log: `"Nomad login ignored: resolved backend is Wire production"`).
2. **Nomad-only logout:** Verify that sending `com.wire.ACTION_LOGOUT` broadcast when logged in via a non-nomad session does nothing. When logged in via a nomad session, the logout proceeds as before.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/NomadLogoutReceiver.kt
+++ b/app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/NomadLogoutReceiver.kt
@@ -62,6 +62,11 @@ class NomadLogoutReceiver : CoroutineReceiver() {
 
         appLogger.i("$TAG Received logout broadcast")
 
+        if (!coreLogic.getGlobalScope().doesValidNomadAccountExist()) {
+            appLogger.i("$TAG Logout ignored: current session is not a nomad account")
+            return
+        }
+
         @Suppress("TooGenericExceptionCaught")
         try {
             performLogout()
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/NomadLogoutReceiver.kt
```diff
@@ -62,6 +62,11 @@ class NomadLogoutReceiver : CoroutineReceiver() {
 
         appLogger.i("$TAG Received logout broadcast")
 
+        if (!coreLogic.getGlobalScope().doesValidNomadAccountExist()) {
+            appLogger.i("$TAG Logout ignored: current session is not a nomad account")
+            return
+        }
+
```
ohassine: I'd suggest to also check if there is a valid session to perform logout

ohassine (approved):

app/src/main/kotlin/com/wire/android/notification/broadcastreceivers/NomadLogoutReceiver.kt
MohamadJaara: hmmm how about i make it to chekc the current session if it is a nomad or not since for nomad we will always have one account and one only 
this way we avoid any issues

yamilmedina (approved):

PR #4687, opened 2026-04-01
fix: stabilize message media loading [WPB-18914]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-18914

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Opening a conversation with a large backlog of missed messages could leave the message list in an unstable loading state.
- Instead of showing a single loading phase, the conversation rendered messages progressively one by one, with many per-message loading indicators visible in the list.
- Media loading inside the conversation was tightly coupled to the message list read model, which also increased the risk of visual churn while messages were being loaded.

### Causes (Optional)

- The conversation paging query depended on local asset cache data through `MessageDetailsView`, so writes to the `Asset` table could invalidate the message list while the conversation was loading.
- Local asset paths were propagated through the message read model, even though they represent local runtime/cache state rather than stable message data.
- Some media UI flows relied on local asset path data being present directly in the message model instead of resolving it on demand for the specific item.

### Solutions

- Removed the dependency between conversation message paging and local asset cache writes by detaching `MessageDetailsView` from the `Asset` table and adding a forward DB migration.
- Removed local asset path from the message read model and stopped treating it as part of the main message state.
- Switched media components to resolve local asset paths on demand for the specific message item instead of loading them through the whole conversation list model.
- Updated image/video/audio-related flows so they no longer rely on local asset path being embedded in the message read model.
- Fixed follow-up media regressions caused by this refactor, including item flicker/blinking and concurrent temporary file collisions during asset downloads.


app/src/main/kotlin/com/wire/android/media/audiomessage/AudioMessageViewModel.kt
```diff
@@ -92,9 +96,15 @@ class AudioMessageViewModelImpl @Inject constructor(
     }
 
     private fun preloadAudioMessage() {
+        if (preloadStates.putIfAbsent(args.key, PreloadState.InFlight) != null) return
         viewModelScope.launch {
-            // calls preload to initially fetch the audio asset data to be ready and schedule waves mask generation if needed
-            audioMessagePlayer.preloadAudioMessage(args.conversationId, args.messageId)
+            try {
+                // calls preload to initially fetch the audio asset data to be ready and schedule waves mask generation if needed
+                audioMessagePlayer.preloadAudioMessage(args.conversationId, args.messageId)
```
saleniuk, line 98: `ConversationAudioMessagePlayer` is a singleton, wouldn't be better to have this logic of checking and not preloading the same audio message again inside this singleton player instead of having this not so pretty approach here with keeping a map in companion object?

app/src/main/kotlin/com/wire/android/media/audiomessage/AudioMessageViewModel.kt
```diff
@@ -110,7 +120,10 @@ class AudioMessageViewModelImpl @Inject constructor(
                 }
                 .distinctUntilChanged()
                 .firstOrNull { it != null } // wait for the first non-null value
-                .let { state = state.copy(wavesMask = it) }
+                ?.let {
+                    cachedWavesMasks[args.key] = it
```
saleniuk: Waves masks are being generated only once and then are stored in the DB, so there isn't much overhead here. `fetchWavesMask` waits until the waves mask is generated, or just takes the value from DB if it's already there.

I don't understand this idea of having `cachedWavesMasks` - `fetchWavesMask` method is still executed, even if we initially already have a wave mask in `AudioMessageState`, so there's no advantage.

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/MessageContentAndStatus.kt
```diff
@@ -143,21 +150,67 @@ private fun MessageContent(
 ) {
     when (messageContent) {
         is UIMessageContent.ImageMessage -> {
+            val viewModel: AssetLocalPathViewModel =
+                hiltViewModelScoped<AssetLocalPathViewModelImpl, AssetLocalPathViewModel, AssetLocalPathArgs>(
+                    AssetLocalPathArgs(message.conversationId, message.header.messageId)
+                )
+            var rememberedAssetDataPath by rememberSaveable(message.header.messageId) {
```
saleniuk: I'm not sure if I get that but isn't it redundant?
I mean, `viewModel.localAssetPath` is already a composable state, and it's stored in a ViewModel so it survives when the screen recomposes, why then do we need also `rememberSaveable` variable here?

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/AssetLocalPathViewModel.kt
```diff
@@ -0,0 +1,121 @@
...
+                    if (resolvingJob === this@launch) {
+                        resolvingJob = null
+                    }
+                }
+            }
+        }
+    }
+
+    private companion object {
+        val cachedLocalAssetPaths = ConcurrentHashMap<String, String>()
```
saleniuk: I'm still not convinced whether it's a good approach. I think I would rather prefer having some in-memory cache in kalium, like we do with `FlowCache` for instance, because this approach honestly feels like an anti-pattern, it breaks the separation of concerns, mixes logic layer with storage and makes it so that the data doesn't flow the way it should so it can also make it harder to test.

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/SwipeableMessageBox.kt
```diff
@@ -73,16 +73,21 @@ internal fun SwipeableMessageBox(
     modifier: Modifier = Modifier,
     content: @Composable () -> Unit,
 ) {
+    if (configuration is SwipeableMessageConfiguration.NotSwipeable) {
+        content()
+        return
+    }
+
     SwipeableBox(
```
saleniuk: I fixed some excessive recompositions some time ago: https://github.com/wireapp/wire-android/pull/4398
This change will actually introduce a regression and make it so that these unwanted recompositions start happening again, because when the message's state changes, then `message.isSwipeable` changes and depending on that we will again either wrap the content in SwipeableMessageBox or not which causes the whole message’s content to recompose.

saleniuk (changes requested):

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/SwipeableMessageBox.kt
```diff
@@ -132,41 +135,47 @@ private fun SwipeableBox(
...
-        val dragState = remember(onSwipeLeft, onSwipeRight) {
-            AnchoredDraggableState(
-                initialValue = SwipeAnchor.CENTERED,
-                positionalThreshold = { dragWidth },
-                velocityThreshold = { screenWidth },
-                snapAnimationSpec = tween(),
-                decayAnimationSpec = splineBasedDecay(density),
-                anchors = DraggableAnchors {
+        val dragState: AnchoredDraggableState<SwipeAnchor>? = if (hasSwipeActions) {
+            remember(onSwipeLeft, onSwipeRight) {
```
saleniuk: Maybe put `hasSwipeActions` inside `remember` so that it remembers the current state even if it doesn't have swipe actions?
Like:
```
val dragState: AnchoredDraggableState<SwipeAnchor>? = remember(onSwipeLeft, onSwipeRight) {
    if(onSwipeRight != null || onSwipeLeft != null) {
        AnchoredDraggableState(...)
    } else {
        null
    }
}
```

saleniuk (approved):

app/src/main/kotlin/com/wire/android/media/audiomessage/AudioMessageViewModel.kt
```diff
@@ -92,8 +92,12 @@ class AudioMessageViewModelImpl @AssistedInject constructor(
 
     private fun preloadAudioMessage() {
         viewModelScope.launch {
-            // calls preload to initially fetch the audio asset data to be ready and schedule waves mask generation if needed
-            audioMessagePlayer.preloadAudioMessage(args.conversationId, args.messageId)
+            try {
+                // calls preload to initially fetch the audio asset data to be ready and schedule waves mask generation if needed
+                audioMessagePlayer.preloadAudioMessage(args.conversationId, args.messageId)
+            } catch (_: Throwable) {
```
sbakhtiarov, line 101: Must re-throw the CancellationException

app/src/main/kotlin/com/wire/android/media/audiomessage/AudioMessageViewModel.kt
```diff
@@ -108,8 +112,10 @@ class AudioMessageViewModelImpl @AssistedInject constructor(
                     }
                 }
                 .distinctUntilChanged()
-                .firstOrNull { it != null } // wait for the first non-null value
-                .let { state = state.copy(wavesMask = it) }
+                .firstOrNull { it != null }
```
sbakhtiarov, line 118: I guess this is just `firstOrNull()`?

app/src/main/kotlin/com/wire/android/ui/home/conversations/AudioMessageScopeResolver.kt
```diff
@@ -0,0 +1,7 @@
+package com.wire.android.ui.home.conversations
```
sbakhtiarov, line 19: Copyright header missing here

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/MessageContentAndStatus.kt
```diff
@@ -143,21 +147,82 @@ private fun MessageContent(
...
+            LaunchedEffect(assetStatus) {
+                viewModel.resolveIfNeeded(
+                    transferStatus = assetStatus ?: AssetTransferStatus.NOT_DOWNLOADED,
+                    downloadIfNeeded = true
+                )
+            }
+
             MessageImage(
-                asset = messageContent.asset,
+                asset = null,
```
sbakhtiarov: Should we remove all code handling non-null asset for MessageImage?

app/src/main/kotlin/com/wire/android/ui/common/multipart/MultipartAttachmentUi.kt
```diff
@@ -73,7 +73,7 @@ fun AssetContent.toUiModel(progress: Float?) = MultipartAttachmentUi(
     uuid = this.remoteData.assetId,
     source = AssetSource.ASSET_STORAGE,
     fileName = this.name,
-    localPath = this.localData?.assetDataPath,
```
sbakhtiarov, line 76: I suspect this may break the image UX for local images.
When you send image it is displayed immediately in the conversation, but there is no preview URL available, so local image is used to show.
After this change we may have empty image placeholders displayed until the image is sent, preview generated and preview URL fetched from server.

sbakhtiarov (changes requested):

sbakhtiarov (approved):

PR #4691, opened 2026-04-03
feat: handle call moderation actions [WPB-3185]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-3185

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

User needs to be informed of the actions taken towards them by the moderator/admin during the call.
Managing and storing these events has been added in kalium, along with the use case to observe them.
This PR includes changes needed to display them correctly:
- Added two methods to `ExpiringMap` to be able to add item which expires in a given custom period and to add one without expiration
- Added new `InCallToast` model which currently has two main types: 
    - `ModerationAction` to show all actions done by the moderators/admins
    - `Fullscreen` which handles already existing labels informing about how to open or close the fullscreen
- Added logic of observing and expiring toasts in `OngoingCallViewModel` - toasts are stored by their ID, but currently toasts of the same type share the same ID, thanks to that there is always only one toast of a given type on the screen, and when a new one appears, it overwrites the existing one of the same type
- Created dedicated composable and panel to show these in-call toasts chronologically and animate adding and removing them
- Removed old `DoubleTapToast` and its usages in `FullscreenTile` and `OngoingCallScreen` as it’s now handled by the `InCallToast`

### Dependencies (Optional)

Needs releases with:

- https://github.com/wireapp/kalium/pull/4014

#### Test Coverage (Optional)

#### How to Test

Join a call, unmute and let moderator/admin to mute you remotely. 
Open and close fullscreen.

### Attachments (Optional)

https://github.com/user-attachments/assets/84df684a-854f-41e9-adaa-21626d5f1377

----

----

----
##### References


app/src/main/kotlin/com/wire/android/ui/calling/ongoing/toast/InCallToastPanel.kt
```diff
@@ -0,0 +1,117 @@
...
+
+            key(key) { // Use the toast key as the Compose key to maintain identity across recompositions
+                val visibilityState = remember {
+                    MutableTransitionState(isPreview).apply { // For preview, start visible, otherwise start hidden
+                        targetState = true // Target 'true' immediately to guarantee 'Enter' animation plays for new items
+                    }
+                }
+                // Now, sync with the real 'isPresent' logic. This only changes the target to 'false' when the item is actually removed
+                visibilityState.targetState = isPresent
+                // Animate visibility based on the presence of the item in the source set. This controls the enter/exit animations
```
ohassine: Feels a bit over-commented, maybe we can keep only the ones that explain *why* and drop the obvious ones to make it easier to read.

saleniuk: Okay, I removed most of the commits, kept only the most relevant ones

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/toast/InCallToast.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.wire.android.R
+import com.wire.android.ui.common.colorsScheme
+import com.wire.android.ui.common.dimensions
+import com.wire.android.ui.common.typography
+import com.wire.android.ui.theme.WireTheme
+import com.wire.android.util.ui.PreviewMultipleThemes
+import com.wire.android.util.ui.stringWithStyledArgs
+
+sealed class InCallToast(open val time: Long) {
+    abstract val id: String
```
yamilmedina: Question: I was wondering if we could have a stricter value for this, to avoid hardcoded strings. 🤔 

Maybe something like `ToastKey`, which can be sealed class or interface, or maybe an enum would do it.

```kotlin
sealed class InCallToast(open val time: Long) {
    abstract val key: ToastKey // this insteed of  String
    
    // and the InCallToast classes could be:
    data class Fullscreen(override val time: Long, val type: Type) : InCallToast(time) {
        override val key = ToastKey.Fullscreen(type)

        enum class Type { DoubleTapToOpen, DoubleTapToClose }
    }

    data class ModerationAction(
        override val time: Long,
        val actionId: String, // others...
    ) : InCallToast(time) {
        override val key = ToastKey.ModerationAction(actionId)  // or maybe this is also an enum instead of 

        enum class Type { Muted }
    }
}

sealed interface ToastKey {
    data class Fullscreen(val type: Fullscreen.Type) : ToastKey
    data class ModerationAction(val actionId: String) : ToastKey
}


```

yamilmedina (approved):
Amazing to have now this unified toast component. 🚀 
Just left a small suggestion/question if we might be able to avoid the hardcoded IDs for the Toast Keys/ IDs.

ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/calling/ongoing/toast/InCallToast.kt
saleniuk: @yamilmedina thanks, nice suggestion!
I played with your idea a bit and I think I found a simple way by just making `Type`s to extend `sealed interface Key`, this way we don't need to have duplicated declarations of `FullScreen` and `ModerationAction`, and in future, if some new type needs a more complex key, it can be just extended inside this type 😃

PR #4737, opened 2026-04-17
feat: Search for Apps (with bots coexisting) #WPB-20357

alexandreferris:
https://wearezeta.atlassian.net/browse/WPB-20357

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
    - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
    - [X] contains a reference JIRA issue number like `SQPIT-764`
    - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
    - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

There was no way to search for Apps and not the full handling of Apps available feature flag

### Causes (Optional)

Not fully handled/implemented yet

### Dependencies (Optional)

Needs releases with:

- [X] https://github.com/wireapp/kalium/pull/4025 
- [X] https://github.com/wireapp/kalium/pull/4034 
- [X] https://github.com/wireapp/kalium/pull/4041 
- [ ] https://github.com/wireapp/kalium/pull/4050

#### Test Coverage (Optional)

- [X] I have added automated test to this contribution

#### How to Test

Testing based on Default Protocol / Supported Protocols / Apps feature flag and Conversation state (creation / created) can be evaluated based on the matrix table available [here](https://wearezeta.atlassian.net/wiki/spaces/integrations/pages/2716401836/App+Flow+with+Feature+flag+and+Protocols+for+Clients+Test+cases+-+Bots+and+New+Apps)

### Notes (Optional)

This is currently a DRAFT PR and depends on the merge of Kalium PR's to its epic branch first.

PR #4738, opened 2026-04-21
feat: Service Details Screen (WPB-24495)

alexandreferris:
https://wearezeta.atlassian.net/browse/WPB-24495

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Service details screen were still with old design

### Causes (Optional)

Not implemented new design

### Solutions

- Re-design Service Details Screen and Service Not Found Screen
- Add new string for Author of Service

### Dependencies (Optional)

Needs releases with:

- [ ] https://github.com/wireapp/kalium/pull/4054
- [ ] #4737 

#### How to Test

- Have a team with Apps / Services enabled
- From a group conversation, go into details and open an App / Service details screen
- New design should be shown

### Note
Currently targeting another feat branch, but as soon as its merged we can point to `develop` as usual

### Attachments (Optional)

Note: bottom button at the first before/after for no service details found was shown only for testing (as I was forcing the no data found)

| Before | After |
| unavailable_before  | unavailable_after  |
|  | info_new |


app/src/main/kotlin/com/wire/android/ui/userprofile/service/ServiceDetailsScreen.kt
```diff
@@ -130,46 +133,97 @@ private fun ServiceDetailsTopAppBar(
...
+                    horizontalArrangement = Arrangement.spacedBy(dimensions().spacing8x)
+                ) {
+                    Text(
+                        text = serviceDetails.name,
+                        style = MaterialTheme.wireTypography.title02,
+                        color = if (serviceDetails.name.isNotBlank()) {
+                            MaterialTheme.colorScheme.onBackground
+                        } else {
+                            MaterialTheme.wireColorScheme.secondaryText
+                        },
```
saleniuk: Question: do we need this `if`? I think that if the `serviceDetails.name` is blank then there won't be any text anyway so no need to change the color of the text 🤔

app/src/main/kotlin/com/wire/android/ui/userprofile/service/ServiceDetailsScreen.kt
```diff
@@ -130,46 +133,97 @@ private fun ServiceDetailsTopAppBar(
...
+                    Text(
+                        text = stringResource(id = R.string.service_details_created_by, creator),
+                        style = MaterialTheme.wireTypography.body01,
+                        color = if (creator.isNotBlank()) {
+                            MaterialTheme.colorScheme.onBackground
+                        } else {
+                            MaterialTheme.wireColorScheme.secondaryText
+                        },
+                    )
+                }
```
saleniuk: Question: same here with the color, but also should we maybe check if it's not blank before creating the whole `Text`? Otherwise if it's empty then we will end up having just a text "Created by " with empty name at the end

saleniuk (approved):

MohamadJaara (approved):

PR #4756, opened 2026-04-24
feat: make bottom calling controls a draggable bottom sheet [WPB-1057]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-1057

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Call screen needs to have a draggable bottom sheet with calling controls, where list of participants will be placed.

### Solutions

`OngoingCallScreen` using `BottomSheetScaffold` back again. It got reduced and simplified some time ago, when implementing in-call reactions panel and in the meantime we implemented landscape mode with support for tablets and edge-to-edge, so some changes had to be made.
- added `BottomSheetScaffold` and changed so that it works more like on designs and similarly to `ModalBottomSheet` with closing on back and tapping outside and scrim color
- extracted `WireDragHandle` into a dedicated class and made it animated so that it can change state from arrow to line
- `BackHandler`s taken out from the children and moved up in the hierarchy so that it doesn't interfere with the one implemented for the bottom sheet
- `CallingControls` moved to the sheet content 
- sheet "peek height" calculated dynamically depending on the `CallingControls` height, to include security banner if visible for instance
- created a way to preview `OngoingCallScreen` with `SecurityClassificationBanner` even if it uses its own scoped ViewModel by introducing `PreviewSecurityClassificationBannerState`

#### How to Test

Start or join a call and check the layout (currently dragging/opening bottom sheet is disabled until list of participants is implemented).

### Attachments (Optional)

https://github.com/user-attachments/assets/30e1400e-0b08-4a0f-839b-1c7f59beae2c

----

----

----
##### References


sbakhtiarov (approved):

sbakhtiarov:
Is this UI also adapted for tablets?

saleniuk:
> Is this UI also adapted for tablets?

Currently we follow the most recent designs for landscape and big screens, so the sheet takes the whole width (fortunately now it's already possible to set the max width):
image
but thank you very much for your comment, as it made me notice that the calling controls should have a max width so that they don't stretch to the full width on landscape, so I changed that. 🙇

core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireBottomSheetScaffold.kt
```diff
@@ -15,63 +15,205 @@
...
+                    alpha = scrimAlpha,
+                    onDismissRequest = dismissSheet,
+                    modifier = Modifier.fillMaxSize()
+                )
+            }
         )
+        BackHandler(enabled = properties.shouldDismissOnBackPress && scaffoldState.bottomSheetState.currentValue == SheetValue.Expanded) {
+            dismissSheet()
+        }
+    }
```
ohassine: There is backhandler here and in OngoingCallScreen, those may may compete causing an expected behavior.
We could delegate everything to one centralized handler, or disable backhandler on scaffold and have it only in screen

core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireBottomSheetScaffold.kt
```diff
@@ -15,63 +15,205 @@
...
             contentColor = contentColor,
-            content = content
+            content = { paddingValues ->
+                content(paddingValues)
+                Scrim(
+                    color = properties.scrimColor,
+                    alpha = scrimAlpha,
+                    onDismissRequest = dismissSheet,
+                    modifier = Modifier.fillMaxSize()
+                )
```
ohassine: can we render only one top-level Scrim, instead of having one in content and one in topbar

```
Box {
    BottomSheetScaffold(...)

    Scrim(
        modifier = Modifier.fillMaxSize(),
        alpha = scrimAlpha,
        onDismissRequest = dismissSheet
    )
}
```

core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireBottomSheetScaffold.kt
```diff
@@ -15,63 +15,205 @@
...
+private fun Scrim(
+    modifier: Modifier = Modifier,
+    color: Color = Color.Unspecified,
+    alpha: Float = 0f,
+    onDismissRequest: (() -> Unit)? = null,
+) {
+    val closeSheet = stringResource(ComposeUiR.string.close_sheet)
+    Canvas(
+        modifier.let {
+            if (alpha > 0f && onDismissRequest != null) {
```
ohassine, line 153: I think we should use `shouldDismissOnClickOutside` here as it's not used at all
`    if (alpha > 0f && onDismissRequest != null && shouldDismissOnClickOutside) { ... }`

ohassine (changes requested):

core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireBottomSheetScaffold.kt
saleniuk: Ok, although it's not a problem when we keep the order as we want it to be and that's what I did in this PR, because previously there were even more BackHandlers, but they were being created after the WireBottomSheetScaffold so they were intercepting back actions.
The thing to keep in mind for BackHandlers is that they work in the LIFO order, so the innermost one, the one that is added last, gets the action first. If we follow that, then we are good.

But I get your point as it's easier to understand what's the actual order without the need to look at all child composables, so I applied your suggestion. I had to move a few things around though, and we lost the advantage that this one uses the same logic as ModalSheetLayout with parameters `shouldDismissOnBackPress` and `shouldDismissOnClickOutside` as it doesn't make sense since now we handle it outside. I replaced it with `SheetScrimState` to determine whether we want the scrim or not and set the action on click.

core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireBottomSheetScaffold.kt
saleniuk: No, it's not possible, they need to be on top of the content but below the sheet content, so there's no other way and your proposition would draw it on top of the sheet content as well.

core/ui-common/src/main/kotlin/com/wire/android/ui/common/bottomsheet/WireBottomSheetScaffold.kt
saleniuk: With the changes from the first comment about back handler, this is no longer relevant.

ohassine (approved):

PR #4760, opened 2026-04-24
fix: move message asset paths to conversation-scoped cache VM [WPB-25002]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-25002

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
- Single-asset messages could get stuck in loading/downloading when switching conversations quickly.
- Local asset paths were being lost because state was tied to short-lived per-item lifecycle.
- Loading overlays could still appear even when a valid local image source existed.

### Causes (Optional)
- Asset path state was stored per item, so detach/reattach could recreate state with `null` path.
- UI loading decisions relied too much on transfer status instead of source availability.

### Solutions
- Moved message-list asset path handling to a conversation-scoped cache ViewModel (`messageId -> localPath`) and updated image/video/asset rendering to use it.
- Updated image overlay/render logic to prioritize available local sources and prevent unnecessary loading placeholders.
- Added unit tests for the new conversation-scoped asset path flow, including regression coverage for in-progress transfer scenarios.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/MessageContentAndStatus.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/MessageContentAndStatus.kt
@@ -24,7 +24,6 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.LocalInspectionMode
 import androidx.hilt.navigation.compose.hiltViewModel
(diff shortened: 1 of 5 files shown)


MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/item/MessageContentAndStatus.kt
```diff
@@ -333,31 +302,12 @@ private fun MessageContent(
...
-                        AssetLocalPathViewModelImpl.Factory,
-                        >(
-                        args
-                    )
-                }
-            LaunchedEffect(assetStatus) {
-                viewModel.resolveIfNeeded(
+            val messageId = message.header.messageId
+            val localAssetPath = conversationAssetPathsViewModel.localAssetPath(messageId)
+            LaunchedEffect(messageId, assetStatus) {
```
sbakhtiarov: Could we move these 'LaunchEffect' blocks into ViewModel? Like this:

```
    override fun localAssetPath(conversationId: ConversationId, messageId: String, assetStatus: AssetTransferStatus?): String? = localAssetPaths[messageId].also { path ->
        if (path == null && resolvingJobs[messageId]?.isActive != true) {
            resolveIfNeeded(
                messageId = messageId,
                transferStatus = assetStatus ?: AssetTransferStatus.NOT_DOWNLOADED,
                downloadIfNeeded = true,
            )
        }
    }

```

sbakhtiarov (approved):

PR #4773, opened 2026-04-29
feat: add e2ei expiration control in debug tools and shake routing by flavor [WPB-24275]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24275

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- No usable debug UI for short E2EI certificate TTL testing.
- Shake shortcut behavior was not flavor-aware.

### Solutions

- Added E2EI TTL controls in Developer Tools (minutes + enroll), wired to debug API.
- Enforced UI minimum of `6 min` (`360s`).
- Improved section UI consistency and added TTL scope hint for enroll/update.
- Updated shake routing:
  - private + debug-screen-enabled builds -> Debug Settings,
  - other builds -> Log Management.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
@@ -18,13 +18,30 @@
 package com.wire.android.ui.debug
 
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.wrapContentWidth
+import androidx.compose.foundation.background
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.OutlinedTextField
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.unit.dp
 import com.wire.android.BuildConfig
 import com.wire.android.R
 import com.wire.android.di.hiltViewModelScoped
@@ -67,6 +84,8 @@ fun DebugDataOptions(
         onForceUpdateApiVersions = viewModel::forceUpdateApiVersions,
         onDisableEventProcessingChange = viewModel::disableEventProcessing,
         enrollE2EICertificate = viewModel::enrollE2EICertificate,
+        e2eiCertificateExpirationSeconds = viewModel.state.e2eiCertificateExpirationSeconds,
+        onE2EICertificateExpirationChange = viewModel::updateE2EICertificateExpiration,
         handleE2EIEnrollmentResult = viewModel::handleE2EIEnrollmentResult,
         dismissCertificateDialog = viewModel::dismissCertificateDialog,
         checkCrlRevocationList = viewModel::checkCrlRevocationList,
@@ -89,6 +108,8 @@ fun DebugDataOptionsContent(
     onRestartSlowSyncForRecovery: () -> Unit,
     onForceUpdateApiVersions: () -> Unit,
     enrollE2EICertificate: () -> Unit,
+    e2eiCertificateExpirationSeconds: Long,
+    onE2EICertificateExpirationChange: (Long) -> Unit,
     handleE2EIEnrollmentResult: (FinalizeEnrollmentResult) -> Unit,
     dismissCertificateDialog: () -> Unit,
     checkCrlRevocationList: () -> Unit,
@@ -181,9 +202,11 @@ fun DebugDataOptionsContent(
                 trailingIcon = R.drawable.ic_arrow_right,
             )
 
-            if (BuildConfig.DEBUG) {
+            if (BuildConfig.PRIVATE_BUILD && BuildConfig.DEBUG_SCREEN_ENABLED) {
                 GetE2EICertificateSwitch(
-                    enrollE2EI = enrollE2EICertificate
+                    expirationSeconds = e2eiCertificateExpirationSeconds,
+                    onExpirationChange = onE2EICertificateExpirationChange,
+                    enrollE2EI = enrollE2EICertificate,
                 )
 
                 if (state.showCertificate) {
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -235,30 +258,86 @@ fun DebugDataOptionsContent(
 
 @Composable
 private fun GetE2EICertificateSwitch(
+    expirationSeconds: Long,
+    onExpirationChange: (Long) -> Unit,
     enrollE2EI: () -> Unit
 ) {
+    val minExpirationMinutes = 6L
```
ohassine: We have `minExpirationMinutes` here, and _MIN_E2EI_CERTIFICATE_EXPIRATION_SECONDS_ in adifferent class..
can we have one const in a shared place to avoid inconsistency

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -235,30 +258,86 @@ fun DebugDataOptionsContent(
...
+                    modifier = Modifier.width(128.dp),
+                    value = expirationInput,
+                    onValueChange = { input ->
+                        val digitsOnly = input.filter { it.isDigit() }
+                        expirationInput = digitsOnly
+                        digitsOnly.toLongOrNull()?.let { minutes ->
+                            if (minutes >= minExpirationMinutes) {
+                                onExpirationChange(minutes * 60)
+                            }
+                        }
```
ohassine: I think we  should show a feedback here in case < minExpirationMinutes otherwise we will have a silent input rejection

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptionsViewModel.kt
```diff
@@ -70,6 +73,9 @@ interface DebugDataOptionsViewModel {
     fun checkCrlRevocationList() {}
     fun restartSlowSyncForRecovery() {}
     fun enrollE2EICertificate() {}
+    fun updateE2EICertificateExpiration(seconds: Long) {}
+    fun increaseE2EICertificateExpiration() {}
+    fun decreaseE2EICertificateExpiration() {}
```
ohassine: `increaseE2EICertificateExpiration()` and `decreaseE2EICertificateExpiration()` are unused

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -235,30 +258,86 @@ fun DebugDataOptionsContent(
 
 @Composable
 private fun GetE2EICertificateSwitch(
```
ohassine: the name now is misleading, it is no longer Switch.
how about : `E2EICertificateEnrollmentSection` ?

ohassine (approved):

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -234,31 +258,85 @@ fun DebugDataOptionsContent(
...
-                    text = stringResource(R.string.label_get_e2ei_cetificate),
-                    fillMaxWidth = false
+                    text = stringResource(R.string.debug_settings_e2ei_enroll_button),
+                    fillMaxWidth = false,
+                    minSize = MaterialTheme.wireDimensions.buttonMediumMinSize,
+                    minClickableSize = MaterialTheme.wireDimensions.buttonMinClickableSize
                 )
             }
-        )
+            HorizontalDivider(modifier = Modifier.height(1.dp))
```
saleniuk: use `dimensions()`

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -234,31 +258,85 @@ fun DebugDataOptionsContent(
...
+                    .padding(
+                        start = dimensions().spacing8x,
+                        end = dimensions().spacing8x,
+                        top = dimensions().spacing8x,
+                        bottom = dimensions().spacing8x
+                    ),
+                verticalAlignment = Alignment.CenterVertically
+            ) {
+                OutlinedTextField(
+                    modifier = Modifier.width(128.dp),
```
saleniuk: use `dimensions()`

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -234,31 +258,85 @@ fun DebugDataOptionsContent(
...
+                    value = expirationInput,
+                    onValueChange = { input ->
```
saleniuk: You could also use `TextFieldState` instead, so that we use more recent textfields v2 if possible

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -234,31 +258,85 @@ fun DebugDataOptionsContent(
...
+    val expirationMinutes = expirationSeconds / 60
+    var expirationInput by remember { mutableStateOf(expirationMinutes.toString()) }
+    val isInputBelowMinimum = expirationInput.toLongOrNull()?.let { it < minExpirationMinutes } == true
+
+    LaunchedEffect(expirationSeconds) {
+        val minutesFromState = (expirationSeconds / 60).toString()
+        if (expirationInput != minutesFromState) {
+            expirationInput = minutesFromState
+        }
+    }
```
saleniuk: This feels a little bit too complex, it could definitely be simpler 😄 
If you pass `expirationSeconds` state already then why do we need another `expirationInput` state here instead of just using the one that's passed?
Also, it's being reduced so that the input shows only full minutes, so when `expirationSeconds` is 90 seconds, then input will show 1 minute, which is not correct, right?

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -234,31 +258,85 @@ fun DebugDataOptionsContent(
...
+                        top = dimensions().spacing8x,
+                        bottom = dimensions().spacing8x
+                    ),
+                verticalAlignment = Alignment.CenterVertically
+            ) {
+                OutlinedTextField(
+                    modifier = Modifier.width(128.dp),
+                    value = expirationInput,
+                    onValueChange = { input ->
+                        val digitsOnly = input.filter { it.isDigit() }
```
saleniuk: Instead of that, pass `inputTransformation` so that it won't accept anything other than digits, so no need to worry about that and we can assume that entered value always contains only digits, we already have a `InputTransformation.maxLengthDigits(maxLength: Int)` specifically for that

app/src/main/kotlin/com/wire/android/ui/debug/DebugDataOptions.kt
```diff
@@ -234,31 +258,85 @@ fun DebugDataOptionsContent(
...
+                .background(MaterialTheme.wireColorScheme.surface)
+        ) {
+            Row(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(
+                        start = dimensions().spacing8x,
+                        end = dimensions().spacing8x,
+                        top = dimensions().spacing8x,
+                        bottom = dimensions().spacing8x
```
saleniuk: All the same values, can be reduced to a single `dimension().spacing8x`

saleniuk (approved):

PR #4778, opened 2026-05-04
fix: do not show toast when trying to log in again [WPB-25221]

marcoconti83:
https://wearezeta.atlassian.net/browse/WPB-25221

# What's new in this PR?

When starting a nomad profile login flow via intent, if the user was already logged in, the user would see a toast informing them that they can't log in again.

However, the customer using nomad profiles will always start the app via the intent. They should not see an error if they are already logged in with a nomad profile account.


MohamadJaara (approved):
LGTM

PR #4792, opened 2026-05-06
fix: talkBack labels for message avatars and rich text actions [WPB-14771] [WPB-14767] [WPB-14773] [WPB-14812]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-14771

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- TalkBack announced duplicated role text for rich text editing actions, for example “Enable rich text mode button button”.
- Message avatars were not properly exposed as profile-opening buttons when TalkBack was enabled.
- Disabled/unavailable sender avatars could announce “open profile” even when opening the profile was not possible.

### Causes (Optional)

- Some accessibility strings included the word “button” while Compose also announced the `Button` role.
- The avatar click target did not provide a dedicated TalkBack description and button semantics.
- The avatar description appended the profile action label regardless of whether profile redirection was enabled.

### Solutions

- Removed explicit “button” wording from rich text and attachment action content descriptions.
- Added avatar TalkBack semantics with a profile-focused content description and button role when the avatar is clickable.
- Included “open profile” in the avatar description only when profile redirection is enabled.


core/ui-common/src/main/kotlin/com/wire/android/ui/common/avatar/UserProfileAvatar.kt
```diff
@@ -148,11 +148,20 @@ fun UserProfileAvatar(
         legalHoldIndicatorVisible = false
     ),
 ) {
+    val accessibilityModifier = if (contentDescription != null && clickable?.enabled == true) {
+        Modifier.clearAndSetSemantics {
+            this.contentDescription = contentDescription
+            role = Role.Button
+        }
+    } else {
```
saleniuk: I feel like we should always override the content description if it's non-null, not only when it's also clickable, but only add a role if it's clickable:
```suggestion
    val accessibilityModifier = if (contentDescription != null) {
        Modifier.clearAndSetSemantics {
            this.contentDescription = contentDescription
            if (clickable?.enabled == true) {
                role = Role.Button
            }
        }
    } else {
```

What do you think about it?

saleniuk (approved):

yamilmedina (approved):

PR #4808, opened 2026-05-11
fix: Nomad profiles should not see app lock - WPB-25543

marcoconti83:
https://wearezeta.atlassian.net/browse/WPB-25543

# What's new in this PR?

When an account is configured as a nomad profile, that account should not see any app lock prompt, even if the team administrator enforces app lock for the team.

### Testing
I've manually tested this on Fulu on a team with app lock enabled

### DIFF
--- a/app/src/test/kotlin/com/wire/android/feature/ObserveAppLockConfigUseCaseTest.kt
+++ b/app/src/test/kotlin/com/wire/android/feature/ObserveAppLockConfigUseCaseTest.kt
@@ -110,6 +110,42 @@ class ObserveAppLockConfigUseCaseTest {
         }
 
     @Test
+    fun givenValidSessionAndTeamConfigIsNull_whenObservingAppLock_thenSendDisabledStatus() = runTest {
+        val (_, useCase) = Arrangement()
+            .withValidSession()
+            .withTeamAppLockNull()
+            .withAppNonLockedByCurrentUser()
+            .arrange()
+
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/wire/android/feature/ObserveAppLockConfigUseCase.kt
```diff
@@ -39,6 +39,11 @@ class ObserveAppLockConfigUseCase @Inject constructor(
         coreLogic.getGlobalScope().session.currentSessionFlow().collectLatest { sessionResult ->
             when {
                 sessionResult is CurrentSessionResult.Success && sessionResult.accountInfo.isValid() -> {
+                    // Nomad profiles must never be subject to app lock, even when a team enforces it.
+                    if (coreLogic.getGlobalScope().doesValidNomadAccountExist()) {
+                        send(AppLockConfig.Disabled(DEFAULT_APP_LOCK_TIMEOUT))
+                        return@collectLatest
+                    }
```
MohamadJaara: would prefer to do this in kalium lvl where the use case for the team flag app lock will not emit enforced if the user have a nomad URL, doing this here is another DB call when the app starts that we do not need 

my suggesiton will mean that the user will not add app lock in the first place

marcoconti83: On it

PR #4819, opened 2026-05-13
feat: UI for promoting next user to admin role (WPB-25278)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25278

https://wearezeta.atlassian.net/browse/WPB-25278
----

# What's new in this PR?

New dialog for leave conversation options and screen for selecting next admin user.

[Screen_recording_20260513_090643.webm](https://github.com/user-attachments/assets/e4985dda-51f5-4ef7-9b42-496a3aef6e59)

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/promoteadmin/PromoteAdminScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/promoteadmin/PromoteAdminScreen.kt
@@ -1,0 +1,200 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.android.ui.home.conversations.promoteadmin
+
+import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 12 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversations/promoteadmin/PromoteAdminScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+                SearchBarInput(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(
+                            horizontal = dimensions().spacing16x,
+                            vertical = dimensions().spacing8x,
+                        ),
+                    placeholderText = stringResource(R.string.promote_admin_search_placeholder),
+                    leadingIcon = {
+                        IconButton(onClick = {}) {
```
saleniuk: I think it may still have the ripple effect when user clicks on that so it can confuse the user as nothing happens after clicking - do we need the `IconButton` here without the click action? You can try to add `enabled = false` but it may affect the color or alpha of the icon. 🤔 
The safest option would be to wrap it with `Box` that has similar size to the `IconButton`, in that case we usually do like that (for instance in `SearchTopBar`): 
```
 Box(
    contentAlignment = Alignment.Center,
    modifier = Modifier.size(dimensions().buttonCircleMinSize)
) { ... }
```

app/src/main/kotlin/com/wire/android/ui/home/conversations/promoteadmin/PromoteAdminViewModel.kt
```diff
@@ -0,0 +1,94 @@
...
+data class PromoteAdminState(
+    val searchQuery: String = "",
+    val allMembers: List<PromoteAdminMemberItem> = emptyList(),
+    val selectedUserId: UserId? = null,
+) {
+    val filteredMembers: List<PromoteAdminMemberItem>
+        get() = if (searchQuery.isBlank()) {
+            allMembers
+        } else {
+            allMembers.filter { it.name.contains(searchQuery, ignoreCase = true) }
```
ohassine: should we also consider handle ?
```
allMembers.filter {
    it.name.contains(searchQuery, true) ||
    it.handle.contains(searchQuery, true)
}
```

ohassine (approved):

Garzas (approved):

PR #4837, opened 2026-05-19
fix: improve keyboard navigation in search conversations [WPB-14778]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-14778

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Fixed issues:
- Search fields could receive keyboard input, but search results and related controls were not reliably reachable or actionable with keyboard navigation.
- The home search field could trap focus or make surrounding controls, such as profile, clear search, and New/FAB, difficult to reach in the expected TAB order.
- The New Conversation button shown for empty search results could not be reached reliably with keyboard navigation.
- The clear `x` button in the search field was skipped by keyboard focus in some search states.
- Some controls were reachable only backwards with `SHIFT+TAB`, but not forwards with `TAB`.

### Causes

The affected Compose search components relied mostly on default focus traversal. This was not enough for dynamic UI states such as inactive/active search, conditionally visible clear actions, empty search results, and FAB visibility.

Some focus behavior also needed to be attached directly to the actual text input or focusable control so keyboard traversal would not skip intermediate actions.

### Solutions

- Added explicit focus requesters and `focusProperties` for the Conversations home top bar, profile action, FAB/New button, search field, clear search button, and search back button.
- Updated search handling so keyboard focus opens the search field, shows the keyboard, supports backwards focus to the back button, and allows TAB to continue into the conversation/search results list.
- Updated empty search results so the New Conversation action is reachable from the search field with keyboard navigation.
- Updated search clear action focus handling so TAB can move from the search field to the `x` button when text is present.
- Added reusable `onEscapeOrBackKey` modifier used by search UI to close keyboard-driven search state consistently.


app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerKeyboardNavigationState.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+private fun handleEnterKeyInComposerInput(
+    event: KeyEvent,
+    messageTextState: TextFieldState,
+    canSendMessage: Boolean,
+    onSendButtonClicked: () -> Unit,
+) {
+    if (event.isShiftPressed) {
```
sbakhtiarov: We should probably disable "send on enter" when input field is in "expanded" mode.

Garzas: Removed changes from MessageComposer, because it will need to be a separate refactor

app/src/main/kotlin/com/wire/android/ui/home/conversationslist/ConversationsScreenContent.kt
```diff
@@ -194,7 +196,10 @@ fun ConversationsScreenContent(
                     }
                 )
                 // when there is no conversation in any folder
-                searchBarState.isSearchActive -> SearchConversationsEmptyContent(onNewConversationClicked = onNewConversationClicked)
+                searchBarState.isSearchActive -> SearchConversationsEmptyContent(
+                    onNewConversationClicked = onNewConversationClicked,
+                    newConversationFocusRequester = emptySearchResultFocusRequester
```
saleniuk: Do we need to explicitly pass it as a parameter to the `SearchConversationsEmptyContent`? Inside that composable it just adds it to the modifier so we could do that here directly:
```
SearchConversationsEmptyContent(
    onNewConversationClicked = onNewConversationClicked,
    modifier = emptySearchResultFocusRequester?.let { Modifier.focusRequester(it) } ?: Modifier,
)
```

app/src/main/kotlin/com/wire/android/ui/home/HomeTopBar.kt
```diff
@@ -52,9 +55,13 @@ fun HomeTopBar(
     onHamburgerMenuClick: () -> Unit,
     onNavigateToSelfUserProfile: () -> Unit,
     onOpenConversationFilter: () -> Unit,
+    modifier: Modifier = Modifier,
+    searchFocusRequester: FocusRequester? = null,
+    fabFocusRequester: FocusRequester? = null,
```
saleniuk: Would it be possible to have only single parameter and make `HomeTopBar` not having the logic of deciding itself which one should be used next? For instance if we use this top bar in another scenario where we have completely different elements on a screen then we will need to add more focusRequesters. 🤔 
Instead, we could pass single `nextFocusRequester` and move the logic of deciding which one it should be to the parent, because now it feels like some part of the logic which decides which one to use next is here inside `HomeTopBar` and some is in the parent for instance when passing parameters:
`fabFocusRequester = if (currentNavigationItem.fab != null) fabFocusRequester else null,`

saleniuk (approved):

sbakhtiarov (approved):

PR #4853, opened 2026-05-21
refactor: improve conversation list rendering performance [WPB-26102]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-26102

Refactors conversation list item rendering to avoid doing UI-only and expensive work during paged item mapping. Search highlighting, legal-hold visibility, and currently playing audio state are now applied closer to rendering, reducing unnecessary remapping when those UI states change.

Also defers last-message markdown preview parsing to the UI layer and adds a lightweight markdown preview parser so conversation subtitles avoid full markdown parsing while preserving formatted previews.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversationslist/common/ConversationItemFactory.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversationslist/common/ConversationItemFactory.kt
@@ -50,6 +50,7 @@ import com.wire.android.ui.common.rowitem.RowItemTemplate
 import com.wire.android.ui.home.conversations.info.ConversationAvatar
 import com.wire.android.ui.home.conversations.model.MessageBody
 import com.wire.android.ui.home.conversations.model.UILastMessageContent
+import com.wire.android.ui.home.conversationslist.showLegalHoldIndicator
 import com.wire.android.ui.home.conversationslist.model.BadgeEventType
 import com.wire.android.ui.home.conversationslist.model.BlockingState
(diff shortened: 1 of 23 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversationslist/common/ConversationItemFactory.kt
```diff
@@ -168,6 +178,9 @@ private fun GeneralConversationItem(
     modifier: Modifier = Modifier,
     selectOnRadioGroup: () -> Unit = {},
     subTitle: @Composable () -> Unit = {},
+    searchQuery: String = conversation.searchQuery,
+    showLegalHoldIndicator: Boolean = conversation.showLegalHoldIndicator,
+    playingAudio: PlayingAudioInConversation? = conversation.playingAudio,
```
saleniuk: I think that `conversationItem.searchQuery` and `conversationItem.playingAudio` are not being set anywhere anymore, apart from some mocks and previews, so to avoid the confusion as to whether these fields in the `ConversationItem` are filled in and correct or not, maybe we should decide to only provide them externally like here and remove these fields completely from `ConversationItem`?
About legal hold field, maybe we could simplify it to make `ConversationItem` to have `LegalHoldStatus` field instead and move the logic of determining `showLegalHoldIndicator` completely to `ConversationItemFactory` as now it's scattered around and some part is determined in `ConversationMapper`, then in `ConversationListViewModel` and finally in `ConversationItemFactory` 🤔

saleniuk (approved):

Garzas (approved):

PR #4866, opened 2026-05-25
fix: improve keyboard focus order in search screens [WPB-14781]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-14781

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Keyboard focus order was inconsistent on search-based screens.
- `Search conversations`, `Search for people`, and related search fields could trap focus or skip visible controls.
- Inactive search fields opened the keyboard immediately, which could hide bottom actions such as FAB, `New Group`, or `Continue`.
- Some actions after search results, including FAB, first contact/conversation rows, and bottom buttons, were not reachable in the expected sequential TAB order.
- Screens with tabs could move focus unpredictably between search, tabs, list content, and bottom actions.

### Causes

The affected Compose screens relied partly on default focus traversal. That was not reliable with dynamic layouts such as collapsing top bars, inactive/active search states, animated bottom actions, tabs, lazy lists, and hidden keyboard-dependent UI.

Some focus targets existed in the Compose tree but did not map to visible actionable controls, which caused “empty” TAB steps.

### Solutions

- Added explicit `FocusRequester` and `focusProperties` wiring for search fields, navigation buttons, FAB, first list items, tabs, and bottom actions.
- Updated inactive `SearchTopBar` behavior so keyboard focus does not immediately open the keyboard; search activates on click, Enter/Space, or printable keyboard input.
- Added focus routing from Home search to FAB/list and from FAB directly to the first conversation row.
- Added focus routing from people search to the first visible contact and then to `New Group`/`Continue`.
- Restored tab focus handling for screens with People/Services tabs.
- Kept visible focus indicators by using the existing text field interaction styling instead of custom borders.

https://github.com/user-attachments/assets/2d87af07-df20-4b29-80a1-ffbe7b3188b3


app/src/main/kotlin/com/wire/android/ui/home/conversations/search/SearchAllPeopleScreen.kt
```diff
@@ -261,7 +296,8 @@ private fun LazyListScope.internalSearchResults(
                     isSelected = isSelected,
                     onCheckClickable = onCheckClickable,
                     actionType = actionType,
-                    clickable = remember { Clickable(onClickDescription = clickDescription) { onOpenUserProfile(contact) } }
+                    clickable = remember { Clickable(onClickDescription = clickDescription) { onOpenUserProfile(contact) } },
```
saleniuk: Could you please add `contact` as a key to the `remember`? Otherwise, currently it's possible for the `remember` ot keep the old `contact` data when it changes 🤔

app/src/main/kotlin/com/wire/android/ui/home/conversations/search/SearchAllPeopleScreen.kt
```diff
@@ -339,6 +378,18 @@ private fun LazyListScope.externalSearchResults(
...
+private fun Contact.firstContactFocusModifier(
+    firstContactId: String?,
+    focusRequester: FocusRequester?,
+): Modifier = if (id == firstContactId && focusRequester != null) {
+    Modifier.focusRequester(focusRequester)
+} else {
+    Modifier
+}
+
+private val androidx.compose.ui.input.key.KeyEvent.isForwardTabKey: Boolean
```
saleniuk: Nitpick but can we just import `androidx.compose.ui.input.key.KeyEvent`? 😄

core/ui-common/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchTopBar.kt
```diff
@@ -327,16 +386,50 @@ private fun InactiveSearchBarInput(
...
+    keyboardFocusInteraction: MutableState<FocusInteraction.Focus?>,
+    interactionSource: MutableInteractionSource,
+) {
+    LaunchedEffect(activateSearchOnFocus, isSearchBarFocused) {
+        val interaction = keyboardFocusInteraction.value
+        when {
+            !activateSearchOnFocus && isSearchBarFocused && interaction == null -> {
+                FocusInteraction.Focus().also {
+                    keyboardFocusInteraction.value = it
+                    interactionSource.emit(it)
```
saleniuk: I think `interactionSource` should be wrapped in `rememberUpdatedState`, otherwisse `LaunchedEffect` will keep the initial one so if it changes then it will try to emit into outdated element 🤔

core/ui-common/src/main/kotlin/com/wire/android/ui/common/topappbar/search/SearchTopBar.kt
```diff
@@ -327,16 +386,50 @@ private fun InactiveSearchBarInput(
...
+                interactionSource.emit(FocusInteraction.Unfocus(interaction))
+                keyboardFocusInteraction.value = null
+            }
+        }
+    }
+
+    DisposableEffect(Unit) {
+        onDispose {
+            keyboardFocusInteraction.value?.let {
+                interactionSource.tryEmit(FocusInteraction.Unfocus(it))
```
saleniuk: Is it optional or required to unfocus here? If it needs to be executed then we can't use `tryEmit` as it can skip in some scenarios

saleniuk (approved):

sbakhtiarov (approved):

PR #4869, opened 2026-05-26
feat: add forward option when sharing assets [WPB-26687]

MohamadJaara:

Adds separate “Forward in Wire” and external-share flows for logs and shared assets.

This PR builds on zipped-log sharing and extends sharing UX so users can choose whether to send content inside Wire or share it through external apps. It also hardens handling of Wire-owned FileProvider URIs, with separate behavior for Android 15/API 35+ and older Android versions.

Changes

- Add separate sharing actions for:
  - forwarding logs/files in Wire
  - sharing logs/files externally
- Add share-options bottom sheets/menu entries for log and asset sharing.
- Route Wire log sharing through `ImportMediaScreen` using shared file URIs.
- Introduce `ImportMediaNavArgs` for passing internal shared media URIs through navigation.
- Update message, conversation media, and media gallery asset sharing flows to support both Wire and external destinations.
- Keep external sharing behind the Android chooser and exclude Wire’s own share targets.
- Add Android 15/API 35+ trusted-caller handling for Wire FileProvider share intents.
- Reject public share intents containing Wire FileProvider URIs on older Android versions where trusted caller identity is unavailable.
- Restrict accepted internal Wire share URIs to the shared FileProvider root.
- Rework log archive sharing helpers so zipped logs can be shared either externally or imported back into Wire.
- Add/update tests for internal/public share URI validation, trusted vs untrusted Wire provider shares, and media sharing menu behavior.

### DIFF
--- a/app/src/androidTest/kotlin/com/wire/android/ui/debug/DebugScreenComposeTest.kt
+++ b/app/src/androidTest/kotlin/com/wire/android/ui/debug/DebugScreenComposeTest.kt
@@ -43,6 +43,7 @@ class DebugScreenComposeTest {
                     onFlushLogs = { CompletableDeferred(Unit) },
                     debugDataOptionsContent = {},
                     dangerOptionsContent = {},
+                    onShareLogsViaWire = {},
                 )
             }
         }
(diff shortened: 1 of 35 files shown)


MohamadJaara:
closed for smaller PRs

PR #4896, opened 2026-06-01
refactor: migrate scoped message view models to metro [WPB-25946]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-25946

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- Migrates scoped message and composer ViewModels from Hilt assisted factories to Metro-backed Resaca creation.
- Adds `ScopedMessageViewModelFactory` and graph accessors while preserving scoped lifecycle behavior.
- Keeps preview/Espresso fallback behavior in the shared scoped ViewModel helper.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/di/ViewModelScoped.kt
+++ b/app/src/main/kotlin/com/wire/android/di/ViewModelScoped.kt
@@ -19,9 +19,13 @@ package com.wire.android.di
 
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.platform.LocalInspectionMode
+import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
+import com.wire.android.di.metro.LocalMetroViewModelGraph
+import com.wire.android.di.metro.MetroViewModelGraph
 import com.sebaslogen.resaca.KeyInScopeResolver
 import com.sebaslogen.resaca.hilt.hiltViewModelScoped as resacaHiltViewModelScoped
+import com.sebaslogen.resaca.viewModelScoped as resacaViewModelScoped
 import kotlin.time.Duration
 
 /**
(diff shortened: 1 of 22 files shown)


app/src/main/kotlin/com/wire/android/di/ViewModelScoped.kt
```diff
@@ -85,12 +87,11 @@ inline fun <reified T, reified S, reified R : ScopedArgs, reified F : AssistedVi
  * [ViewModel] needs to implement an interface annotated with [ViewModelScopedPreview] and with default
  * implementations.
  */
-@Suppress("BOUNDS_NOT_ALLOWED_IF_BOUNDED_BY_TYPE_PARAMETER")
 @Composable
-inline fun <reified T, reified S> wireViewModelScoped(): S where T : ViewModel, T : S = when {
+inline fun <reified T, reified S> wireViewModelScoped(): S where T : ViewModel = when {
```
saleniuk: Is this change necessary? I believe having `T : S` is actually useful to make sure that we always use proper ViewModel interfaces and their implementations

Garzas: Yeah, I agree with the intention. I tried restoring T : S, but with the current helper shape Kotlin does not allow T : ViewModel, T : S because T would be bounded both by a concrete class and another type parameter. We still need T : ViewModel because Resaca/Android ViewModel APIs require the concrete VM type there.

So this change is mostly a Kotlin type-system limitation from the Metro/Resaca bridge, not a desired weakening of the contract. The call sites still pass ImplementationViewModel, InterfaceViewModel, but the T implements S check can’t be expressed in this generic helper without changing the API shape more deeply.

saleniuk (approved):

MohamadJaara (approved):

PR #4902, opened 2026-06-01
feat: add Compose Stability Analyzer to CI pipeline [WPB-8645]

MohamadJaara:

add Compose Stability Analyzer to CI pipeline


saleniuk (approved):

Garzas:
One thing I’d change before merging:

`stabilityCheck` is now a blocking quality gate in `.github/workflows/build-unified.yml`, but the plugin is left with its default validation behavior. By default, Compose Stability Analyzer fails not only on regressions, but also on non-regressive baseline changes such as newly added or removed composables (`+` / `-`). That means ordinary UI work will have to update committed `.stability` baselines even when stability did not get worse, which is likely to make this CI check noisy and hurt DevEx.

If the goal is to catch stability regressions, please configure the plugin with `ignoreNonRegressiveChanges.set(true)` and keep the gate focused on actual regressions.

Docs: https://skydoves.github.io/compose-stability-analyzer/gradle-plugin/stability-validation/

Garzas (approved):

PR #4903, opened 2026-06-01
fix: show invalid credentials error on login screens [WPB-14814]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-14814

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- Shows a persistent invalid credentials error on the login screen after the dialog is dismissed
- Applies the same error marking to both old and new email/password login flows
- Keeps the message generic to avoid revealing whether the email/username or password was incorrect
- Clears the inline error once the user edits the login input
- Adds ViewModel coverage for preserving and clearing the error state


app/src/main/kotlin/com/wire/android/ui/authentication/login/email/LoginEmailScreen.kt
```diff
@@ -183,6 +183,17 @@ private fun LoginEmailContent(
                     .padding(bottom = MaterialTheme.wireDimensions.spacing16x),
                 passwordState = passwordTextState,
             )
```
saleniuk, line 191: We should also pass error state to these two inputs so that they are highlighted in red, to unify the behavior with all other error types that we have

app/src/main/kotlin/com/wire/android/ui/authentication/login/email/LoginEmailScreen.kt
```diff
@@ -232,6 +250,7 @@ private fun LoginEmailContent(
 private fun UserIdentifierInput(
     userIdentifierState: TextFieldState,
     error: String?,
+    isError: Boolean = false,
```
saleniuk: We now have two separate error-related params, maybe instead just pass `WireTextFieldState` directly?
so here:
```
private fun UserIdentifierInput(
    userIdentifierState: TextFieldState,
    state: WireTextFieldState,
)
```
and then in parent when calling it:
```
UserIdentifierInput(
    modifier = Modifier
        .fillMaxWidth()
        .padding(bottom = MaterialTheme.wireDimensions.spacing16x),
    userIdentifierState = userIdentifierTextState,
    state = when {
        !loginEmailState.userIdentifierEnabled -> WireTextFieldState.Disabled
        loginEmailState.flowState is LoginState.Error.TextFieldError.InvalidValue -> WireTextFieldState.Error(stringResource(R.string.login_error_invalid_user_identifier))
        loginEmailState.showInvalidCredentialsError -> WireTextFieldState.Error(null) // error message in this case is under the password input
    }
)
```

saleniuk (approved):

sbakhtiarov (approved):

PR #4908, opened 2026-06-02
feat: show offline indicator in conversation (WPB-25293)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25293

https://wearezeta.atlassian.net/browse/WPB-25293
----

# What's new in this PR?

Show "You are offline" message in conversation when there is no Internet connection.

Screenshot_20260602_110718


app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/OfflineMessageIndicator.kt
```diff
@@ -0,0 +1,73 @@
...
+internal fun PagingData<UIMessage>.withOfflineIndicator(
+    conversationId: ConversationId,
+    isOffline: Boolean,
+): PagingData<UIMessage> {
+    if (!isOffline) return this
+
+    val offlineMessage = offlineMessage(conversationId)
+    return insertSeparators { before, after ->
+        when {
+            before == null && after == null -> offlineMessage
```
saleniuk, line 45: Do you really want to add "offline" system message when the list is empty? That can disrupt some of our "empty" states if they check whether the list of elements is empty or not

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/OfflineMessageIndicator.kt
```diff
@@ -0,0 +1,73 @@
...
+            before == null && after?.isPending == false -> offlineMessage
```
saleniuk, line 46: Also, do we want the "offline" system message to be visible if there are no "pending" messages? 🤔

sbakhtiarov: According to design it is visible even if there are no pending messages. I guess as an extra indication of "Offline" mode right next to the message input field.

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/OfflineMessageIndicator.kt
sbakhtiarov: I will check this with design. Currently I do not see any impact. Empty state for messages is just empty screen.

sbakhtiarov: Yes, we should always show it, even if there are no messages.

saleniuk (approved):

Garzas (approved):

PR #4932, opened 2026-06-04
fix: metro graph handoff after logout [WPB-26140] [WPB-26141]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-26140

## Summary

Fixes the logout/login handoff after the Hilt to Metro migration.

## What was happening

On logout, the app clears the current user while Compose may still be drawing the old logged-in screen. That old screen still needs the session ViewModel graph for a short moment.

Keeping that graph around fixed the logout crash, but the login screen could briefly reuse the old user-scoped ViewModel key.

Also logging in with a second account was not working
## Fix

- Keep the last session graph alive briefly during logout.
- Release it once the app reaches the login/auth flow.
- Make auth ViewModels always use the auth scope, even during that short transition.

## Testing

Manual verification


app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -305,6 +314,14 @@ class WireActivity : BaseActivity() {
                         ?.destination
                         ?.route
                         ?.getBaseRoute()
+                    val effectiveBaseRoute = currentBaseRoute ?: startDestination.baseRoute
+                    LaunchedEffect(currentUserId, effectiveBaseRoute) {
+                        if (currentUserId == null && effectiveBaseRoute in authenticationGraphRoutes) {
+                            delay(SESSION_GRAPH_RELEASE_DELAY_MILLIS)
+                            retainedSessionViewModelGraph = null
```
ohassine, line 321: I'm a bit concerned about relying on a fixed 500ms delay, as it may not be sufficient on slower devices
can we rely on completion of the logout flow ?

PR #4953, opened 2026-06-10
fix: metro session runtime stability [WPB-26140] [WPB-26134]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-26140

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR fixes runtime regressions introduced while moving ViewModel creation to Metro-backed graphs.

The main issue was that some session-backed UI could keep using stale or missing session graph state during account transitions, logout, removed-client events, and authentication edge cases. That caused blank screens, wrong session reuse, or crashes when the current account disappeared while UI was still composing session-scoped ViewModels.

**What changed**
- Preserves session transition state while logout/account removal is in progress.
- Handles removed-session navigation more defensively.
- Keeps the login password flow on the no-session graph where appropriate.
- Refactors `WireActivity` session graph resolution into smaller helpers.
- Guards session-backed UI from creating current-session ViewModels when there is no valid current session.
- Updates validation/stability changes needed for the Metro session flow.

**Why**
Metro does not magically mirror Hilt’s activity/session scoping unless the graph lifecycle is modeled explicitly. This PR tightens that lifecycle so session-scoped ViewModels are recreated or dropped consistently when the active account changes.

**Side Note**
This is a stabilization step for the transitional Hilt-to-Metro phase. The long-term target is to simplify this further once all session-scoped ViewModels are owned by an explicit SessionGraph(userId) and the remaining legacy current-session assumptions are removed.

**Manual QA focus**
- App startup with existing accounts.
- Switching between two logged-in accounts.
- Add new account flow.
- Too many devices flow: open, cancel/back, remove device.
- Logout with one account and with multiple accounts.
- Removed-client event with one account and with multiple accounts.
- Login after returning from account/session edge cases.
- Conversation list and open conversation after account transitions.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
+++ b/app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
@@ -29,6 +29,7 @@ import com.ramcosta.composedestinations.spec.DestinationSpec
 import com.ramcosta.composedestinations.spec.Direction
 import com.ramcosta.composedestinations.spec.NavGraphSpec
 import com.ramcosta.composedestinations.spec.Route
+import com.ramcosta.composedestinations.utils.destination
 import com.ramcosta.composedestinations.utils.findDestination
 import com.ramcosta.composedestinations.utils.navGraph
 import com.ramcosta.composedestinations.utils.route
@@ -40,8 +41,8 @@ import com.wire.android.util.CustomTabsHelper
 @Suppress("CyclomaticComplexMethod")
 internal fun NavController.navigateToItem(command: NavigationCommand) {
 
-    fun firstDestination() = currentBackStack.value.firstOrNull { it.route() is DestinationSpec }
-    fun lastDestination() = currentBackStack.value.lastOrNull { it.route() is DestinationSpec }
+    fun firstDestination() = currentBackStack.value.firstOrNull { it.safeRoute() is DestinationSpec }
+    fun lastDestination() = currentBackStack.value.lastOrNull { it.safeRoute() is DestinationSpec }
     fun lastNestedGraph() = lastDestination()?.takeIf { it.navGraph() != navGraph }?.navGraph()
     fun firstDestinationWithRoute(route: String) =
         currentBackStack.value.firstOrNull { it.destination.route?.getBaseRoute() == route.getBaseRoute() }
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/com/wire/android/navigation/NavigationUtils.kt
```diff
@@ -133,6 +136,10 @@ fun Direction.handleNavigation(context: Context, handleOtherDirection: (Directio
 }
 
 @SuppressLint("RestrictedApi")
-fun NavController.startDestination() = currentBackStack.value.firstOrNull { it.route() is DestinationSpec }
+fun NavController.startDestination() = currentBackStack.value.firstOrNull { it.safeRoute() is DestinationSpec }
+
+internal fun NavBackStackEntry.safeRoute(): Route? = runCatching { route() }.getOrNull()
```
MohamadJaara, line 141: hmmm why do we need this saferoute i would rather crash to know if there is an issue with a route rather than silently fail.

Garzas: Agree in principle. This is not meant as a final route-validation strategy or to hide broken destinations.

This was added as a transitional guard because during logout/session graph transitions currentBackStack can contain entries before DestinationsNavHost can resolve route(), and that was crashing while we were only trying to scan for the first app destination.

I’ll add a TODO / follow-up to remove this once the session graph/navigation transition is cleaned up. If you prefer, I can also narrow the helper name/comment so it’s clear this is only for transition-time back stack scanning, not general navigation.

MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/WireActivityViewModel.kt
```diff
@@ -659,6 +729,49 @@ class WireActivityViewModel @Inject constructor(
...
+            var resolvedUserId = currentSessionUserId()
+            repeat(retries) { attempt ->
+                if (resolvedUserId != null) return@repeat
+                delay(CURRENT_SESSION_RESOLUTION_RETRY_DELAY_MS)
+                resolvedUserId = currentSessionUserId()
+                if (resolvedUserId != null) {
+                    appLogger.i("Resolved current session after retry ${attempt + 1}")
+                }
+            }
+            resolvedUserId
```
saleniuk: I know that it works with `repeat`, but `withTimeoutOrNull` would look better 😄 
Something like:
```
withTimeoutOrNull(CURRENT_SESSION_RESOLUTION_TIMEOUT_MS) {
    coreLogic.value.getGlobalScope().session.currentSessionFlow()
        .filterIsInstance(CurrentSessionResult.Success::class)
        .filter { it.accountInfo.isValid() }
        .firstOrNull()
}
```

saleniuk (approved):

PR #4960, opened 2026-06-11
refactor: metro session graph final [WPB-26295]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-26295

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR continues the Metro session graph migration and stabilizes runtime behavior around account/session transitions.

### Metro session graph refactor

- Introduces an explicit session-backed authentication boundary for screens that still need an existing session during auth/device flows.
- Moves session-dependent ViewModel creation toward `AppSessionViewModelGraph(userId)` instead of relying on implicit current-session resolution.
- Updates auth/login/device/profile/settings/calling/home ViewModel factories and callsites to pass session context more explicitly.
- Keeps no-session auth flows on the root/auth graph, while session-backed screens resolve a session graph only when a valid `UserId` is available.
- Updates tests for the changed current-session/session-transition behavior.

### Runtime stability fixes

- Keeps session graphs lifecycle-retained from `WireActivity`, so recomposition/navigation does not recreate session-scoped graph objects too aggressively.
- Moves `WireSessionImageLoader` and image loading bindings from root graph to session graph, so avatar/image cache is tied to the active account and does not flicker/reset when navigating.
- Makes `MessageSharedState` session-scoped instead of root-scoped.
- Restores `NewConversationViewModel` as a graph-scoped dependency from `NewConversationGraph`, removing the temporary `SharedNewConversationViewModel` bridge.
- Fixes focus trap behavior in `HomeDrawer` so drawer content is not skipped when focus trap is disabled.
- Moves some conversation message loading work to IO dispatcher to avoid blocking UI when opening conversations.

### Issues fixed / validated manually

This targets the regressions we saw during Metro migration:

- app opening login screen while valid accounts already exist,
- account switching keeping stale session data,
- login to second account navigating back to previous account/home,
- too-many-devices flow not opening or crashing on cancel/back,
- logout of last account crashing or leaving a blank screen,
- removed-client event causing crash/blank screen instead of dialog/login/account switch,
- create conversation leaving stale creation screens on back stack,
- drawer items appearing with noticeable delay,
- avatar flicker/disappearing before conversation navigation,
- delay when opening conversations.

## How to test

Recommended manual health check:

- Fresh install with no accounts -> login flow opens.
- App startup with existing account(s) -> conversation list opens, not login screen.
- Switch between two accounts -> selected account/user data changes correctly.
- Add second account with too many devices -> remove-device screen opens.
- Cancel/back from remove-device screen -> no crash, returns to expected auth/account flow.
- Remove device and continue login -> registration proceeds.
- Logout with one account -> app lands on login flow without crash.
- Logout with two accounts -> app switches to remaining account.
- Trigger removed-client event from another device:
  - one account: dialog appears, then login flow;
  - multiple accounts: dialog appears, then switches to another valid account.
- Open drawer repeatedly -> items appear without delay.
- Open conversation from list/profile -> no avatar flicker, conversation opens promptly.
- Create group/channel conversation -> back/close returns to conversation list correctly.
- Open Cells / Shared Drive -> no Metro graph crash.


app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -1135,12 +1190,16 @@ private data class SessionNavigationState(
...
+    CreatePersonalAccountOverviewScreenDestination.baseRoute,
+    CreateTeamAccountOverviewScreenDestination.baseRoute,
+    CreateAccountEmailScreenDestination.baseRoute,
+    CreateAccountDetailsScreenDestination.baseRoute,
+    CreateAccountCodeScreenDestination.baseRoute,
+    CreateAccountSummaryScreenDestination.baseRoute,
+    CreateAccountUsernameScreenDestination.baseRoute,
 )
 
 private val authenticationGraphRoutes = setOf(
```
saleniuk, line 1216: it feels like it's just
```
private val autoenticationGraphRoutes = noSessionAuthenticationGraphRoutes.plus(
    RemoveDeviceScreenDestination.baseRoute,
)
```
😄

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -1135,12 +1190,16 @@ private data class SessionNavigationState(
 private val noSessionAuthenticationGraphRoutes = setOf(
```
saleniuk: Do we need all of these lists just to compare routes?
Can't we re-use the `CurrentScreen` where many destinations are already grouped into some meaningful groups?

core/search/src/main/kotlin/com/wire/android/search/util/ui/PreviewMultipleThemes.kt
```diff
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see http://www.gnu.org/licenses/.
  */
-package com.wire.android.util.ui
+package com.wire.android.search.util.ui
```
saleniuk, line 18: Was this change necessary?

app/src/main/kotlin/com/wire/android/ui/AppLockActivity.kt
```diff
@@ -46,19 +47,29 @@ class AppLockActivity : BaseActivity() {
     @Inject
     lateinit var loginTypeSelector: LoginTypeSelector
 
+    private val qualifiedIdMapper = QualifiedIdMapper(null)
+
     override fun onCreate(savedInstanceState: Bundle?) {
         wireApplicationGraph.inject(this)
         super.onCreate(savedInstanceState)
+        val sessionViewModelGraph = intent.getStringExtra(EXTRA_USER_ID)
```
saleniuk, line 55: Hmm, I think that app-lock shouldn't be session-scoped, app-lock is set globally, not per-account

Garzas: I agree app lock is conceptually global, but not all current app-lock screens are session-independent yet. Unlock is mostly global, but set/forgot flows still use session-backed use cases, so moving the whole app-lock graph to root in this PR could reintroduce logout/session crashes.

core/search/src/main/kotlin/com/wire/android/search/util/ui/PreviewMultipleThemes.kt
Garzas: This is only a package/path cleanup after the search module split. The annotation already belongs to com.wire.android.search.util.ui; moving the file under the matching directory avoids module/package confusion.

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
Garzas: I split the route sets by intent: no-session auth routes, session-backed auth routes, and the full auth route set. This should make it clearer why account creation/login screens do not create session graphs, while remove/register-device screens still need an explicit session user id.

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
Garzas: This logic is not about UI grouping or analytics/current screen classification. It controls whether WireActivity is allowed to create a session graph, create a session-backed auth graph, or stay in no-session auth mode.  
Some auth routes look similar from a “current screen” perspective, but differ in session ownership: login/account creation must not create a session graph, while remove/register-device screens need an explicit session user id.

MohamadJaara (approved):

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
saleniuk: Ok, but keep in mind that from now on, devs need to be aware that when someone adds some new destination then it potentially needs to be added to one of these lists for the app to work properly, so either we find some better way or at least document it properly and make everyone aware of that 🤔

saleniuk (approved):

PR #4961, opened 2026-06-12
feat(drive): In app video player (WPB-25407)

ohassine:
https://wearezeta.atlassian.net/browse/WPB-25407

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Description

- Implement in-app video player that will play drive videos instead of relying on different app using Intents.
-The player is using `ExoPlayer`

https://github.com/user-attachments/assets/e44f8520-c7c1-4590-ad15-aa5c16a8ebb2

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


features/cells/src/main/java/com/wire/android/feature/cells/ui/videoplayer/VideoPlayerViewModel.kt
```diff
@@ -0,0 +1,167 @@
...
+        player.addListener(playerListener)
+        videoUri()?.let {
+            player.setMediaItem(MediaItem.fromUri(it))
+            player.prepare()
+        }
+        if (player.isPlaying) startPositionPolling()
+    }
+
+    fun play() {
+        player.play()
```
sbakhtiarov: I think we need to connect this to lifecycle to avoid video playback when application is in background.

features/cells/src/main/java/com/wire/android/feature/cells/ui/videoplayer/VideoPlaybackState.kt
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.android.feature.cells.ui.videoplayer
+
+data class VideoPlaybackState(
+    val isPlaying: Boolean = false,
+    val isStarted: Boolean = false,
+    val isCompleted: Boolean = false,
+    val isBuffering: Boolean = false,
```
saleniuk, line 24: What's the difference between `isPlaying` and `isStarted`?
Also, wouldn't be better to have a single field representing the current state? Something like:
`enum class State { Buffering, Playing, Started, Completed }`

features/cells/src/main/java/com/wire/android/feature/cells/ui/videoplayer/VideoPlayerViewModel.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    private val navArgs: VideoViewerNavArgs = VideoPlayerScreenDestination.argsFrom(savedStateHandle)
+
+    val localPath: String? = navArgs.localPath
+    val contentUrl: String? = navArgs.contentUrl
+    val fileName: String? = navArgs.fileName
+
+    // Held in the ViewModel so playback survives configuration changes (e.g. rotating to full screen)
+    // without re-buffering the media.
+    val player: ExoPlayer = ExoPlayer.Builder(context).build()
```
saleniuk, line 53: Should ExoPlayer be part of ViewModel? ViewModels survive configuration changes and ExoPlayer holds heavy hardware resources that must be released quickly when the screen is destroyed and recreated

ohassine: The idea is that keeping ExoPlayer in the ViewModel lets playback survive rotations without rebuffering. 
We release it in onCleared(), so resources are freed once the screen actually leaves the back stack.

features/cells/src/main/java/com/wire/android/feature/cells/ui/videoplayer/VideoPlaybackState.kt
ohassine: `isPlaying` comes directly from the player and shows if playback is active right now. isStarted is a UI flag that tracks whether playback has ever begun. For example, if the user pauses, you’ll have `isStarted = true` but `isPlaying = false`

ohassine: Also a single enum doesn’t cover all the cases we need. For example, `isPlaying` and `isBuffering` can both be true at the same time, the player is actively trying to render frames but waiting for more data. Similarly, `isStarted` and `isCompleted` can both be true if playback has run to the end but was started earlier. 
These flags aren’t mutually exclusive

features/cells/src/main/java/com/wire/android/feature/cells/ui/videoplayer/VideoPlayerViewModel.kt
saleniuk: I get the idea, but context is used to create ExoPlayer and ViewModel survives configuration changes, so context can change over time and then inside it will still hold a hard reference to the previous one and use the dead context. Also, because the ViewModel is still alive and ExoPlayer points to the dead Activity, then garbage collector cannot remove it and the entire dead Activity is trapped in RAM, including all video-related UI objects and bitmaps.
Also, it makes testing more complicated as it requires Android context, so no regular unit tests.

ohassine: We’re not using an Activity context here, it’s an application context. it means it never goes stale across rotations, ExoPlayer never holds a reference to an Activity, and no dead Activity or its UI gets stuck in RAM. 

The context is only passed into `ExoPlayer.Builder(context)` and isn’t stored, and the player itself is released in `onCleared()`.
the leak you describe is a real risk with Activity context, but it doesn’t apply here

saleniuk (approved):

yamilmedina (approved):

PR #4971, opened 2026-06-16
feat: foreground service for sending pending messages (WPB-26422)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-26422

https://wearezeta.atlassian.net/browse/WPB-26422
----

# What's new in this PR?

### Issues

Current solution for sending pending messages after restoring Internet connection does not work when application is not in opened by user.

### Causes (Optional)

Caused by recent Android restrictions for background work:
- WorkManager work started when the application is in background is not allowed call setForeground.
- Without "setForeground" call the work cannot make network calls.

### Solutions

Use foreground service which is waiting for connection and retry sending pending messages.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/services/PendingMessagesForegroundService.kt
+++ b/app/src/main/kotlin/com/wire/android/services/PendingMessagesForegroundService.kt
@@ -1,0 +1,294 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+
+package com.wire.android.services
+
+import android.app.ForegroundServiceStartNotAllowedException
+import android.app.Notification
+import android.app.Service
+import android.content.Context
+import android.content.Intent
+import android.content.pm.ServiceInfo
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkRequest
+import android.os.Build
+import android.os.IBinder
+import androidx.core.app.NotificationCompat
+import androidx.core.app.ServiceCompat
+import com.wire.android.R
+import com.wire.android.appLogger
+import com.wire.android.di.KaliumCoreLogic
+import com.wire.android.di.metro.wireApplicationGraph
+import com.wire.android.notification.NotificationChannelsManager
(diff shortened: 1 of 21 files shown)


app/src/main/kotlin/com/wire/android/services/PendingMessagesForegroundService.kt
```diff
@@ -0,0 +1,294 @@
...
+        if (userId == null) {
+            appLogger.w("$TAG: missing user id, skipping pending messages send")
+            stopSelf(startId)
+            return START_NOT_STICKY
+        }
+
+        scope.launch {
+            run(userId)
+            stopSelf(startId)
+        }
```
saleniuk: I see a problem here - `onStartCommand` can be executed multiple times, so for instance when you have 2 accounts logged in and both have some pending messages - then `ACTION_SENDING_OF_PENDING_MESSAGES_SCHEDULED` will be called twice with each `userId` and here two coroutines will start for each user. 
It's using `stopSelf` with `startId`, but this logic is designed for sequential or single-task services, it does not understand concurrent operations, like multiple `userId`s processing side-by-side. If a newer `startId` completes first, it tells the OS "I am the latest command, we can shut down now," which triggers `onDestroy()` and risks killing the entire process while the first `userId` is still being processed, because the scope is cancelled in `onDestroy`, or the OS may cut off the internet and/or resources for the first coroutine which hasn't finished yet. 🤔 
Also, if another `onStartCommand` appears when some other just called `stopSelf` but the service hasn't been fully closed, then OS may decide to keep that service, but `scope` is already closed and dead.

app/src/main/kotlin/com/wire/android/services/PendingMessagesForegroundService.kt
```diff
@@ -0,0 +1,294 @@
...
+            appLogger.i("$TAG: network did not become available before timeout")
+            return
+        }
+
+        startAsForeground(createNotification(waitingForConnection = false))
+
+        PendingMessagesForegroundSyncHandler(coreLogic, sendPendingMessagesAfterForegroundSync).sendPendingMessagesForCurrentSession(userId)
+    }
+
+    private fun networkAvailability(): Flow<Boolean> = callbackFlow {
```
saleniuk: Can't we reuse our `NetworkStateObserver` here instead of implementing and registering another network callback?

app/src/main/kotlin/com/wire/android/services/PendingMessagesForegroundService.kt
```diff
@@ -0,0 +1,294 @@
...
+                NotificationIds.PENDING_MESSAGES_SYNC_NOTIFICATION_ID.ordinal,
+                notification,
+                ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC
+            )
+        }
+    }
+
+    private fun createNotification(waitingForConnection: Boolean): Notification =
+        NotificationCompat.Builder(this, PENDING_MESSAGES_SYNC_CHANNEL_ID)
+            .setContentText(
```
saleniuk, line 195: We had some issues on huawei devices crashing with "bad foreground notification" when it didn't have `setContentTitle` as well

app/src/main/kotlin/com/wire/android/services/PendingMessagesForegroundService.kt
sbakhtiarov: New version of the service sends messages for all valid accounts.

saleniuk (approved):

MohamadJaara (approved):

PR #4980, opened 2026-06-17
feat: add customizable message swipe actions

MohamadJaara:
## Summary
Adds customizable message swipe actions so users can choose what swiping left or right does in conversations.

## How it works
A new settings section lets users assign Reply, React, or Message details to each swipe direction. The app stores those choices and applies them in the conversation message list, while preventing both directions from using the same action by swapping selections.


ohassine:
Could you please link a Jira ticket for this? So, we don't lose track of it

PR #4991, opened 2026-06-23
fix: theme option padding [WPB-26141]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-26141

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- fixed padding for theme options

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/settings/appearance/CustomizationScreen.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/settings/appearance/CustomizationScreen.kt
@@ -18,7 +18,6 @@
 
 package com.wire.android.ui.home.settings.appearance
 
-import com.wire.android.navigation.annotation.app.WireRootDestination
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/wire/android/ui/home/settings/appearance/CustomizationScreen.kt
```diff
@@ -177,7 +178,9 @@ fun ThemeOptionItem(
             .background(color = MaterialTheme.wireColorScheme.surface),
         verticalAlignment = Alignment.CenterVertically
     ) {
+        HorizontalSpace.x12()
         RadioButton(selected = isSelected, onClick = null)
+        HorizontalSpace.x8()
         Text(
```
MohamadJaara, line 184: this is not the fix that is needed adding HorizontalSpace will not fix it

MohamadJaara (changes requested):

app/src/main/kotlin/com/wire/android/ui/home/settings/appearance/CustomizationScreen.kt
saleniuk: @MohamadJaara why do you think that? 
`HorizontalSpace` adds the missing paddings/spacings to the left and between RadioButton and Text label. The ticket is exactly about the broken left margin and RadioButtons being "glued" to the left side of the screen without proper left padding.

sbakhtiarov (approved):

MohamadJaara (approved):

PR #4995, opened 2026-06-24
refactor: extract viewmodel with join and start call actions [WPB-25067]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-25067

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Logic related to joining and starting a call extracted from `ConversationCallViewModel` into a dedicated reusable new view model. This new one is more universal, it doesn't take `conversationId` from `navArgs` but allows to pass it as a function parameter. It allows to use this view model in other places on screens which are not tied to a single conversation, and `ConversationCallViewModel` just extends this new one. Also, the logic from `startCallIfPossible` function is also moved to this new view model so that everything is in one place and dialog states could be unified. Added tests for both view models.

#### Test Coverage (Optional)

#### How to Test

Open group conversation and try to join ongoing or start a new call.

----

----

----
##### References


Garzas (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/call/ConversationCallViewModel.kt
```diff
@@ -72,24 +63,27 @@ class ConversationCallViewModel(
     private val observeDegradedConversationNotified: ObserveDegradedConversationNotifiedUseCase,
     private val observeConferenceCallingEnabled: ObserveConferenceCallingEnabledUseCase,
     private val observeSelf: ObserveSelfUserUseCase
-) : ActionsViewModel<ConversationCallViewActions>() {
```
sbakhtiarov, line 75: I an not sure about this inheritance. JoinOrStartCallViewModel inherits from ActionsViewModel with it's own set of actions. What if we want to re-use this logic for another screen where we need some additional actions?
Should we apply "prefer composition over inheritance" principle here and extract common logic into another class which can be re-used from any viewmodel which needs calling features?

sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/call/ConversationCallViewModel.kt
saleniuk: Good idea! I implemented it, composition instead of inheritance, manager instead of viewmodel there, and did similar thing for GroupConversationParticipantsViewModel as it was a similar case

PR #5002, opened 2026-06-25
fix: conversation and users names cut with large font (WPB-15168)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-15168

https://wearezeta.atlassian.net/browse/WPB-15168
----

# What's new in this PR?

### Issues
Conversation and user names are not fully displayed when large font is set in accessibility settings.

### Solutions
Use more lines for conversation and user names when font scaling is appied.


core/ui-common/src/main/kotlin/com/wire/android/ui/common/TitleMaxLines.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.ui.platform.LocalDensity
+
+@Composable
+@ReadOnlyComposable
+fun maxTitleLines(): Int = if (LocalDensity.current.fontScale > DEFAULT_FONT_SCALE) {
+    EXPANDED_FONT_SCALE_MAX_LINES
+} else {
+    DEFAULT_FONT_SCALE_MAX_LINES
+}
```
MohamadJaara, line 31: should we have it either 1 or 4 or maybe add a bit more scaling?

sbakhtiarov: not sure what you mean by "a bit more scaling". This is for maxLines value, so that the full title is visible, but it will not take full 4 lines unless it is needed.

MohamadJaara: i mean depending on the font scale and how big it is 
if there is a small difference between DEFAULT_FONT_SCALE and device then keep 1
a bigger difference use 2 
and a large difference use 4

sbakhtiarov: not sure if it is worth it, I really wanted to keep it simple

saleniuk (approved):

Garzas (approved):

PR #5018, opened 2026-07-01
fix: issue where when the status is set to away, the app does not ring and no join btn is visible [WPB-22834]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-22834

- Move ongoing/joinable call status out of conversation state and into an in-memory call store keyed by conversation ID.
- Use the in-memory joinable call data for conversation list/search ordering and join-call row markers.
- Remove public conversation call-state fields from Kalium conversation models.
- Keep existing DB call updates for compatibility with current consumers.
- Add a changelog fragment for the breaking API/source change.
- Fix the issue where when the status is set to away, the app does not ring and no join btn is visible, making users thinks calls are not received.
## Data Flow

```mermaid
flowchart LR
    AVS["AVS call events"]
    CallRepo["CallRepository"]
    Memory["In-memory calls map<br/>ConversationId -> Call"]
    DB["Call DB writes<br/>(kept for existing consumers)"]
    Ordering["Conversation list/search ordering"]
    AppUI["Android conversation UI<br/>join button / row marker"]

    AVS --> CallRepo
    CallRepo --> Memory
    CallRepo --> DB
    Memory --> Ordering
    Memory --> AppUI
```

## Why
Conversation list state should not depend on persisted calling rows or session-init reset work. Keeping active AVS call state in memory avoids DB pressure during Kalium init and gives consumers faster `ConversationId -> Call` lookup.

## Breaking Change
Removed call-state fields from public Kalium conversation models:
- `ConversationDetails.Group.hasOngoingCall`
- `ConversationDetailsWithEvents.hasOngoingCall`

Consumers should derive join-call UI from call APIs / the in-memory joinable-call map instead.


saleniuk (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/usecase/GetConversationsFromSearchUseCase.kt
```diff
@@ -69,41 +82,112 @@ class GetConversationsFromSearchUseCase @Inject constructor(
...
-            }.flowOn(dispatchers.io())
+        }.flowOn(dispatchers.io())
+    }
+
+    private fun Flow<PagingData<ConversationDetailsWithEvents>>.mapToConversationItems(
+        selfUserTeamId: TeamId?,
+        initialJoinableCallsByConversationId: Map<ConversationId, Call>,
+        joinableCallsFlow: Flow<Map<ConversationId, Call>>
+    ): Flow<PagingData<ConversationItem>> = channelFlow {
+        val latestJoinableCallsByConversationId = MutableStateFlow(initialJoinableCallsByConversationId)
```
ohassine: Do we need `channelFlow` and `MutableStateFlow` here? It looks like only one coroutine calls send(), and the MutableStateFlow seems to be used just to keep the latest value rather than as a reactive stream.

MohamadJaara: Good catch We still need to collect joinableCallsFlow concurrently because PagingData.map is lazy and newly loaded pages should use the latest call state without call changes reemitting the same PagingData. However, you’re right that only one coroutine emits and the MutableStateFlow is only being used as a latest-value holder. I’ll simplify this to a regular flow with a child collector and use an AtomicReference for the thread-safe snapshot.

ohassine (approved):

PR #5023, opened 2026-07-03
feat: default backend none [WPB-26206] [WPB-26586] [WPB-26624]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-26206

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Adds support for app builds that do not ship with a preconfigured default backend. When no backend is configured, the app shows a setup screen and waits for backend configuration from supported configuration sources.

## Changes

- Add `default_backend_enabled` config flag.
- Support empty default backend configuration.
- Show setup landing page when no backend is configured.
- Support backend setup through:
  - MDM / managed configuration
  - configuration deeplink
  - manually entered configuration link
  - QR code scanned externally by the system camera app
- Show success state after backend configuration is applied.
- Continue to the regular login flow after setup.
- Add support link fallback for empty support URLs using backend `websiteURL/support`.
- Support optional `supportEmail` from the configuration payload without DB migration.
- Add `feedback_menu_item_enabled` flag to hide the generic feedback menu item.
- Update setup/login copy and DE translations.

## Validation

- Setup flow tested with configuration deeplink.
- Success page verified before login form.
- Support URL fallback covered by unit test.
- Static analysis/lint passed before final copy updates.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
@@ -59,6 +59,9 @@ import com.wire.android.ui.common.textfield.WireTextFieldState
 import com.wire.android.ui.common.typography
 import com.wire.android.ui.theme.wireTypography
 import com.wire.kalium.logic.configuration.server.ServerConfig
+import java.net.URI
+import java.net.URLDecoder
+import java.nio.charset.StandardCharsets.UTF_8
 import kotlinx.coroutines.launch
 import com.wire.android.ui.common.R as CommonR
 
@@ -198,8 +201,12 @@ fun String.toBackendConfigUrl(): String? {
     }
 
     return runCatching {
-        Uri.parse(sanitizedInput)
-            .getQueryParameter(BACKEND_CONFIG_QUERY_PARAMETER)
+        URI.create(sanitizedInput)
+            .rawQuery
+            ?.split('&')
+            ?.firstOrNull { it.substringBefore('=') == BACKEND_CONFIG_QUERY_PARAMETER }
+            ?.substringAfter('=', missingDelimiterValue = "")
+            ?.let { URLDecoder.decode(it, UTF_8.name()) }
             ?.takeIf(String::isNotBlank)
     }.getOrNull()
 }
--- a/app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
@@ -50,18 +50,17 @@ import com.wire.android.ui.WireActivity
 import com.wire.android.ui.common.button.WireButtonState
 import com.wire.android.ui.common.button.WirePrimaryButton
 import com.wire.android.ui.common.colorsScheme
+import com.wire.android.ui.common.dimensions
 import com.wire.android.ui.common.snackbar.LocalSnackbarHostState
 import com.wire.android.ui.common.spacers.HorizontalSpace
 import com.wire.android.ui.common.spacers.VerticalSpace
 import com.wire.android.ui.common.textfield.WireTextField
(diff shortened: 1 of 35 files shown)


app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
```diff
@@ -0,0 +1,229 @@
...
+        )
+        VerticalSpace.x16()
+        WireTextField(
+            textState = backendConfigTextState,
+            placeholderText = stringResource(R.string.missing_backend_config_input_placeholder),
+            labelText = stringResource(R.string.missing_backend_config_input_label),
+            state = errorText?.let(WireTextFieldState::Error) ?: WireTextFieldState.Default,
+            keyboardOptions = KeyboardOptions.Default,
+            modifier = Modifier.testTag("backendConfigInput"),
+            testTag = "backendConfigInput",
```
saleniuk: is one of them redundant?

app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
```diff
@@ -0,0 +1,229 @@
...
+                    )
+                }
+            },
+        )
+        VerticalSpace.x8()
+        WirePrimaryButton(
+            text = stringResource(R.string.missing_backend_config_button_setup),
+            fillMaxWidth = true,
+            state = if (isConfigInputEmpty || isLoading) WireButtonState.Disabled else WireButtonState.Default,
+            onClick = { (onConfigurationLinkEntered ?: context::openBackendConfig)(backendConfigTextState.text.toString()) },
```
saleniuk: If `onConfigurationLinkEntered` is null then there's no action to execute for the click, so maybe it's better to hide the button or at least change the state to "disabled"?

app/src/main/kotlin/com/wire/android/ui/authentication/BackendConfigSetup.kt
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fun ServerConfig.Links.isConfigured() = api.isNotBlank()
+
+fun String.toBackendConfigUrl(): String? {
+    val sanitizedInput = trim().takeIf { it.isNotBlank() } ?: return null
+    return if (sanitizedInput.startsWith(WIRE_ACCESS_DEEPLINK_PREFIX)) {
+        runCatching {
+            URLDecoder.decode(
+                sanitizedInput.removePrefix(WIRE_ACCESS_DEEPLINK_PREFIX).substringBefore("&"),
```
saleniuk: I'm not sure if it's safe to only cut off parameters after `&` and keep the first one, because it will fail if the order of parameters is changed. I guess the goal is to get the url with config param and ignore all the rest, so it will work if we have it like this:
`wire://access/?config=https://wire-config.com&some_other_param=false`
but the params can appear in any position so for instance these two can be swapped:
`wire://access/?some_other_param=false&config=https://wire-config.com`
and then it will take wrong parameter

The better option would be to parse it to Uri, extract `config` param and build it only with that param ignoring all others

app/src/main/kotlin/com/wire/android/util/BackendSupportConfig.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    fun supportPageIntent(): Intent? =
+        CustomTabsHelper.resolveUrl("")?.let {
+            Intent(Intent.ACTION_VIEW, Uri.parse(it))
+        }
+
+    @Suppress("TooGenericExceptionCaught")
+    private suspend fun fetchSupportEmail(configUrl: String): String? = withContext(Dispatchers.IO) {
+        try {
+            val connection = URL(configUrl).openConnection() as HttpURLConnection
```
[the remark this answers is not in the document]
sbakhtiarov: /other supportEmail will be configured differently

saleniuk (approved):

sbakhtiarov (approved):

PR #5033, opened 2026-07-07
refactor: add share-safe file provider helpers PR 1 [WPB-26687]

MohamadJaara:

## Summary
- Adds share-safe FileProvider URI helpers backed by a dedicated shared cache directory.
- Routes app-generated file shares through chooser intents that can exclude Wire targets.
- Enables Android 15+ trusted share identity for Wire-origin shares.
- Updates QR/profile, feedback, avatar, and Cells share call sites to use the shared helpers.

## General Idea

```mermaid
flowchart TD

```

## Stack

### DIFF
--- a/app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
+++ b/app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
@@ -20,6 +20,7 @@ package com.wire.android.util
 
 import android.content.Context
 import android.net.Uri
+import androidx.core.content.FileProvider
 import androidx.core.net.toUri
 import okio.Path
 import dev.zacsweers.metro.Inject
@@ -31,7 +32,11 @@ class AvatarImageManager @Inject constructor(val context: Context) {
         return file.toUri()
     }
 
-    fun getShareableTempAvatarUri(filePath: Path): Uri {
-        return context.shareableFileProviderUri(context.fileProviderSharedCacheFile(filePath.name))
+    /**
+     * Creates a temporary URI that the camera app writes the captured avatar into.
+     */
+    fun createCameraOutputAvatarUri(filePath: Path): Uri {
+        val cameraOutputFile = context.fileProviderSharedCacheFile(filePath.name)
+        return FileProvider.getUriForFile(context, context.getProviderAuthority(), cameraOutputFile)
     }
 }
--- a/app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
+++ b/app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
@@ -20,7 +20,6 @@ package com.wire.android.util
 
 import android.content.Context
 import android.net.Uri
-import androidx.core.content.FileProvider
 import androidx.core.net.toUri
 import okio.Path
 import dev.zacsweers.metro.Inject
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/com/wire/android/util/FileUtil.kt
```diff
@@ -387,6 +385,94 @@ fun shareAssetFileWithExternalApp(assetDataPath: Path, context: Context, assetNa
...
+        addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+        if (!excludeOwnComponents) return@apply
+        val ownShareComponents = packageManager.queryIntentActivitiesCompat(sendIntent)
+            .map { ComponentName(it.activityInfo.packageName, it.activityInfo.name) }
+            .filter { it.packageName == packageName }
+            .toTypedArray()
+        if (ownShareComponents.isNotEmpty()) {
+            putExtra(Intent.EXTRA_EXCLUDE_COMPONENTS, ownShareComponents)
+        }
+    }
```
ohassine, line 404: I think this would read a bit more naturally as:
```
if (excludeOwnComponents) {
    ...
}
```
instead of `if (!excludeOwnComponents) return@apply`

app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
```diff
@@ -33,6 +32,6 @@ class AvatarImageManager @Inject constructor(val context: Context) {
     }
 
     fun getShareableTempAvatarUri(filePath: Path): Uri {
-        return FileProvider.getUriForFile(context, context.getProviderAuthority(), filePath.toFile())
+        return context.shareableFileProviderUri(context.fileProviderSharedCacheFile(filePath.name))
```
ohassine: Could you clarify this part? It looks like we're creating a new file, but I can't see where it's populated with the avatar data

app/src/main/kotlin/com/wire/android/util/FileUtil.kt
MohamadJaara: this is intentional to return early avoiding nested logic 

```kotlin
if (excludeOwnComponents) {
    ...
        if (ownShareComponents.isNotEmpty()) {
            ...
        }
}
```

sbakhtiarov:
I think this will break Cells code from com.wire.android.feature.cells.util.FileHelper since it is not updated. Will not be able to open Cell files in external apps.

MohamadJaara:
> I think this will break Cells code from com.wire.android.feature.cells.util.FileHelper since it is not updated. Will not be able to open Cell files in external apps.

Unfortunately, there is no e2e tests to verify it but i will test it manually and if it is broken i will handle it in its own PR

sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
MohamadJaara: This creates an empty temporary file. The camera writes the photo into it through the provided URI. After that, we process the photo and copy it to the final avatar path. I agree the name is confusing and will clarify it.

getShareableTempAvatarUri → targetPictureFileUri → ActivityResultContracts.TakePicture → camera writes data → updatePickedAvatarUri.

ohassine (approved):

PR #5034, opened 2026-07-07
feat: guard internal Wire share imports PR2 [WPB-26687]

MohamadJaara:

## Summary
- Adds import media nav args for internal Wire share flows.
- Rejects untrusted public share intents that include Wire FileProvider URIs.
- Allows trusted Wire-origin shares on Android 15+ using share identity.
- Adds internal-share validation so only Wire shared-cache URIs are accepted.
- Adds regression coverage for trusted/untrusted provider URI handling.

## General Idea

```mermaid
flowchart TD

```

## Stack


app/src/main/kotlin/com/wire/android/util/AvatarImageManager.kt
sbakhtiarov, line 40: How does it compile with removed import?

app/src/main/kotlin/com/wire/android/ui/sharing/ImportMediaAuthenticatedViewModel.kt
```diff
@@ -212,19 +251,25 @@ class ImportMediaAuthenticatedViewModel(
...
         }
 
-    private suspend fun handleImportedAsset(uri: Uri): ImportedMediaAsset? = withContext(dispatchers.io()) {
-        when (val result = handleUriAsset.invoke(uri, saveToDeviceIfInvalid = false)) {
-            is HandleUriAssetUseCase.Result.Failure.AssetTooLarge -> {
-                appLogger.w("$TAG: Failed to import asset message: Asset too large")
-                ImportedMediaAsset(result.assetBundle, result.maxLimitInMB)
-            }
+    private suspend fun handleImportedAsset(uri: Uri, rejectOwnFileProviderUri: Boolean): ImportedMediaAsset? {
+        if (rejectOwnFileProviderUri) {
```
sbakhtiarov: This is a private function in this ViewModel always called with rejectOwnFileProviderUri = false. Do we still need this condition and function parameter?

app/src/main/kotlin/com/wire/android/ui/sharing/ImportMediaScreen.kt
```diff
@@ -246,16 +253,22 @@ private fun ImportMediaAuthenticatedContent(
...
+                        context.getActivity()
+                            ?.let { activity -> importMediaViewModel.handleReceivedDataFromSharingIntent(activity) }
+                    }
                 }
             }
         }
     }
 }
 
+private fun ImportMediaNavArgs.isInternalShare(): Boolean = internalAssetUriList.isNotEmpty()
```
ohassine: I don't think this is very clean or obvious. Using `internalAssetUriList.isNotEmpty()` to determine whether the flow is internal feels implicit
I'd suggest making the source explicit with an enum
```
data class ImportMediaNavArgs(
    val source: ImportSource,
    val internalAssetUriList: ArrayList<Uri> = arrayListOf()
)

enum class ImportSource {
    EXTERNAL_SHARE,
    INTERNAL_SHARE
}
```

ohassine (commented):
I still feel using the same forward flow across all Android versions would be more consistent and easier to maintain than two different behaviors based on OS version

sbakhtiarov (approved):

ohassine (approved):

PR #5036, opened 2026-07-07
feat: share logs via Wire PR 4 [WPB-26687]

MohamadJaara:

## Summary
- Packages logs into a single zip archive before sharing.
- Adds separate log share actions for Wire and external targets on older Android versions.
- Reuses the internal import flow for sharing logs into Wire.
- Updates Debug and Log Management screens to expose the new flow.

## General Idea

```mermaid
flowchart TD

```

## Stack

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/ConversationMessagesViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/ConversationMessagesViewModel.kt
@@ -433,7 +433,7 @@ class ConversationMessagesViewModel(
         }
     }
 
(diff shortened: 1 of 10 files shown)


sbakhtiarov (approved):

app/src/main/kotlin/com/wire/android/ui/home/conversations/messages/ConversationMessagesViewModel.kt
ohassine: Minor naming suggestion: `shareAssetViaWire` doesn't actually perform the share, it only preparesthe asset. Maybe `prepareAssetForWireShare` or `getAssetForWireShare` would make more sense ?

PR #5062, opened 2026-07-16
feat(logging): add platform-independent file logging [WPB-26806]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-26806

Add a file logger that does not depend on Android Logcat.

On some devices, Logcat is disabled or unavailable. The previous log collection flow could therefore produce empty `0 B` files, leaving support teams without useful diagnostic information.

## How to reproduce

1. Run the app on a device where Logcat access is disabled or restricted.
2. Use the app and generate logs.
3. Export the diagnostic logs.
4. The resulting log files may be empty.

## What changed

- Added a Kermit-backed file log writer.
- Writes application and Kalium logs directly to a file instead of collecting them from Logcat.
- Captures AVS and CoreCrypto logs that Kalium forwards through its registered Kermit callbacks.
- Supports starting, stopping, flushing, clearing, and rotating log files.
- Compresses rotated files using gzip.
- Uses unique archive names to prevent rotations from overwriting each other.
- Preserves uncompressed logs when compression fails so they can be recovered later.
- Recovers incomplete rotations when the logger starts again.
- Prevents an in-progress rotation from recreating files after logs have been cleared.
- Flushes pending logs on low-memory callbacks without permanently stopping the logger.
- Uses the shared Kalium logger configuration as the single source of truth for the log level.
- Keeps the implementation behind `USE_PLATFORM_INDEPENDENT_FILE_LOGGER` for controlled rollout.

## Logging flow

```mermaid
flowchart LR
    App["Application logs"] --> Kermit["Kermit"]
    Kalium["Kalium logs"] --> Kermit
    AVS["AVS native callback"] --> Kalium
    CoreCrypto["CoreCrypto native callback"] --> Kalium
    Kermit --> Writer["Platform-independent file writer"]
    Writer --> Active["Active log file"]
    Active --> Rotation["Rotation and gzip compression"]
    Rotation --> Export["Diagnostic log export"]
```

## Result

Diagnostic log files no longer rely on Logcat being available. Devices with restricted or disabled Logcat can still produce useful application and supported native-library logs.

## Notes

- CoreCrypto currently forwards native logs starting at `WARN`.
- Native output written directly to Logcat or stderr without going through the registered Kalium callbacks is outside the scope of this logger.
- The legacy Logcat-based implementation remains available while the new logger is rolled out through the feature flag.


app/src/main/kotlin/com/wire/android/util/logging/LogLineTimestampFormatter.kt
```diff
@@ -0,0 +1,63 @@
...
+        append(millis)
+    }
+
+    private companion object {
+        const val TIMESTAMP_PATTERN = "yyyy-MM-dd HH:mm:ss"
+        const val MILLIS_IN_SECOND = 1000L
+        const val MILLIS_SUFFIX_LENGTH = 4
+        const val HUNDRED_MILLIS = 100
+        const val TEN_MILLIS = 10
+    }
```
MohamadJaara, line 62: this part need carfull review in case i missed something 
it does call System::currentTimeMillis and cache the the prefix and for each log line calculate only the ms 
meaning the parsing for the full date is only one time per second and the rest only MS is added to the end 
the next second it will parsed again 

if there is a better and more efficient way to do so let me know

MohamadJaara:
NEEDS MORE BENCHMARKS BEFORE REVIEWING

PR #5084, opened 2026-07-21
fix(navigation): reset graph state after client removal  [WPB-27228] [WPB-27362]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27228

## Intent

Make removed-client recovery safe while the app process remains alive: rebuild navigation from synchronized session state, discard all state owned by the invalid session, and allow the same user to log in again with a fresh Kalium session.

Depends on [wireapp/kalium#4318](https://github.com/wireapp/kalium/pull/4318).

## Reproduction

1. Log in and delete the current client remotely.
2. Let the app receive its own client-removal event.
3. Confirm the removed-client dialog without swiping the app away.
4. The activity can navigate with a replacement controller before its graph is installed.
5. After logging in again, retained session graphs, activity-scoped view models, or image requests can still belong to the deleted session.
6. The replacement session can then fail to join MLS conversations or be logged out again by stale session state.

## Root cause

The navigation controller was replaced in place while Compose could still expose the old back-stack entry. That stale route could trigger navigation against the new graphless controller.

The activity also retained its session graph and activity-scoped view models for the process lifetime. Removing a client did not invalidate those objects, so logging the same user in again could reuse dependencies and resources tied to the destroyed Kalium session.

```mermaid
flowchart LR
    A["Remote client removal"] --> B["Block session UI"]
    B --> C["Confirm dialog"]
    C --> D["Invalidate retained graph"]
    D --> E["Clear session view models and image loader"]
    E --> F["Rebuild authentication graph"]
    F --> G["Login same user"]
    G --> H["Create fresh app and Kalium session state"]
    H --> I["Join MLS conversations"]
```

## Changes

- Reset observed back-stack state whenever the navigation controller is replaced.
- Derive the background from the active back-stack entry rather than the discarded controller.
- Give each retained session graph its own `ViewModelStore` and session image loader lifetime.
- Invalidate the retained graph when the UI is blocked by a terminal session error or a session transition starts.
- Recreate the graph and activity-scoped view models when the same user logs in again.
- Point the app at the Kalium client-registration and teardown fixes from PR #4318.
- Cover repeated removal/re-login cycles, multi-account isolation, all terminal Kalium logout reasons, and image-loader shutdown.
- Extend the staging E2E flow through same-process re-login and bidirectional MLS messaging, with the test-service device verifying receipt and decryption of the app reply.

## Coverage

- Removed-client navigation and retained-graph regression tests.
- Complete app unit suite and static analysis.
- Staging E2E source compiled with the real backend and test-service flow.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
@@ -397,7 +397,7 @@ class WireActivity : BaseActivity() {
             finish = this@WireActivity::finish,
             isAllowedToNavigate = ::isNavigationAllowed
         )
-        val currentBackStackEntryState = rememberWireActivityCurrentBackStackEntryState(navigator)
+        val currentBackStackEntryState = wireActivityCurrentBackStackEntryAsState(navigator)
         val currentBaseRoute = currentBackStackEntryState.value
             ?.destination
             ?.route
@@ -1307,7 +1307,7 @@ internal fun observeAppLockUserId(
 }.filterNotNull()
 
 @Composable
-internal fun rememberWireActivityCurrentBackStackEntryState(
+internal fun wireActivityCurrentBackStackEntryAsState(
     navigator: Navigator,
 ): State<NavBackStackEntry?> = key(navigator.navController) {
     // collectAsState retains its previous value while changing flows, so reset its composition
--- a/app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
@@ -39,6 +39,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.State
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
@@ -57,6 +58,8 @@ import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelStore
+import androidx.lifecycle.ViewModelStoreOwner
 import androidx.lifecycle.flowWithLifecycle
 import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.repeatOnLifecycle
@@ -347,7 +350,7 @@ class WireActivity : BaseActivity() {
         sessionGraphStore: SessionGraphStoreViewModel = viewModel(
             factory = viewModelFactory {
                 initializer {
-                    SessionGraphStoreViewModel(appGraph)
+                    SessionGraphStoreViewModel(appGraph::createSessionViewModelGraph)
                 }
             }
         ),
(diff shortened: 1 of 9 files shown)


app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -416,12 +432,8 @@ class WireActivity : BaseActivity() {
...
             else -> null
         }
-        val backgroundType by remember {
-            derivedStateOf {
-                currentBackStackEntryState.value?.safeDestination()?.style.let {
-                    (it as? BackgroundStyle)?.backgroundType() ?: BackgroundType.Default
-                }
-            }
+        val backgroundType = currentBackStackEntryState.value?.safeDestination()?.style.let {
+            (it as? BackgroundStyle)?.backgroundType() ?: BackgroundType.Default
```
saleniuk, line 469: Why not a `derivedStateOf` anymore?

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -1237,6 +1267,15 @@ internal fun rememberWireActivityNavigator(
     )
 }
 
+@Composable
+internal fun rememberWireActivityCurrentBackStackEntryState(
```
saleniuk: It doesn't actually use `remember`, so maybe remove the "remember" prefix from the name? 🤔

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
```diff
@@ -456,6 +468,17 @@ class WireActivity : BaseActivity() {
...
     )
 
+    private fun invalidateRetainedSessionGraphIfNeeded(
+        shouldInvalidate: Boolean,
+        lastSessionGraphContext: WireActivityGraphContext?,
+        sessionGraphStore: SessionGraphStoreViewModel,
+    ): WireActivityGraphContext? {
+        if (!shouldInvalidate) return lastSessionGraphContext
+
+        lastSessionGraphContext?.sessionGraph?.currentAccount?.let(sessionGraphStore::invalidate)
```
Garzas: Could we avoid relying exclusively on `lastSessionGraphContext` to identify the graph to invalidate?
`SessionGraphStoreViewModel` survives an Activity recreation, while `lastSessionGraphContext` is held with Compose `remember` and is reset. If the Activity is recreated while a terminal error is already blocking the UI, `rememberWireActivityGraphContext()` returns null, so this value is never restored and the retained graph remains in sessionGraphs.
Logging the same user in again then returns the old `RetainedSessionGraph` from `getOrPut()`, including its session-scoped state.
I reproduced this with a unit test: retain a graph, simulate recreation by losing the remembered context while keeping the store, trigger invalidation, and request the same user again. It fails with:

```
expected: not same but was: <RetainedSessionGraph@...>
```

Could the store retain enough information to invalidate the affected user independently of the composition state, or invalidate all retained graphs for a terminal session transition?

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
MohamadJaara: Because the navigation state can now be replaced when the controller changes. derivedStateOf could keep reading the old state. The calculation is cheap, so reading the current state directly is simpler and safer.

app/src/main/kotlin/com/wire/android/ui/WireActivity.kt
MohamadJaara: good catch

saleniuk (approved):

Garzas (approved):

emmaoke-w (approved):

PR #5116, opened 2026-07-30
fix: sanitization of input fields and alignment with other platforms (WPB-16972) (WPB-24943)

yamilmedina:
https://wearezeta.atlassian.net/browse/WPB-16972

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
 
- Fix username length validation and adjust to what Backend and other platform support.
- Fix trimming of leading and trailing spaces in group name validator, to do same as web/desktop.

#### Test Coverage (Optional)

----

----

----
##### References


saleniuk (approved):

app/src/main/res/values/strings.xml
```diff
@@ -1232,7 +1232,7 @@
     <string name="settings_myaccount_display_name_title">Your profile name</string>
     <string name="settings_myaccount_display_name_description">People see your profile name in conversations, calls, and user search. It is not unique in Wire.</string>
     <string name="settings_myaccount_display_name_error">Please enter your profile name</string>
-    <string name="settings_myaccount_display_name_exceeded_limit_error">Minimum of 1 and maximum of 64 characters</string>
+    <string name="settings_myaccount_display_name_exceeded_limit_error">Minimum of 1 and maximum of 128 characters</string>
```
ohassine, line 1235: can we update the string for other languages ?

yamilmedina: Good idea, and done. 
I think they are going to be reset in crowdin, since we change the base, but nevertheless, is worth doing 🚀

app/src/main/kotlin/com/wire/android/ui/home/settings/account/displayname/ChangeDisplayNameViewModel.kt
```diff
@@ -71,6 +71,6 @@ class ChangeDisplayNameViewModel @Inject constructor(
         }
     }
     companion object {
-        const val NAME_MAX_COUNT = 64
+        const val NAME_MAX_COUNT = 128
```
MohamadJaara, line 74: i would really suggest to remove this check client side and do the rename it will fail with too long from server and we show an error message, wdyt ?

yamilmedina: It could be 🤔 , but I will leave it as it is, mainly because:
- Client side validation is not something new; we do this in other places, as it provides faster feedback to the user.
- More importantly, this is alignment with other platforms; I don't want to introduce another pattern here.

ohassine (approved):

PR #5167, opened 2026-08-12
fix: message composer accessibility [WPB-25864]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-25864

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Conversation Screen controls were not fully accessible with an external keyboard:

- messages could not be sent reliably from the composer;
- the message list and individual messages could not be navigated;
- attachment, mention and rich-text options were unreachable or had inconsistent focus order;
- some controls were reachable only with `Shift+Tab`;
- keyboard focus was not always visible;
- composer controls depended on software keyboard visibility, which does not work with an external keyboard.

### Causes

The Conversation Screen used IME visibility as a proxy for keyboard interaction. A connected hardware keyboard does not display the software keyboard, so composer controls could disappear or remain unreachable.

Some complex components also relied on the default Compose focus traversal, which did not match their visual order or expected keyboard interaction.

### Solutions

- Detects whether a hardware keyboard is connected and enables the keyboard-specific interaction flow.
- Makes the message list a single initial tab stop:
  - `Enter` enters message navigation;
  - `Up` and `Down` move between messages;
  - `Escape` returns focus to the message list container.
- Adds an activation step before editing the message composer:
  - `Enter`, `Space` or typing activates the input;
  - mouse/touch activation continues to work;
  - the software keyboard is not opened for hardware-keyboard interaction.
- Supports composer keyboard actions:
  - `Enter` sends a message;
  - `Shift+Enter` inserts a new line;
  - `Tab` and `Shift+Tab` traverse controls in both directions.
- Keeps composer actions and the collapse control available while they or the text input are focused.
- Adds dedicated keyboard navigation for attachment options:
  - `Tab`/`Shift+Tab` traverse enabled options;
  - arrow keys navigate the attachment grid;
  - `Enter` or `Space` activates an option.
- Adds keyboard navigation for mention suggestions:
  - `Tab`/`Shift+Tab` and arrow keys iterate through suggestions;
  - `Enter` or `Space` selects a mention;
  - `Escape` closes the suggestions, finishes the mention query with a space and returns focus to the composer.
- Adds cyclic keyboard navigation for rich-text actions.
- Uses the current conversation accent color for focus backgrounds instead of a fixed focus color.
- Improves semantics so keyboard wrappers expose one actionable accessibility node.
- Keeps the existing touch/software-keyboard flow separate to avoid changing mobile interaction.
- Moves the shared `Modifier.applyIf` extension to `core:ui-common`.

### Dependencies

Built on top of the Conversation Screen refactor:

- [x] https://github.com/wireapp/wire-android/pull/5143

#### Test Coverage

- [x] I have added automated tests to this contribution

Added coverage for:

- hardware keyboard detection;
- attachment focus order and disabled-option handling;
- mention navigation, activation and dismissal;
- rich-text focus wrapping;
- accessibility semantics of keyboard options.

#### How to Test

With an external keyboard connected:

1. Open a conversation.
2. Use `Tab` and `Shift+Tab` to move through the top bar, message list and composer.
3. Focus the message list, press `Enter`, and navigate messages with `Up`/`Down`.
4. Press `Escape` to return to the message list container.
5. Focus the composer and activate it using `Enter`, `Space`, typing or a mouse click.
6. Verify that `Enter` sends and `Shift+Enter` adds a new line.
7. Open attachments and navigate all enabled options with `Tab`, `Shift+Tab` and arrow keys.
8. Type `@`, navigate the mention list, select a member and verify that `Escape` dismisses the list without leaving the conversation.
9. Open rich-text options and verify cyclic keyboard navigation.
10. Repeat the basic composer and attachment flows without an external keyboard and verify that the software keyboard and touch interaction remain unchanged.


MohamadJaara (commented):
Dead-key regression reproduced; see the inline test.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerInput.kt
```diff
@@ -276,48 +317,195 @@ private fun MessageComposerTextInput(
...
+            true
+        } ?: false
+    }
+}
+
+private val KeyEvent.printableCharacter: String?
+    get() {
+        if (type != KeyEventType.KeyDown || key in NON_PRINTABLE_COMPOSER_KEYS) return null
+        val unicodeChar = utf16CodePoint
+        return unicodeChar.takeIf { it != 0 }?.let { String(Character.toChars(it)) }
```
MohamadJaara: gpt-5.6-sol commenting on behalf of Mo

--------

`utf16CodePoint` can include the `KeyCharacterMap.COMBINING_ACCENT` flag for dead keys such as `´` or `^`. Passing that flagged value directly to `Character.toChars()` crashes composer activation.

I reproduced this on `9042200f6`. The test below fails here with `IllegalArgumentException: Not a valid Unicode code point: 0x800000B4`:

```kotlin
package com.wire.android.ui.home.messagecomposer

import android.view.KeyCharacterMap
import android.view.KeyEvent
import io.mockk.every
import io.mockk.mockk
import org.junit.jupiter.api.Assertions.assertDoesNotThrow
import org.junit.jupiter.api.Test
import java.lang.reflect.InvocationTargetException

class MessageComposerDeadKeyTest {

    @Test
    fun givenDeadKey_whenActivatingComposer_thenPrintableCharacterConversionDoesNotCrash() {
        val event = mockk<KeyEvent> {
            every { action } returns KeyEvent.ACTION_DOWN
            every { keyCode } returns KeyEvent.KEYCODE_APOSTROPHE
            every { unicodeChar } returns (KeyCharacterMap.COMBINING_ACCENT or ACUTE_ACCENT.code)
        }
        val printableCharacterGetter = Class
            .forName("com.wire.android.ui.home.messagecomposer.MessageComposerInputKt")
            .declaredMethods
            .single { it.name.startsWith("getPrintableCharacter-") }
            .apply { isAccessible = true }

        assertDoesNotThrow {
            try {
                printableCharacterGetter.invoke(null, event)
            } catch (exception: InvocationTargetException) {
                throw exception.targetException
            }
        }
    }

    private companion object {
        const val ACUTE_ACCENT = '\u00B4'
    }
}
```

Please handle the combining-accent flag before converting this value, or let the text field's normal dead-key input path process it.

Garzas: Good catch, thanks. Fixed in `99e65bf63`: the composer now removes `KeyCharacterMap.COMBINING_ACCENT` via `COMBINING_ACCENT_MASK` and validates the resulting code point before calling `Character.toChars()`. I also added direct regression coverage for a dead key, a regular character, and an invalid code point. The focused unit test and `make compose-stability` both pass.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/EnabledMessageComposer.kt
```diff
@@ -132,13 +130,40 @@ fun EnabledMessageComposer(
...
     var hideRipple by remember { mutableStateOf(true) }
 
     with(messageComposerStateHolder) {
         val inputStateHolder = messageCompositionInputStateHolder
+        val mentionListFirstFocusRequester = remember { FocusRequester() }
+        var focusMentionListOnOpen by remember { mutableStateOf(false) }
+        val mentionSearchResult = messageComposerViewState.value.mentionSearchResult
+
+        LaunchedEffect(mentionSearchResult, focusMentionListOnOpen, isHardwareKeyboardConnected) {
+            if (isHardwareKeyboardConnected && focusMentionListOnOpen && mentionSearchResult.isNotEmpty()) {
```
MohamadJaara, line 147: gpt-5.6-sol commenting on behalf of Mo

--------

`mentionSearchResult.isNotEmpty()` does not guarantee that this requester has a target. `MembersToMentionUseCase` can return service contacts, while `MembersMentionList` removes them before composing `KeyboardMentionList`. With a service-only result, `requestFocus()` returns `false`, but `focusMentionListOnOpen` is still cleared, so hardware-keyboard navigation never enters the suggestions.

This instrumentation test fails on the PR head on a Pixel 6 (Android 17) with `The mention result should expose a focus target`:

```kotlin
@Test
fun givenOnlyServiceResults_whenOpeningMentionNavigation_thenFocusRequestSucceeds() {
    lateinit var firstItemFocusRequester: FocusRequester
    val service = Contact(
        id = "service",
        domain = "wire.com",
        name = "Service",
        handle = "service",
        label = "@service",
        membership = Membership.Service,
        connectionState = ConnectionState.ACCEPTED,
    )

    composeTestRule.setContent {
        firstItemFocusRequester = remember { FocusRequester() }
        WireTestTheme {
            MembersMentionList(
                membersToMention = listOf(service),
                searchQuery = "",
                onMentionPicked = {},
                useKeyboardNavigation = true,
                firstItemFocusRequester = firstItemFocusRequester,
            )
        }
    }

    val focused = composeTestRule.runOnIdle {
        firstItemFocusRequester.requestFocus()
    }

    assertTrue("The mention result should expose a focus target", focused)
}
```

Please make the open/focus condition use the same filtered list that is rendered, and do not clear the pending-focus flag when the focus request fails.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/KeyboardMentionList.kt
```diff
@@ -0,0 +1,201 @@
...
+
+        MentionKeyAction.Previous, MentionKeyAction.Next -> {
+            val move = if (action == MentionKeyAction.Previous) {
+                MentionFocusMove.Previous
+            } else {
+                MentionFocusMove.Next
+            }
+            context.focusRequesters[
+                mentionFocusTargetIndex(context.index, context.focusRequesters.size, move)
+            ].requestFocus()
```
MohamadJaara, line 166: gpt-5.6-sol commenting on behalf of Mo

--------

This requester may belong to an item that `LazyColumn` has not composed. For example, Shift+Tab on the first item wraps to the last index, but in a long list the last item is offscreen and its `FocusRequester` is unattached. `requestFocus()` returns `false`; this branch still returns `true`, consuming the key while focus remains on the first item.

This instrumentation test fails on the PR head on a Pixel 6 (Android 17): after Shift+Tab and Enter, `member-0` is picked instead of the expected `member-29`.

```kotlin
@Test
fun givenScrollableMentionList_whenMovingBeforeFirstItem_thenFocusWrapsToLastItem() {
    val members = List(30) { index ->
        Contact(
            id = "member-$index",
            domain = "wire.com",
            name = "Member $index",
            handle = "member$index",
            label = "@member$index",
            membership = Membership.Standard,
            connectionState = ConnectionState.ACCEPTED,
        )
    }
    var pickedMember: Contact? = null
    lateinit var firstItemFocusRequester: FocusRequester

    composeTestRule.setContent {
        firstItemFocusRequester = remember { FocusRequester() }
        WireTestTheme {
            KeyboardMentionList(
                membersToMention = members,
                searchQuery = "",
                onMentionPicked = { pickedMember = it },
                onDismissRequest = {},
                firstItemFocusRequester = firstItemFocusRequester,
                modifier = Modifier.height(96.dp),
            )
        }
    }

    assertTrue(composeTestRule.runOnIdle {
        firstItemFocusRequester.requestFocus()
    })
    composeTestRule.onNode(hasContentDescription("Member 0, @member0"))
        .performKeyInput {
            keyDown(Key.ShiftLeft)
            pressKey(Key.Tab)
            keyUp(Key.ShiftLeft)
            pressKey(Key.Enter)
        }

    composeTestRule.runOnIdle {
        assertEquals(members.last(), pickedMember)
    }
}
```

Please scroll/compose the destination before requesting focus (for example through `LazyListState`) and only consume the navigation key after focus actually moves.

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationTopAppBar.kt
```diff
@@ -276,7 +278,10 @@ private fun CallControlButton(
         StartCallButton(
             onPhoneButtonClick = onPhoneButtonClick,
             onAudioPermissionPermanentlyDenied = onAudioPermissionPermanentlyDenied,
-            isCallingEnabled = isCallingEnabled
+            isCallingEnabled = isCallingEnabled,
+            colors = wireSecondaryButtonColors().copy(
+                focused = MaterialTheme.wireColorScheme.primaryVariant,
```
saleniuk, line 283: Why do you edit colors? These colors schemes for buttons are synced with our design documentation so they should be used as they are - does it mean that our color schemes for buttons are incorrect?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionInputStateHolder.kt
```diff
@@ -144,6 +144,13 @@ class MessageCompositionInputStateHolder(
         inputFocused = true
     }
 
+    fun requestHardwareFocus() {
+        focusRequester.freeFocus()
+        focusRequester.requestFocus()
+        keyboardController?.hide()
+        inputFocused = false
```
saleniuk, line 151: focus is requested, but `inputFocused` is set to false 🤔

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/AttachmentOptions.kt
```diff
@@ -124,50 +126,72 @@ fun AttachmentOptionsComponent(
...
+                labelStyle = labelStyle,
+            )
+        } else {
+            LazyVerticalGrid(
+                columns = columns,
+                modifier = Modifier.fillMaxSize(),
+                contentPadding = contentPadding,
+                verticalArrangement = Arrangement.SpaceEvenly,
+                horizontalArrangement = Arrangement.Center
+            ) {
```
saleniuk, line 157: Can we combine both into one composable element but with different focus actions or modifiers?
I mean if I get it correctly both are supposed to look exactly the same, right?

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/EnabledMessageComposer.kt
```diff
@@ -397,14 +445,19 @@ fun EnabledMessageComposer(
...
                             isFileSharingEnabled = messageComposerViewState.value.isFileSharingEnabled,
                             areAttachmentOptionsEnabled = messageComposerViewState.value.areAttachmentOptionsEnabled,
+                            useKeyboardNavigation = isHardwareKeyboardConnected,
                         )
                     }
                 }
             }
+            BackHandler(mentionSearchResult.isNotEmpty()) {
+                closeMentionSuggestions()
+            }
```
saleniuk: Can this BackHandler be moved down to where other BackHandlers are? So that we have them in one place and we can make sure that their order is correct

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerFocusColors.kt
```diff
@@ -0,0 +1,27 @@
...
+/** Conversation-scoped focus colors until the app-wide focus tokens are aligned with the active accent. */
+@Composable
+internal fun messageComposerSecondaryButtonColors() = wireSecondaryButtonColors().copy(
+    focused = MaterialTheme.wireColorScheme.primaryVariant,
+)
+
+@Composable
+internal fun messageComposerTertiaryButtonColors() = wireTertiaryButtonColors().copy(
+    focused = MaterialTheme.wireColorScheme.primaryVariant,
+)
```
saleniuk, line 27: Why should it apply only to the message composer? We should have consistent colors across the whole app, so every secondary or tertiary button should look the same 🤔

app/src/main/kotlin/com/wire/android/ui/home/conversations/ConversationTopAppBar.kt
Garzas: The default focused colors are currently fixed to LightBlue50/DarkBlue800. Design asked us to use the active accent color here, so I changed it only for the controls touched in this PR. I agree the shared tokens should be updated, but I would keep the app-wide change in a separate PR.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/state/MessageCompositionInputStateHolder.kt
Garzas: Yeah, the naming is confusing here. `inputFocused` is used more like the software input/expanded state. This function returns focus to the hardware keyboard activation gate without opening the software keyboard, so setting it to `true` would change the Back and IME behavior. I would keep the behavior for now and clean up the naming separately.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/AttachmentOptions.kt
Garzas: They should look the same, but the touch version uses an animated `LazyVerticalGrid`, while the keyboard version uses a stable row/column layout for explicit directional focus. We can share more of the item rendering, but I would prefer to do that in the planned composer refactor instead of expanding this PR.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/EnabledMessageComposer.kt
Garzas: Good point. I moved all the BackHandlers into one place and outside the attachment popup condition. The mention handler is last, so it gets priority while the suggestions are open.

app/src/main/kotlin/com/wire/android/ui/home/messagecomposer/MessageComposerFocusColors.kt
Garzas: This is scoped to the composer because the shared secondary and tertiary focus tokens still use fixed blue. Design asked for the active accent color, so this is a local workaround for now. I agree the shared tokens should be fixed for the whole app in a separate PR.

MohamadJaara (approved):
LGTM

saleniuk (approved):

PR #5168, opened 2026-08-12
feat(conversation): add manual MLS migration (WPB-27446)

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27446

## Goal

Add a temporary manual path for testing Proteus/Mixed conversation migration to MLS during the transition phase.

Jira: WPB-27446

## Changes

- reveal a confirmation dialog after five consecutive taps on the Protocol row
- allow the action only when MLS migration is enabled, the protocol is Proteus or Mixed, and the conversation team matches the self-user team
- keep the dialog open on outside taps and show migration loading, success, and failure states
- invoke the backend migration flow and include backend HTTP code and error label when available
- add concise lifecycle logs and focused unit coverage

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsViewModel.kt
@@ -19,6 +19,7 @@
 package com.wire.android.ui.home.conversations.details
 
 import androidx.lifecycle.viewModelScope
+import com.wire.android.R
 import com.wire.android.appLogger
 import com.wire.android.ui.common.ActionsViewModel
 import com.wire.android.ui.home.conversations.details.options.GroupConversationOptionsState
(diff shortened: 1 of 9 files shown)


app/src/main/kotlin/com/wire/android/ui/home/conversations/details/GroupConversationDetailsViewModel.kt
```diff
@@ -263,6 +309,74 @@ class GroupConversationDetailsViewModel @AssistedInject constructor(
...
+            return
+        }
+
+        appLogger.i(
+            "[$TAG] Manual MLS migration started " +
+                    "(protocol=${groupOptionsState.value.protocolInfo.debugName()})"
+        )
+        updateState(groupOptionsState.value.copy(isMigratingToMLS = true))
+        viewModelScope.launch {
+            when (val result = migrateConversationToMLS(conversationId)) {
```
sbakhtiarov, line 347: I tried to follow the calls inside the MigrateConversationToMLSUseCase and it seems like not all of the branches with IO operations are executed in IO dispatcher.
I would recommend switching to IO dispatcher in the use case to call it safely from view model (main thread safety).

sbakhtiarov (approved):

saleniuk (approved):

PR #5170, opened 2026-08-12
feat: add keep.xml to manage resource shrinking for ringing sound [WPB-8645]

MohamadJaara:

add keep rules for the calling ringtone to make sure it does not get effected via resource shrinking


yamilmedina (approved):

app/src/main/res/raw/keep.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools"
+    tools:keep="@raw/ringing_from_them" />
```
saleniuk: Shouldn't we add the whole `raw` folder, just in case?

PR #5190, opened 2026-08-18
fix: starting conversation with user without clients (WPB-18997)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-18997

https://wearezeta.atlassian.net/browse/WPB-18997
----

# What's new in this PR?

### Issues
Android client allows creating 1:1 conversation with the user without clients (e.g. never logged in wire app).

### Solutions
Using new use case introduced in https://github.com/wireapp/kalium/pull/4404 to resolve one-on-one conversations correctly.

### DIFF
--- a/app/src/main/kotlin/com/wire/android/ui/userprofile/other/OtherUserProfileScreenViewModel.kt
+++ b/app/src/main/kotlin/com/wire/android/ui/userprofile/other/OtherUserProfileScreenViewModel.kt
@@ -41,7 +41,7 @@ import com.wire.kalium.logic.data.conversation.Conversation
 import com.wire.kalium.logic.data.id.QualifiedID
 import com.wire.kalium.logic.feature.client.FetchUsersClientsFromRemoteUseCase
 import com.wire.kalium.logic.feature.client.ObserveClientsByUserIdUseCase
-import com.wire.kalium.logic.feature.conversation.IsOneToOneConversationCreatedUseCase
+import com.wire.kalium.logic.feature.conversation.CheckOneToOneConversationIsReadyUseCase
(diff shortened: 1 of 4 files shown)


yamilmedina (approved):

ohassine:
Shouldn't we show an alert telling current user, that user X needs to open Wire or log in again first ?

app/src/main/kotlin/com/wire/android/ui/userprofile/other/OtherUserProfileScreenViewModel.kt
```diff
@@ -104,8 +104,13 @@ class OtherUserProfileScreenViewModel @AssistedInject constructor(
     }
     private fun getIfConversationExist() {
         viewModelScope.launch {
-            val isOneToOneConversationCreated = isOneToOneConversationCreated(userId)
-            state = state.copy(isConversationStarted = isOneToOneConversationCreated)
+            val readiness = checkOneToOneConversationIsReady(userId)
+            if (readiness is CheckOneToOneConversationIsReadyUseCase.Result.Failure) {
+                appLogger.w("Failed to check one-to-one conversation readiness: ${readiness.coreFailure}")
```
saleniuk, line 114: so what happens if there is a failure? `isConversationStarted` is `false` by default, so when there is a failure here and user clicks the button then the app tries to create a new conversation, right? will that work?

sbakhtiarov: Yes, it will try to resolve the 1:1 conversation anyways. This check here is just to show Start / Open conversation label consistently.

saleniuk (approved):

PR #5211, opened 2026-08-26
fix: disable pending message retry on app open

MohamadJaara:
## Goal

Prevent the foreground pending-message replay introduced in 4.32 from crashing the app when a persisted pending row contains content that cannot be encoded.

## Reproduction

1. Have a pending message whose persisted content is not valid for outbound protobuf encoding.
2. Bring the app to the foreground.
3. The global observer retries pending messages after sync and the mapper exception escapes the coroutine.

## Change

- Stop registering the app-open pending-message retry observer.
- Update observer tests to assert foreground transitions do not trigger pending-message replay.
- Keep existing explicit and manual resend paths unchanged.

A follow-up will harden pending-message encoding and re-enable the foreground trigger safely.


MohamadJaara:
already fixed in develop

PR #5220, opened 2026-08-27
feat: fetch apps and collaborators (WPB-25747)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25747

https://wearezeta.atlassian.net/browse/WPB-25747
----

# What's new in this PR?

### Issues

The Apps search relied on locally cached users, which can be incomplete for teams with more than 2,000 members. This could incorrectly show an empty Apps page even though apps were available.

### Solutions

Refresh apps once when the Apps page is opened and wait for the refresh before resolving the initial empty-query results. Subsequent searches use the refreshed local cache without repeating the synchronization.
If refreshing fails, log the failure and continue using the existing cached apps.

When apps are disabled for the conversation, render the disabled-state content directly without creating the Apps ViewModel or making network requests. The legacy services flow remains unchanged.

### DIFF
--- a/core/search/src/main/kotlin/com/wire/android/search/apps/EmptySearchDisabledByConversationContent.kt
+++ b/core/search/src/main/kotlin/com/wire/android/search/apps/EmptySearchDisabledByConversationContent.kt
@@ -17,15 +17,12 @@
  */
 package com.wire.android.search.apps
 
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.wrapContentHeight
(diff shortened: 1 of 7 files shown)


yamilmedina (approved):

core/search/src/main/kotlin/com/wire/android/search/apps/EmptySearchDisabledByConversationContent.kt
```diff
@@ -35,13 +32,10 @@ import com.wire.android.ui.theme.wireTypography
...
         modifier = modifier
-            .fillMaxWidth()
-            .wrapContentHeight()
+            .fillMaxSize()
             .padding(dimensions().spacing16x),
-        horizontalAlignment = Alignment.CenterHorizontally,
-        verticalArrangement = Arrangement.Center,
     ) {
         Text(
             text = stringResource(R.string.search_results_apps_empty_description_disabled_for_conversation),
```
ohassine, line 41: the new Box doesn't specify any alignment, the text will move to the top-start. we could use contentAlignment = Alignment.Center here

ohassine (approved):

PR #5221, opened 2026-08-27
refactor(content): add common asset models

MohamadJaara:
## Intent
Move shared asset identity and app-private file state behind KMP-safe models without pulling Android APIs into common code.

## Changes
- add focused content KMP contracts for prepared assets, opaque external references, portable platform results, and naming policy
- split Android URI navigation wrappers from the common prepared-asset model
- replace app-private File and Path state in affected ViewModels with okio.Path and the existing KaliumFileSystem where practical
- keep URI parsing and WireActivityViewModel at the Android edge

## Scope
The current develop inventory still contains all 19 directly affected ViewModels. This first layer introduces only the shared models and filesystem state needed by later PRs.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -213,6 +213,7 @@ dependencies {
     implementationWithCoverage(projects.core.di)
     implementationWithCoverage(projects.core.media)
     implementationWithCoverage(projects.core.mediaPlayer)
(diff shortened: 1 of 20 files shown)


MohamadJaara:
Closing because this stack belongs in wireapp/wire-prism.

PR #5226, opened 2026-08-27
fix(sharing): preserve trusted caller validation [WPB-8645]

MohamadJaara:

## Goal

Keep public shares containing Wire FileProvider URIs gated by the verified sender throughout activity intent handling.

## Repro

1. Keep Wire running so its singleTask activity receives a share through onNewIntent.
2. Share a Wire FileProvider URI into the app.
3. On Android 15 or newer, replacing the activity intent drops the verified ComponentCaller.
4. Intent queueing also loses the caller validation result before the early deep-link security check runs.

## Change

- Preserve the current ComponentCaller when replacing an intent on Android 15 and newer.
- Capture caller validation when each intent is received and carry it through the intent coordinator.
- Pass the captured result into the existing early deep-link rejection check.
- Keep the import-screen validation as a second security check.
- Use the shared Android-version capability boundary for inbound validation.

## Security behavior

- Android 14 and older continue treating Wire-provider public shares as untrusted.
- Android 15 and newer require the verified caller to hold the URI read grant.
- External provider URIs keep their existing behavior.

## Coverage

Regression coverage verifies Android 14 rejection, Android 15 grant acceptance and denial, caller preservation, per-intent queueing, and both existing validation layers.


yamilmedina (approved):
nice fix! I think this is why sharing logs to wire does not work now 🤔

ohassine (approved):

PR #5266, opened 2026-09-09
feat: adminless group delete reminder (WPB-25284)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25284

https://wearezeta.atlassian.net/browse/WPB-25284
----

# What's new in this PR?

Show "adminless delete reminder" event announcing that the group will be automatically deleted.


app/src/main/kotlin/com/wire/android/ui/home/conversations/messagelist/ConversationMessageList.kt
```diff
@@ -353,6 +363,12 @@ fun ConversationMessageList(
                         }
                     }
 
+                    // Multiple children in a reverse-layout item are placed in reverse order,
+                    // so emitting the reminder first displays it below the newest message.
+                    if (index == 0 && adminlessGroupDeletionTimestamp != null) {
+                        AdminlessGroupDeleteReminderItem(adminlessGroupDeletionTimestamp)
+                    }
```
saleniuk, line 370: We have adding `AdminlessGroupDeleteReminderItem` in two places, so I think that if we remove this one addition and keep only the one that's above this `items` section (lines 306-313), but remove `&& lazyPagingMessages.itemCount == 0` from line 306, then that one will be also always added before the whole list meaning it will be the first element below the newest message, but this way we won't have the code that adds `AdminlessGroupDeleteReminderItem` in two places, right?

sbakhtiarov: This will break message indexing

app/src/main/kotlin/com/wire/android/mapper/ConversationMapper.kt
```diff
@@ -183,6 +186,18 @@ fun ConversationDetailsWithEvents.toConversationItem(
     }
 }
 
+private fun ConversationDetailsWithEvents.adminlessDeleteAwareLastMessage(
```
Garzas, line 189: I would rather move this toUiPreview functionality than having special named function here

saleniuk (approved):

app/src/main/kotlin/com/wire/android/mapper/ConversationMapper.kt
sbakhtiarov: This one does the last message override rather then doing a regular message mapping.

app/src/main/kotlin/com/wire/android/ui/home/conversations/messagelist/ConversationMessageList.kt
```diff
@@ -301,6 +303,14 @@ fun ConversationMessageList(
                 ),
                 modifier = Modifier.fillMaxSize(),
             ) {
+                if (adminlessGroupDeletionTimestamp != null && lazyPagingMessages.itemCount == 0) {
```
ohassine, line 306: Should we also check that `adminlessGroupDeletionTimestamp` is still in the future? so we don't show an expired deletion reminder
```
val showAdminlessDeleteReminder =
    adminlessGroupDeletionTimestamp?.let { it > Clock.System.now() } == true
```

sbakhtiarov: The reminder "expires" together with the conversation. When it expires the BE deletes the whole conversation and it will not be visible anymore. 
Yes, if user will stay in the app and disable Internet they may eventually see the "expired" reminder. But considering the very low chance of showing this reminder multiplied by low chance of such scenario I would keep it as is. In worst case it is still better to show expired reminder than hiding it.

ohassine (approved):

PR #5300, opened 2026-09-17
feat: support multi-ingress SSO login (WPB-27414)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-27414

https://wearezeta.atlassian.net/browse/WPB-27414
----

# What's new in this PR?

### Issues

Logging in with SSO always showed the "account already in use" dialog when a retained logged-out account existed, instead of the multi-ingress identity-provider flow that the email/password path already supports.

### Causes (Optional)

The identity-change check needs the IdP ID that started the login, but the SSO flow leaves the app for the browser and returns through a deep link that only carries the cookie and the server config ID. Nothing linked the returning redirect back to the IdP used in `/sso/initiate-login/{idp}`.

### Solutions

- Added `PendingSsoLogin` (parcelable: IdP ID, server config ID, whether the backend capability still needs checking), stored in the login `SavedStateHandle` via `LoginSavedInputStore`, so the context survives the browser round trip and process death.
- `LoginSSOViewModel` and `NewLoginViewModel` record it from `SSOInitiateLoginResult.Success` when the browser is opened, and consume it when the deep link comes back; it is cleared on SSO failure and before each new initiate.
- `LoginSSOViewModelExtension.establishSSOSession` passes that context to `getLoginSession`, requests the backend capability check for the SSO-code path only, and supplies the IdP ID to `AddAuthenticatedUserUseCase` only when the backend reports `ssoIdpChangeDetectionEnabled`. When the flag is off, the existing login flow is unchanged. The email path keeps supplying its IdP directly.


app/src/main/kotlin/com/wire/android/ui/authentication/login/sso/LoginSSOViewModelExtension.kt
```diff
@@ -110,6 +110,7 @@ class LoginSSOViewModelExtension(
             onAddAuthenticatedUserFailure(AddAuthenticatedUserUseCase.Result.Failure.SsoIdentityChanged)
         },
         ssoIdentityProviderId: String? = null,
+        pendingSsoLogin: PendingSsoLogin? = null,
```
saleniuk, line 113: Should we allow nullable `PendingSsoLogin`? If I get it correctly, then `initiateSSO` always needs to be called before `establishSSOSession`, and then it should always be non-null 🤔

app/src/main/kotlin/com/wire/android/ui/authentication/login/sso/LoginSSOViewModelExtension.kt
```diff
@@ -123,7 +124,18 @@ class LoginSSOViewModelExtension(
             }
         }
 
-        val ssoLoginSuccess = when (val ssoLoginResult = authScope.ssoLoginScope.getLoginSession(cookie)) {
+        val hasMatchingContext = pendingSsoLogin?.let { it.serverConfigId == serverConfigId } ?: (ssoIdentityProviderId != null)
```
saleniuk, line 127: In case `pendingSsoLogin` is null, then it will return failure when `ssoIdentityProviderId` is null - is that intentional? Previously it was possible to execute that with `ssoIdentityProviderId` as null and `LoginSSOViewModel` never passes this field, so it's always null when called from that view model

saleniuk (approved):

ohassine (approved):

